Issue #39, opened 2013-11-07
additions to the "usage" part


ccurvey:
new user here, please tell me if this is noise.

after pip install requirements/local.txt....

1) Go create a database for your site to use
2) edit myapp/config/settings.py and change the database URL.  This should be 

```
postgres://username:password@hostname:port/databasename
```

3) python manage.py syncdb (to create basic tables)
4) python manage.py migrate (to install the migrations)
5) python manage.py runserver (to start the server)
6) hit localhost:8000 (to see if everything worked)

At this point, you WON'T have any admin user set up.  (Is this a 1.6 bug?  Usually the initial user is set up at the first syncdb.)

7) edit your settings again and find EMAIL_BACKEND.  change "smtp" to "console".  This is so you don't have to go through all the rigamarole of getting email connectivity working right now. 

8) use the "Sign up" link available from http://localhost:8000 to sign up for an account.  The verification email will appear in the window where you started the server.  (The console.)

9) Copy the confirmation link from the console and paste it into your browser.  Confirm your email, and you'll be taken to the login screen.

10) Now you can login to the site, but you still don't have an administrative user.  To make the user you just registered an admin user, fire up a postgres query tool (psql, pgadmin3, etc), and run:

```
update users_user
set is_superuser = true
, is_staff = true
where id = 1; # or use username or email to restrict your update.
```

Now you are ready to start creating your own app.  (I'll file another ticket with that doco when I get it working.)

audreyr:
I'm working on cleaning this up and moving it into the README :ocean:

Issue #34, opened 2013-10-16
List dependencies outside pip


chriscabral:
I had to install memcached before I could install the requirements. I know this is a pip dependency problem but just thought you might want to list the external dependencies on the readme.

theskumar:
Seems like has already been resolved. @luzfcb can you please confirm and close this issue if it's not needed anymore.

PR #67, opened 2014-01-29
Add documentation on 'getting up and running'

keybits:
Some hopefully useful instructions to help people get up and running, including usage of the new Grunt task.

Re: Issue #34 on my machine I didn't need to install memcached to successfully install the local requirements.

Re: Issue #39 I've linked to this issue for now. It would be good to remove the need for those steps and then this link can be removed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,8 +78,48 @@ Create a GitHub repo and push it there::
 
 Now take a look at your repo. Don't forget to carefully look at the generated README. Awesome, right?
 
-It's time to write the code!!!
+Getting up and running
+----------------------
 
+The steps below will get you up and running with a local development environment. We assume you have the following installed:
+
+* pip
+* virtualenv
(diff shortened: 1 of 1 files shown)


pauloalem:
Hey guys. I'm using grunt for a new project and while deploying to heroku I had some problems because, well... I don't think we should commit css files to the repository when working with Sass, and so I didnt.
On the collect static step, the css files never got built and only the scss files got copied over.
Did I miss anything obvious?
I managed to do the deploy the way I wanted using the [multi buildpack](https://github.com/ddollar/heroku-buildpack-multi.git) and then combined the python buildpack plus a custom [node-grunt-compass buildpack](https://github.com/pauloalem/heroku-buildpack-nodejs-grunt-compass.git).
What do you think? I could work on a pull request if there's any interest to add that to the docs or the template.
Cheers

Issue #136, opened 2014-10-03
Adding database credentials


berndca:
When I ran ./manage migrate for the first time I got 

    django.db.utils.OperationalError: fe_sendauth: no password supplied

I'm proposing to add credentials to the database URL replacing:

    DATABASES = values.DatabaseURLValue('postgres://localhost/{{cookiecutter.repo_name}}')		

with 

    DATABASES = values.DatabaseURLValue('postgres://localhost/django:blank@{{cookiecutter.repo_name}}')	

In addition I would add a security check to make sure that an error is flagged if the production database password is 'blank'. I have opened a pull request #139.

theskumar:
Appropriate documentation has been added and we feel that is enough to address this.

PR #139, opened 2014-10-03
Feature/database url

berndca:
This is my second pull requests related to issue #136. It appears that username and password are required in the database url. I have added default values for development and a check to ensure that a password other than 'blank' is used. Please share your feedback and concerns.

Thanks, Bernd

### DIFF
--- a/{{cookiecutter.repo_name}}/requirements/base.txt
+++ b/{{cookiecutter.repo_name}}/requirements/base.txt
@@ -32,5 +32,7 @@ django-autoslug==1.7.2
 # Useful things
 django-avatar==2.0
 
+# Time zones support
+pytz==2014.7
 
 # Your custom requirements go here
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/__init__.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/__init__.py
@@ -1,5 +1,18 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
+import os
+from django.conf import settings
+from django.core.checks import register, Error
 
 from .local import Local  # noqa
 from .production import Production  # noqa
+
+
+@register(settings)
+def dj_database_url_check(app_configs=None, **kwargs):
+    errors = []
+    password = settings.DATABASES['default'].get('PASSWORD')
+    config = os.environ['DJANGO_CONFIGURATION'].lower()
+    if password == 'blank' and config == 'production':
+        errors.append(Error('Change the database password for production.'))
+    return errors
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/__init__.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/__init__.py
@@ -1,18 +1,21 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
 import os
-from django.conf import settings
-from django.core.checks import register, Error
 
 from .local import Local  # noqa
 from .production import Production  # noqa
 
 
-@register(settings)
-def dj_database_url_check(app_configs=None, **kwargs):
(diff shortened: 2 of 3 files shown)


theskumar:
I think the current implementation with the updated README [1] is clear enough on how to use dj-database url. Generally when you install a postgres, the database you are connecting doesn't require a password. 80% use case. I never required one during development. The database creation using command line has always been just `createdb <project_name>` command.

[1] https://github.com/pydanny/cookiecutter-django#getting-up-and-running

I would like to close this issue without doing anything. @pydanny @luzfcb Thoughts?

theskumar:
Update: probably we can just override the `DATABASE_URL` to be a secretValue or an empty value and force user to set is explicitly. (Not required if you have deployed on Heroku) 

PS: But not sure if `values.DatabaseURL()` can be a secretValue().

philippeluickx:
So I tried setting secretValue() because I am using a password to access the DB. Things don't really work out, as I get complaints about database needing a "default".
Isn't it generally a bad idea to use a password in your conf?

theskumar:
> Isn't it generally a bad idea to use a password in your conf?

Yes it is. 

You'll still need to use `values.DatabaseURL()` and not `secretValue()` for `DATABASE` setting as under the hood. `dj-database-url` [1] is used to parse the database url. 

[1]
https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/requirements/base.txt#L8, 
https://github.com/kennethreitz/dj-database-url
https://django-configurations.readthedocs.org/en/0.5/values/#configurations.values.DatabaseURLValue

philippeluickx:
Which I am, but somehow not feeling happy with it...

theskumar:
As such there is no alternative to it, unless you want to open up an issue at https://github.com/jezdez/django-configurations/. 

This has to be improvement in `values.DatabaseURL()` to accept something like `required=True`. I haven't tested what happens if you keep it empty in `production.py`

philippeluickx:
Opened https://github.com/jezdez/django-configurations/issues/95

theskumar:
> Hence, is it somehow possible to include an environment variable when using values.DatabaseURL()

@philippeluickx the DatabaseURL() can read from the environment variable. The only different between DatabaseURL and SecretValue is that you declare a SecretValue() is the property becomes a required property and django won't start if that value is missing in environment variable. 

What we want is a way to make DatabaseURL as required property and force user to set it via environment variable during production.

philippeluickx:
Right, ok, now I get it. Will adapt the ticket, but less urgent in this case.
So if I get it right: I can leave the value empty and django-configurations will go and fetch it from the env vars?

theskumar:
​Yes. The name of environment variable is `DATABASE_URL`​ it's added to
heroku by default when you add postgres database.

I believe the cause of confusion is all the things are implicit. :?

> Explicit is better than implicit. -- Zen of python

philippeluickx:
You sir, made me smile!
Wondering which would be the best way to improve docs. In all honesty, I am trying to setup a new server and adopting best practices. Coming from Apache and lacking some nice up-to-date docs for Gunicorn.
Where to put environment vars in dev, prod,..?

Which leads back to this ticket, agreed we need more docs and explicity!

theskumar:
@philippeluickx give https://github.com/theskumar/python-dotenv a shot. I've been using it in production servers and it works like charm. Happy to help with anything you are stuck.

jwegis:
@philippeluickx @theskumar Another project to add to the mix: https://github.com/joke2k/django-environ

luzfcb:
@jwegis @theskumar another: https://github.com/henriquebastos/python-decouple

theskumar:
@luzfcb these all look like great projects. I think they should find a place in a wiki or somewhere else as alternatives to tools used in this project. 

django-environ and python-decouple would require not to use django-configuration. It seems `.env` support #197 is coming to django-configuration as well.

PR #154, opened 2014-11-13
adding authors.rst, contributing.rst and readme.rst to docs.

arruda:
Also changed 'CONTRIBUTORS.txt' to 'AUTHORS.txt' following the example of https://github.com/audreyr/cookiecutter-pypackage
and also django (they use only 'AUTHORS' there, but I the idea is the same).


theskumar:
@arruda Hey man! wondering if you still have time to continue with this PR.

arruda:
@theskumar sorry! I'll get back into this tomorrow, I complete forgot about this!

arruda:
@theskumar o/ I think it's done, I removed all the references to testing, since the project don't specify any of these.

theskumar:
@arruda thanks for the updates. Overall looks great! 

Few minor issues, I noticed:

1. I think using `readthedocs` links in `Readme.rst` should be avoided and I would prefer a local or github link instead. I am assuming that majority of people using django-cookiecutter do not host their documentation on readthedocs.
2. `CONTRIBUTING.rst` seems little specific to a particular project. esp. with github issue tagging rules. Wondering if we can use a simplified version. Something like: https://github.com/pythonindia/junction/blob/master/CONTRIBUTING.md

Cheers,

pydanny:
@arruda What did you want to do with this pull request?

arruda:
The idea was to hava a `docs` with more pre-made content, using the cookiecutter-pypackage as an example. Another thing was to reuse the docs in the readme, to add links in the readme to the the full content of a subject that is present in the docs. 
This way it would avoid creating a big readme and it would be consistent with what's on the docs.

The problem is that for what I'm seeing, there isn't a nice way of doing this without considering that the project is going to be hosted in readthedocs (since some rst docs are composed of others and this rst includes don't appear to be taken into account by github). 

So I'm not sure what is the best solution here =/

OBS (sorry for taking so long @theskumar and @pydanny, I started writing the reply long ago by email and forgot to send it)

audreyr:
I think a good solution would be:
* Break up the `README` into separate files in `docs/`. This would give everyone a chance to improve upon the docs
* The new `README` would be pretty high-level. Only project name, badges, description, features, possibly basic usage, and a note to see the full docs for more info.
* Nothing specific to readthedocs. But standard rst includes aren't specific to readthedocs. They don't work on GitHub (https://github.com/github/markup/issues/172) but they work locally. I think it would be fine to use includes to avoid duplication.

pydanny:
@arruda I'm breaking up the README into something more manageable. Following your idea of course. :smile: 

That said, I wonder if we should maintain a non-dynamic set of Sphinx docs at the root of the project. That way we can have docs on ReadTheDocs. The problem then is duplication between two sets of Sphinx docs.

It's a bit ungainly, so I'm open to suggestions.

arruda:
Hi, I liked Audrey idea. Not sure if I understood yours. Are you saying to
have both ways in the docs? One with the includes and one without?
If that's the case then I think it would be better to choose one or the
other.
The idea of using the includes is to avoid duplication, so if we'll have to
duplicate many files just have some of them using includes it miss the
point.
I was thinking of another idea, but I'm not sure it would render ok (in
github or ReadTheDocs). The idea was to avoid the use of includes, and
instead we could add relative links to other docs.
I'm not sure if these relative links to other files work in Sphinx or
github.

What do you think?
Em 30/08/2015 16:20, "Daniel Greenfeld" <notifications@github.com> escreveu:

audreyr:
@arruda Try out the relative links and see how it goes. I'm not sure either, but I'm optimistic that they would work.

arruda:
@audreyr There is another small problem now... you see, if we add links like:

On readme.rst:
```
`Link for Contributing <contributing.rst>`_
```
And check it in github, (opening the readme.rst file in the repo), it will render ok the relative link to `contributing.rst`.

The problem is that if we go over the Sphinx rendered html, it will still link to the `.rst` when it actually point to the `contributing.html` that was generated.

I`m not sure on how we could avoid this problem, if it`s possible to tell sphinx to replace the .rst by the .html at the generation of the html files.

pydanny:
@arruda I've summarized and expanded your ideas into a new issue, #335, which should resolve the documentation issues the project has been facing. Thank you so much for your concept and I hope you can help us with the documentation reorganization. :smile:

arruda:
:+1:

PR #173, opened 2015-01-04
Fix lost import

zcho:
Return lost import for 
raise forms.ValidationError(self.error_messages['duplicate_username'])


luzfcb:
@zcho  please, update the pull-request and include your name and github username in https://github.com/pydanny/cookiecutter-django/blob/master/CONTRIBUTORS.txt

To update this pull-request, make commit on same branch that was used to create it:

https://github.com/zcho/cookiecutter-django/tree/patch-1

luzfcb:
thanks @zcho

PR #174, opened 2015-01-04
Update django-allauth 0.18.0 --> 0.19.0

hairychris:


hairychris:
Not tested properly but checked change log and seem to be no major changes,

PR #175, opened 2015-01-07
Fix test discovery when using explicit relative imports

foxwoods369:
cookiecutter-django fails to run tests if the following users/tests.py is created:
```python
from . import models
```
With the following message:
```
ERROR: Failure: RuntimeError (Conflicting 'user_groups' models in application 'users': <class 'users.models.User_groups'> and <class '{{ cookiecutter.repo_name }}.users.models.User_groups'>.)
```

This occurs because the test runner adds to `sys.path` as it goes up the directory structure looking for the first directory without an `__init__.py`. With the current setup, since we are storing `urls.py` and `wsgi.py` in the second level down, we have `__init__.py` at the same level, so an explicit relative import as above breaks the test db as it references {{ cookiecutter.repo_name }}.users.models instead of users.models.
Some discussion is here: https://code.djangoproject.com/ticket/22280 and here: http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap

I suggest we move `__init__.py`, `urls.py`, and `wsgi.py` down into a module with the same name as `{{ cookiecutter.repo_name }}` and we reference them accordingly in `config/common.py`. 

Tested using django's test runner with the aforementioned tests.py


foxwoods369:
Just realized that this may affect `wsgi.py` a bit more - any suggestions there?

theskumar:
I'll make sure this get fixed while implementing #196. Thanks a lot Skyler for heads up and pointing the issue.

PR #176, opened 2015-02-09
Use double quotes for author name fields

kevgathuku:
This is meant to handle errors caused when the author name
contains punctuation i.e apostrophes
This is causing some errors in that the apostrophes are closing some single quotes fields early
This addresses #177


luzfcb:
@kevgathuku thanks XD

kevgathuku:
You're welcome :+1:

PR #178, opened 2015-02-09
Change full_name to author_name in setup.py

kevgathuku:
Replace `cookiecutter.full_name` with `cookiecutter.author_name`
The full name setting does not exist and creates a
blank author field in setup.py


luzfcb:
Fixed in b2938581d901bb80222747292bf4f426334e3a4a

@kevgathuku Thanks.

kevgathuku:
:+1:

Issue #167, opened 2014-12-18
This question makes no sense


mark0978:
repo_name (default is "repo_name is used for describing the directory structure.")? 

Huh?

theskumar:
closing this, as this has been address by @kevgathuku in #179. We will try to improve upon the UI as much we can.

Thanks @mark0978! :blush:

PR #179, opened 2015-02-09
Set a more pleasant default project_name

kevgathuku:
This suggests a less daunting default `project_name` variable
Related to issue #167


luzfcb:
@pydanny personally I do not like `_` on the name of the github repository. I prefer `-`

PR #185, opened 2015-02-12
Fix small typo in wsgi.py

kevgathuku:
Add missing `cookiecutter`  reference in wsgi.py


luzfcb:
Good point. @kevgathuku Thanks again.

kevgathuku:
You're welcome

Issue #187, opened 2015-02-16
Collectfast and Django 1.7+


antonagestam:
The way Collectfast is added to `INSTALLED_APPS` ([here](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.repo_name%7D%7D/config/production.py#L77)) may be problematic for Django 1.7+, see this thread for more information: https://github.com/antonagestam/collectfast/issues/49#issuecomment-73836059

I'd suggest moving it to common.py and adding `COLLECTFAST_ENABLED = False` to local.py if Collectfast really isn't wanted there. However I don't really see why you'd want it installed on production only but there might be reasons to that that I don't know of.

theskumar:
`Collectfast` was added to production only because S3 config were introduced only in `production.py`. We didn't wanted to have `Collectfast` as common requirement. 

I don't think there is any harm moving it to `common.py` with proper commented doc, given your explanation above. 

@pydanny thoughts?

theskumar:
@antonagestam we have taken care of this in #199. Thanks to @ka7eh!

antonagestam:
:+1:

PR #199, opened 2015-03-10
Updated INSTALLED_APPS for 'collectfast'.

ka7eh:
Quote from https://github.com/antonagestam/collectfast:
>For Django 1.7+, <code>'collectfast'</code> should come before <code>'django.contrib.staticfiles'</code>. For Django versions below 1.7, it should come after <code>'django.contrib.staticfiles'</code>. Please note, that failure to do so will cause Django to use <code>django.contrib.staticfiles</code>'s <code>collectstatic</code>.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -37,6 +37,7 @@ Martin Blech
 Andy Rose
 Andrew Mikhnevich / @zcho
 Kevin Ndung'u / @kevgathuku
+Kaveh / @ka7eh
 
 * Possesses commit rights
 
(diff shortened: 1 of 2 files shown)


theskumar:
Hey @ka7eh, the solution looks great! Let's not do any checks for Django version, we can always make the `collectfast` to be the first installed app and a comment on the top why it needs to be on the top[1].

Also, do make sure to add yourself to https://github.com/pydanny/cookiecutter-django/blob/master/CONTRIBUTORS.txt :blush: 

[1] https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.repo_name%7D%7D/config/production.py#L40-L46

closes #187

ka7eh:
Hey @theskumar,  just updated the PR :)

theskumar:
Awesome! merging it in. Thanks a lot! :)

PR #200, opened 2015-03-14
add requirements file of non-python dependencies - fix issue #34

luzfcb:


luzfcb:
@pydanny @theskumar review me please

pydanny:
:+1:

{{cookiecutter.repo_name}}/install_os_dependencies.sh
```diff
@@ -0,0 +1,20 @@
...
+    echo "sudo ./${0##*/}" 2>&1
+    echo -e "\n" 2>&1
+
+    exit 1
+else
+
+    apt-get update
+
+    # Install the basic compilation dependencies and other required libraries of this project
+    cat requirements.apt | grep -v "#" | xargs sudo apt-get install -y
```
dupuy: These commands will not only install all packages that are necessary, but will also upgrade any packages already installed to the latest version.  That may not be a bad thing, but I prefer to err on the side of caution when doing root-level administration of somebody else's system.  Having an `--(no-)upgrade` option, or at least warning the user in the docs that packages will be updated would be better.

luzfcb:
@dupuy I refactored the install_os_dependencies.sh file. Is good now or I missed some more error?

pydanny:
:+1:

PR #201, opened 2015-03-17
fix(settings): fix non-existing import of s3

theskumar:
- `from S3 import CallingFormat` - the S3 module never existed.
- fixed the calling format accoroding to `STATIC_URL`


theskumar:
cc: @pydanny @luzfcb

Issue #125, opened 2014-08-26
Gunicorn and django-configurations error (after settings refactoring)


restless:
After recent settings refactoring it's not possible to start the application using gunicorn. It raises:

    File "/home/user/www/project/project/settings/__init__.py", line 3, in <module>
    from .local import Local # noqa
    File "/home/user/www/project/project/settings/local.py", line 16, in <module>
    class Local(Common):
    File "/home/user/.virtualenvs/project/local/lib/python2.7/site-packages/configurations/base.py", line 30, in __new__
    raise ImproperlyConfigured(install_failure)
    ImproperlyConfigured: django-configurations settings importer wasn't correctly installed. Please use one of the starter functions to install it as mentioned in the docs: http://django-configurations.readthedocs.org/

The solutions seems to be adding explicit:

    from configurations import importer
    importer.install()

to the config/__init__.py (in my case config dir is called settings)

Also see: https://github.com/jezdez/django-configurations/issues/25

garry-cairns:
I don't know if this is related but I'm also having gunicorn problems with essentially a default setup. I've [sought an answer for the problem without success thus far](http://stackoverflow.com/questions/28420295/why-might-gunicorn-be-hanging-in-this-configuration). Thanks to the way I'm building my project it should be possible to [recreate my conditions entirely](https://github.com/garry-cairns/correspondence) if that's useful. I'll continue searching for the answer in the meantime.

garry-cairns:
I've now [sorted my issue](http://stackoverflow.com/questions/28420295/why-might-gunicorn-be-hanging-in-this-configuration) so you can safely ignore me on this one.

restless:
I'm closing it as I also can't reproduce it at the moment.

pydanny:
Might this be the solution: http://stackoverflow.com/a/28489453/93270

If it is, let's please add it to the documentation. Right now nearly everything points to Heroku deployments but... not everyone is using this on Heroku. :wink:

garry-cairns:
Yes that was my solution. I'm getting very close to deploying on linked Docker containers on AWS EC2. Would you take a docs PR for how I did so? (Subject to my docs and process being up to standard of course!)

pydanny:
@garry-cairns Absoloutely we'll take a docs PR. 

That said, could it go into the `docs` directory as a 'Docker/EC2' file? We're going to be moving all deployment instructions there so we can break the options up by category.

garry-cairns:
[...]

PR #203, opened 2015-03-21
Added Docker develop and deploy instructions in response to https://github.com/pydanny/cookiecutter-django/issues/125#issuecomment-74349410

garry-cairns:
I promised you some documentation about developing and deploying in Docker/EC2 on [issue 125](https://github.com/pydanny/cookiecutter-django/issues/125#issuecomment-74349410). Sorry for taking longer than I said, things been a bit hectic at my end.

I hope this is useful. I'm not totally convinced I have this in the right location based on your comment so you may want to move it should you choose to accept the PR. I haven't added myself to contributors since it's only docs but, again, feel free to change that should you choose to accept the PR and think it appropriate. The reason for two commits is that I accidentally commited a bunch of changes to the Makefile when I was making sure my documentation built correctly.

### DIFF
--- a/{{cookiecutter.repo_name}}/docs/Makefile
+++ b/{{cookiecutter.repo_name}}/docs/Makefile
@@ -1,153 +1,66 @@
-# Makefile for Sphinx documentation
-#
+# Makefile for Sphinx LaTeX output
 
-# You can set these variables from the command line.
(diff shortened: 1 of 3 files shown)


dwcaraway:
Nice!

Issue #121, opened 2014-07-20
'STATIC_URL' value is not set properly


burhan:
The issue is at this line:

    STATIC_URL = 'https://s3.amazonaws.com/%s/' % AWS_STORAGE_BUCKET_NAME

The [`__repr__` method for ``values.Value``](https://github.com/jezdez/django-configurations/blob/master/configurations/values.py#L42) will return a string not suitable to be inserted into a URL: `<Value Default: None>`

It will replace `None` with whatever it could resolve from the environment.  The problem is now the URL becomes:

    https://s3.amazonaws.com/%lt;Value%20Default:%20foo&gt;/

I have a fix - its more of a hack really at [this commit](https://github.com/burhan/cookiecutter-django/commit/c8ee217dd72ec29ccea4f683d83ca7438247461c) that will resolve the issue.

[This commit](https://github.com/burhan/cookiecutter-django/commit/4bf90c08ee07007a4d053c2a19499e1516777ba7) just updates the above to reflect the environment variable as per the documentation.

philippeluickx:
+1
I am having this issue when using https://github.com/Yaco-Sistemas/django-inplaceedit
> {% inplace_static %}

theskumar:
This should be fixed with us replacing django-configuration with django-environ now.

PR #204, opened 2015-03-24
Can't use AWS_STORAGE_BUCKET_NAME until post_setup.

jeffcjohnson:
With old code:
STATIC_URL: "https://s3.amazonaws.com/&lt;Value default: None&gt;/"

With new code:
STATIC_URL: "https://s3.amazonaws.com/mybucket/"

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -38,6 +38,7 @@ Andy Rose
 Andrew Mikhnevich / @zcho
 Kevin Ndung'u / @kevgathuku
 Kaveh / @ka7eh
+Jeff Johnson / @jeffcjohnson
 
 * Possesses commit rights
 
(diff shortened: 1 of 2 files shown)


jeffcjohnson:
This solved part of my problem but set STATIC_URL too late for at least one library I was using.  I'm closing this until I figure out more.

jeffcjohnson:
Replacing the old code with the following seems to do the trick.  I'm not sure why django-configuration Value is doing late or lazy evaluation of the variable but it isn't working well for me.

STATIC_URL = 'https://s3.amazonaws.com/%s/' % os.environ.get('DJANGO_AWS_STORAGE_BUCKET_NAME')

Does anyone have any advice here?

jeffcjohnson:
This is a duplicate of #121.

luzfcb:
@jeffcjohnson You could test with this:
https://github.com/pydanny/cookiecutter-django/blob/24f544083e61db34c457ae18abe58c19041ec1af/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.project_name%7D%7D/config/settings.py#L362-L368

and/or this:

https://github.com/pydanny/cookiecutter-django/pull/201/files

PR #212, opened 2015-04-10
Swap to django-storages-redux, an py3 compatible fork

### LINKED ISSUE #161: support for django-storages / django AWS S3 media store on python3
Hello there,

I've been watching this repo for a while and you guys are doing a great job! :smile: 

What you think about this idea:
I've followed this post (https://ashokfernandez.wordpress.com/2014/03/11/deploying-a-django-app-to-amazon-aws-with-nginx-gunicorn-git/) and now I'm able to create as many ec2 instances as I want with one single command and make my project use RDS and S3. 

The problem is: this setup relies on djang-storages which is not supporting py3 yet. That's why I'm facing problems to upload files into S3. 

My point is: do you consider using S3 in this setup? The post I mentioned is a great starting point. And, if so, I'd be very insterested in a solution that works in a py3 env.

Great project!

caffodian:
This is a bit silly (LOC-wise) but I saw #161, and since I've been doing this in some of my own projects, and thought this might be worth sharing.

[django-storages](https://pypi.python.org/pypi/django-storages) hasn't been updated in over 2 years at this point.  Around December a work [django-storages-redux](https://pypi.python.org/pypi/django-storages-redux/1.2.3) was started to fix various outstanding issues as well as to make it Python 3 compatible.  Not working with Py3 and S3 myself, I can't speak to whether this fixes #161 for free or not, but I have found value in the extra 2 years of bugfix commits ;)  

thanks!

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -38,6 +38,7 @@ Andy Rose
 Andrew Mikhnevich / @zcho
 Kevin Ndung'u / @kevgathuku
 Kaveh / @ka7eh
+Alex Tsai / @caffodian
 
 * Possesses commit rights
 
--- a/{{cookiecutter.repo_name}}/requirements/production.txt
+++ b/{{cookiecutter.repo_name}}/requirements/production.txt
@@ -3,7 +3,7 @@
 -r base.txt
 
 gunicorn==19.3.0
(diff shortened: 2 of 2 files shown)


theskumar:
This looks great! Happy that we have one more line of code ;)

PR #217, opened 2015-04-20
Fixed a bug with escaped spaces in Cache-Control.

ka7eh:
It seems there's a bug in boto that causes spaces in Cache-Control get escaped. (see: https://github.com/boto/boto/issues/2536)
For me, the header for static files rendered as this: 'max-age=604800,%20s-maxage=604800,%20must-revalidate' (under python 3 and django 1.7.7)
Turning the string into bytestring solved the issue for me, though I haven't tested it under other versions of python/django.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/production.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/config/production.py
@@ -70,8 +70,9 @@ INSTALLED_APPS = ('collectfast', ) + INSTALLED_APPS
 # AWS cache settings, don't change unless you know what you're doing:
 AWS_EXPIRY = 60 * 60 * 24 * 7
 AWS_HEADERS = {
-    'Cache-Control': 'max-age=%d, s-maxage=%d, must-revalidate' % (
-        AWS_EXPIRY, AWS_EXPIRY)
+    'Cache-Control': str.encode(
+        'max-age=%d, s-maxage=%d, must-revalidate' % (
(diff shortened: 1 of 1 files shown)


theskumar:
This is strange! Now are are using django-storage-reflux[1][2], just because it has python 3 support for S3. Please confirm that you are facing the issue with django-storage-reflux or not.

Also, if possible, please also an open an issue at https://github.com/jschneier/django-storages/issues. And you can add link to that issue alongside your fix.

__Footnotes__
[1] https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/requirements/production.txt#L6
[2] https://github.com/jschneier/django-storages

ka7eh:
I'm using django-storage-redux. I opened this ticket on the repo:
https://github.com/jschneier/django-storages/issues/47
If you want to merge this, we should add a note so we revert it after the issue is solved either in boto or django-storage-redux.

theskumar:
Thanks, please add the note with link to your issue in your commit and i'm happy to merge this in.

ka7eh:
Updates the PR. Good to go!

theskumar:
Thanks! Merging in.

theskumar:
@ka7eh there seemed to be issue while using `str.encode` in python2. I've used `six.b` to resolve the python2/3 compatibility. 

```
if PY3:
    def b(s):
        return s.encode("latin-1")
else:
    def b(s):
        return s
```

ka7eh:
Glad you caught that. As I mentioned, I hadn't tested it on other versions. Thanks

PR #233, opened 2015-05-07
Update common.py postgres connection string

schinkelg:
By removing localhost we enable the default use of Unix sockets which integrates better with the postgresql access rights in pg_hba.conf.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -9,5 +9,6 @@
     "timezone": "UTC",
     "now": "2015/01/13",
     "year": "{{ cookiecutter.now[:4] }}",
-    "use_whitenoise": "y"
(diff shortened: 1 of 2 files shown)


pydanny:
Will this work on Windows?

schinkelg:
Can't test this now, but I'm guessing it won't. But using localhost both Windows and *nix users need to change the connection string either way? Perhaps a commented example string would be better?

pydanny:
An `if/else/endif` statement would work: {% if cookiecutter.windows == 'y' %}localhost{% else % }{% endif %}

This requires a change to `cookiecutter.json`. The default for a `cookiecutter.windows` setting should be `n`.

PR #238, opened 2015-05-18
Docker Compose development environment

martinblech:
Vagrant support was dropped in favor of Docker based development environments, but no replacement was provided. This should do.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -33,7 +33,7 @@ Chris Franklin
 Benjamin Abel
 Felipe Arruda / @arruda
 Matt Warren / @mfwarren
-Martin Blech
+Martin Blech / @martinblech
 Andy Rose
 Andrew Mikhnevich / @zcho
 Kevin Ndung'u / @kevgathuku
--- a/README.rst
+++ b/README.rst
@@ -149,6 +149,22 @@ To get live reloading to work you'll probably need to install an `appropriate br
 
 .. _appropriate browser extension: http://feedback.livereload.com/knowledgebase/articles/86242-how-do-i-install-and-use-the-browser-extensions-
 
+**Running in an isolated Docker development environment**
+
+If you would rather run your application inside a separate Docker environment and avoid installing Postgres or any dependencies in your host OS, you can use the included Docker Compose configuration (``Dockerfile`` and ``docker-compose.yml``).
+
+Make sure that Docker_ is installed. In the project root run::
+
+    $ docker-compose up
+
+.. _Docker: https://docs.docker.com/installation/
+
+This command will download all the necessary dependencies and start your project. The first time, you will also need to run the following command to initialize the database::
+
+    $ docker-compose web python manage.py migrate
+
+If you're on OS X or Windows, run ``boot2docker ip`` to find out Docker's IP, otherwise it's localhost. Hit ``http://<docker_host>:8000/`` with your browser and you're ready to go.
+
 It's time to write the code!!!
 
 For Readers of Two Scoops of Django 1.8
--- a/{{cookiecutter.repo_name}}/Dockerfile
+++ b/{{cookiecutter.repo_name}}/Dockerfile
@@ -1,0 +1,10 @@
+# Only intended for development purposes.
+# For production deployment options, see the Deployment section in
+# README.rst
+
+FROM python:3.4
+ENV PYTHONUNBUFFERED 1
+RUN mkdir /code
+WORKDIR /code
+ADD . /code/
+RUN pip install -r requirements/local.txt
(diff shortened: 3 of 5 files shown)


luzfcb:
Hello @martinblech, that sounds good.

What will I speak, it is only my opinion, and can not be the same as the rest of the team.

First, I'm still trying to understand how the Docker ecosystem works.

having said that, I have some questions: 

1. All the structure of this project, assumes he is running on Ubuntu 14.04 Server/Desktop (Ok, it works almost unmodified on other Linux systems). This docker setting is based on what operating system?

2. Some libraries from Python dependencies, have C extensions that need to be compiled at installation time. The compile dependencies to build these python libraries in Ubuntu 14.04 are listed on [requirements.apt](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/requirements.apt) file. How this would work in Docker?

3. If someone see this docker configuration and want to put in a production environment. From this configuration uses some images available on [Docker Hub](https://registry.hub.docker.com/search?q=library), to what extent I can trust in the security settings applied in these images (server hardening)? The Database password is auto-generated? The database is accessible outside the container?

I think some of the answers to these questions should be included in the documentation to avoid problems and wrong interpretations.
Every day that passes, I think that "Explicit is better than implicit" is really an incredible excellent idea.

martinblech:
Hello @luzfcb, thank you for your feedback.

1. It's supposed to be OS-independent, but to be honest I have only tried it in OS X via `boot2docker`. I wouldn't expect there to be any problems, as Docker images are supposed to be completely isolated from the host OS. Do you have a Ubuntu or Windows box at reach where you could try it?
2. All the dependencies build fine with `pip install -r requirements/local.txt` during the image build stage. It seems that the base `python:3.4` image has all the compile dependencies.
3. `Dockerfile` and `docker-compose.yml` are only intended for development purposes. Perhaps a comment there stating so is enough?

Couldn't agree more about "explicit is better than implicit".

martinblech:
@luzfcb Please look at the comments I just amended to `Dockerfile` and `docker-compose.yml`.

luzfcb:
@martinblech about 1, docker python3.4 is based on Debian Jessie

in order of dependency:
1. https://registry.hub.docker.com/_/python/
1.1. https://github.com/docker-library/python/blob/a467ae88510bf64dd4f35253d9a41af3b3cef7b6/3.4/Dockerfile#L1
2. https://registry.hub.docker.com/_/buildpack-deps/
2.1 https://github.com/docker-library/buildpack-deps/blob/44070019e0c6c4f83461f2058d110cc418da09c4/jessie/Dockerfile#L1
2.2 https://github.com/docker-library/buildpack-deps/blob/a0a59c61102e8b079d568db69368fb89421f75f2/jessie/scm/Dockerfile#L1
2.3 https://github.com/docker-library/buildpack-deps/blob/a0a59c61102e8b079d568db69368fb89421f75f2/jessie/curl/Dockerfile#L1
3. https://registry.hub.docker.com/_/debian/
3.1 https://github.com/tianon/docker-brew-debian/blob/b637abe5e26f8b283a28d63cdc0413926ed54596/jessie/Dockerfile#L1
4. https://registry.hub.docker.com/_/scratch/

Interesting see how the layers are assembled in a container docker ...

jayfk:
While having a development environment powered by compose is neat, I'd go one step further and add a configuration that works for dev and prod.

I'd suggest the following:

- A configuration for production with four services: `postgres`, `memcached`, `gunicorn` and `nginx` as a reverse proxy. If celery support is enabled, add three more: `rabbitmq`, `celery-worker` and `celery-beat`.
- A configuration for development with `postgres` and a `django` image that runs `runserver_plus` and mounts the code to the running container, so auto reloading etc. works.

As of compose `0.3.0` there's no good solution to inherit from a common configuration like `cookiecutter-django` does with `common.py`, so two files are needed.


- A `docker-compose.yml` for production, so all is needed is to run `docker-compose up -d` on the prod server.
- A `dev.yml` for development. There are two ways to run that:
 - Set the environment variable `COMPOSE_FILE` to `dev.yml` and run `docker-compose up`, **or**
 - Run `docker-compose -f dev.yml up`

In fact, I've ported a couple of projects that were started with `cookiecutter-django` to a working configuration with `docker-compose`.

If there's interest in that, I'd be happy to put something together over the next few days.

martinblech:
The `extends` keyword can be used for sharing common configurations between development and production `.yml` files. Here's an example from the Docker Compose docs: https://docs.docker.com/compose/extends/

jayfk:
Yes, but it has some limitations that make it pretty unusable. What we really need is a concept like include from [#318](https://github.com/docker/compose/issues/318) or `global_extends` from [#1380](https://github.com/docker/compose/issues/1380).

jayfk:
I made a PR here https://github.com/pydanny/cookiecutter-django/pull/284

It is basically working like @martinblech's contribution for development with a few tweaks (env variables are set automatically in `entrypoint.sh`, for example) **and** a working configuration that can be used as a starting point to run it in production.

pydanny:
Closing as #284 encompasses this excellent work.

PR #248, opened 2015-06-22
Include LOCAL_APPS before THIRD_PARTY_APPS in INSTALLED_APPS

narayanaditya95:
This allows custom User model to be registered before THIRD_PARTY_APPS are imported.
THIRD_PARTY_APPS can now get the model registered with AUTH_USER_MODEL without throwing a LookupError.

This particular error was seen when using Django-CMS with this cookiecutter.

### DIFF
--- a/{{cookiecutter.repo_name}}/config/settings/common.py
+++ b/{{cookiecutter.repo_name}}/config/settings/common.py
@@ -48,7 +48,7 @@ LOCAL_APPS = (
 )
 
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#installed-apps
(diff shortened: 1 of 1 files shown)


pydanny:
If I remember correctly, I set things up the way they are because the order that they are in works just fine 95% of the time. In fact, it breaks for some packages when the order is switched to the way you designate. Because your fix breaks other packages and is just for the django-cms edge case, I'm going to unfortunately have to reject this PR.

pydanny:
Have you considered creating your own fork of cookiecutter-django? If you do so, let me know and as long as you maintain it we'll list it in the README.

narayanaditya95:
@pydanny Nice idea. I will maintain a fork with django-cms configured settings. I will post on this thread when the fork is good enough.

PR #250, opened 2015-06-23
Fedora Compatiblity

burhan:
For those of us using Fedora/RH - I am adding OS-install scripts for Fedora compatibility; and renamed the `install_os_dependencies` to `install_ubuntu_dependencies`; also added `requirements.yum` in the same spirit as `requirements.apt`.


{{cookiecutter.repo_name}}/requirements.yum
```diff
@@ -0,0 +1,24 @@
+sic build dependencies on Fedora/RedHat
```
luzfcb: missing `#` comment here

pydanny:
Needs documentation in the interior project README. Please add under Heroku and Dokku sections. Reference: https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/README.rst#deployment

{{cookiecutter.repo_name}}/install_fedora_dependencies.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+# Read the requirements.apt file, and remove comments and blank lines
+function list_packages(){
+     cat ${OS_REQUIREMENTS_FILENAME} | grep -v "#" | grep -v "^$";
+}
+
+function install()
+{
+    list_packages | xargs yum -y install;
```
luzfcb: `yum install` command must explicitly not update a package if it is already installed

luzfcb: what is the `yum` parameter that is equivalent to `--no-upgrade` from `apt-get` like `apt-get --no-upgrade install`

luzfcb:
@burhan thanks for your contribution.

having made the comments, I am -1 to accept it because:

- This project officially only supports ubuntu (ok, works perfectly on other systems without or almost without modification).
- I would like to avoid adding another script
- and I at least I can not test because  I have no machine with fedora or other redhat distribution and have no experience with them

the options I see are:
1 - Put it in a gist.github.com and points it in the documentation.
2 - Improve install_os_dependencies.sh to make it read requirements.yum. For this it is necessary to include some logic to detect which the distribution is currently in use. I do not know a way foolproof to do this.

@pydanny what your thoughts about it?

luzfcb:
option 2: 
perhaps something like
`python -c "import platform;print(platform.platform().lower())"`

in my machine result in:
`linux-3.13.0-34-generic-x86_64-with-ubuntu-14.04-trusty`

@burhan which is the result of running the above command in his fedora?

pydanny:
As this change doesn't break any of the existing architecture and broadens the potential user base, I'm as concerned as @luzfcb. Here are my reasons:

1. Lack of ability to test. @burhan, can you set up a test harness the way we have with Ubuntu?
2. Additional scripts can be messy. Perhaps a dedicated `scripts` directory would mitigate this issue.
3. Support. I simply don't know Fedora, neither does @luzfcb and I doubt @theskumar has experience either. We won't be able to provide support for the inevitable help requests, which from experience I know will anger our users. 

Item 3 on the list is the really hard one for us. I already get grief on a number of my projects because I can't always provide timely support for an unpaid effort on a regular basis. Managing details for a new operating system is especially challenging because none of us have the bandwidth to support it. The only way around this is to get decent sponsorship from Redhat, which isn't likely (I've yet to have a single project sponsored and I doubt it's ever going to happen).

Perhaps I'm missing an issue with item 3. I'm open to suggestions.

theskumar:
I did started with Fedora/Redhat at one point of time. But I no longer keep myself upto date with it. I agree on the points said by @pydanny and @luzfcb. 

I would emphasise on documenting the devops or machine preparation stuff independent of Django scaffolding. We should add a link to available resources here. There is no point adding everything to a single repository. 

That said, i'm happy to have Heroku support in built primarily because of the convenience a developer get to have a quick testable server up an running in minutes and LOC that need to be maintained.

{{cookiecutter.repo_name}}/install_fedora_dependencies.sh
dupuy: Per the very useful Arch Linux Wiki page [Pacman Rosetta](https://wiki.archlinux.org/index.php/Pacman_Rosetta) that gives equivalent commands for different package management commands from various distributions/toolchains, `yum install` will not update a package that is out of date - `yum update` is required for that (and is used below in the `install_or_upgrade()` function).

burhan:
Hello everyone; sorry for the late reply:

The motivation behind this is that we use RH at work (mandated distribution) and thus our test bed is RH and having a template like this works great.

I understand that the support would be problematic, and now that I think about it, I do not want this to turn into a "lets support everything"; otherwise it would be - as correctly observed a maintenance and support headache.

Therefore, I would like to close this request and put it up as a gist and maintain it as a separate branch on my fork.

Issue #221, opened 2015-04-24
Replace memcached with redis as cache


theskumar:
Simplifies overall things and compatibility. See https://github.com/pydanny/cookiecutter-django/issues/135 for more details.

PR #258, opened 2015-07-06
adding redis support

burhan:
Since I too am not happy with memcached, I already had a fork with redis support. Seeing as there was an issue (#221) already raised, here is a pull request to add redis support.

### DIFF
--- a/{{cookiecutter.repo_name}}/requirements/base.txt
+++ b/{{cookiecutter.repo_name}}/requirements/base.txt
@@ -34,4 +34,8 @@ django-autoslug==1.8.0
 # Time zones support
 pytz==2015.4
 
+# Redis support
+django-redis==4.2.0
+redis>=2.10.0
+
 # Your custom requirements go here
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -45,6 +45,7 @@ Adam Dobrawy / @ad-m
 Daniele Tricoli / @eriol
 Harry Percival / @hjwp
 Cullen Rhodes / @c-rhodes
+Burhan Khalid / @burhan
 Audrey Roy Greenfeld / @audreyr (and creator/maintainer of cookiecutter) * 
 
 * Possesses commit rights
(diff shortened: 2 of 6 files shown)


{{cookiecutter.repo_name}}/requirements/base.txt
```diff
@@ -34,4 +34,8 @@ django-autoslug==1.8.0
 # Time zones support
 pytz==2015.4
 
+# Redis support
```
pydanny: Why isn't this in ``{{cookiecutter.repo_name}}/requirements.txt``?

burhan: I put it here because I did not want to have requirements defined in two places. Previously the only requirement at `/requirements.txt` had to do with memcache, and the rest was just a proxy to the actual requirements (as per the comment in the file).

saurabh-fueled: @pydanny  Memcache requirements were in `/requirements.txt` because heroku needs `pylibmc` at root requirements.txt to install some OS libraries and also `django-heroku-memcachify` was only relevant in context of heroku.

theskumar: @pydanny  Memcache requirements were in `/requirements.txt` because heroku needs `pylibmc` at root requirements.txt to install some OS libraries and also `django-heroku-memcachify` was only relevant in context of heroku.

theskumar:
Are we not missing on something w.r.t ubuntu setup?

@pydanny Also, it's a good we start to think about separating out, platform specific setups to a different repo and while supporting on 1 or 2(at max.) in the core cookiecutter.

Currently we have Heroku, dokku, Docker and Ubuntu and we had a PR for Fedora.

burhan:
I recommend adding them as `feature/` branches; and then giving people access to maintain those branches (ie, sponsors for certain platforms).

My personal order of preference would be:

1. Docker (with docker-compose) - as this is most agnostic and has widest support (osx, linux, windows, Amazon, GAE, etc.)
2. Ubuntu
3. Heroku

The others, based on interest/support from the community. I personally use the following; and can support these:

1. AWS BeanStalk
2. Docker on AWS
3. Oracle
4. RHEL

theskumar:
It would worth to move the conversation to a different Github issue ( I've created one here: #260 ). Let's not mix conversation as we would like this PR to get closed asap.

pydanny:
Any reason why we shouldn't merge this in? LGTM.

theskumar:
LGTM as well :+1:

PR #264, opened 2015-07-07
Use maildump instead of ConsoleEmailHandler

burhan:
In development, the `ConsoleEmailHandler` is of limited use if you are sending emails with attachments, or any HTML message.

I propose integrating with [maildump](https://github.com/ThiefMaster/maildump) a Python clone of [mailcatcher](http://mailcatcher.me/) which is a ruby application.

It provides a local SMTP server listening on 1025, and an interface that shows multi-part emails and HTML formatted emails correctly, available at http://127.0.0.1:1080:

image

The way I use it is through two grunt tasks, `start-email-server` and `stop-email-server`.

### DIFF
--- a/{{cookiecutter.repo_name}}/README.rst
+++ b/{{cookiecutter.repo_name}}/README.rst
@@ -126,7 +126,7 @@ To run a celery worker:
 
 Please note: For Celerys import magic to work, it is important *where* the celery commands are run. If you are in the same folder with *manage.py*, you should be right.
 {% endif %}
-
+{% if cookiecutter.use_maildump == "y" %}
 Email Server
 ^^^^^^^^^^^^
 
@@ -147,6 +147,7 @@ To stop the email server::
     $ grunt stop-email-server
 
 The email server listens on 127.0.0.1:1025
+{% endif %}
 
 It's time to write the code!!!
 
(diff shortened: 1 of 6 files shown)


pydanny:
Do emails still send if `start-email-server` is not yet started?

burhan:
No, you get an exception "Connection refused".

pydanny:
@burhan Could you make this an available option during project creation?

burhan:
@pydanny - done.

{{cookiecutter.repo_name}}/requirements/local.txt
```diff
@@ -10,3 +10,8 @@ django-debug-toolbar==1.3.2
 
 # improved REPL
 ipdb==0.8.1
+
+{% if cookiecutter.use_maildump == "y" -%}
+# maildump
```
theskumar: Shall we call this comment as `# Email Debugging`

{{cookiecutter.repo_name}}/README.rst
```diff
@@ -126,6 +126,28 @@ To run a celery worker:
...
+
+This will start an email server. The project is setup to deliver to the email server by default. To view messages
+that are sent by your application, open your browser to http://127.0.0.1:1080
+
+To stop the email server::
+
+    $ grunt stop-email-server
+
+The email server listens on 127.0.0.1:1025
+{% endif %}
```
theskumar: +1

{{cookiecutter.repo_name}}/requirements/local.txt
burhan: Agreed, made it the same as the master README.

PR #274, opened 2015-07-15
Refactor choice of settings module out of taskapp/celery.py

mwdiers:
…s module. This makes it necessary to edit `taskapp/celery.py` after deployment, which is not ideal.

This moves the choice of a settings module to config/celery_config.py, where it belongs.

### DIFF
--- a/{{cookiecutter.repo_name}}/config/celery_config.py
+++ b/{{cookiecutter.repo_name}}/config/celery_config.py
@@ -1,0 +1,4 @@
+# This tells Celery which Django settings module it should use.
+# This decouples the taskapp configuration from a specific instance of the project.
+
+DJANGO_SETTINGS_MODULE = 'config.settings.local'
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/taskapp/celery.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/taskapp/celery.py
@@ -4,10 +4,11 @@ import os
 from celery import Celery
 from django.apps import AppConfig
 from django.conf import settings
+from config import celery_config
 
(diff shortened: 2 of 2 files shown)


{{cookiecutter.repo_name}}/config/celery_config.py
```diff
@@ -0,0 +1,4 @@
+# This tells Celery which Django settings module it should use.
+# This decouples the taskapp configuration from a specific instance of the project.
+
+DJANGO_SETTINGS_MODULE = 'config.settings.local'
```
pydanny: Is this so that the celery tasks can run under different settings modules? In other words, we could change this value to 
`DJANGO_SETTINGS_MODULE = 'config.settings.celery_settings'`?

If that's the case, I'm not certain why this extra module is needed. Is it because of the inherent magic of Celery?

mwdiers:
No, this will not work, because since 3.x, Celery is loosely coupled to Django. When the celeryd daemon process launches, it is not running from within an initialized Django environment. Therefore it does not have access to an initialized settings object.

Take a look at `taskapp/celery.py` . You will notice the lines:

    if not settings.configured:
        os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'config.settings.local')

These lines only run if `taskapp.celery` is imported from celeryd. When celeryd is started, it must be told which settings module to use, in much the same way that django-admin.py must be told. But setting the module directly in the celery.py module forces us to edit the module file at deployment to point to the correct settings file. Of course, that's a really bad idea.

 So really this is needed due to a *lack of* Celery magic.

pydanny:
Hey @mwdiers, we're going with the other technique for now. If it turns out to be cumbersome or problematic, we'll try your option.

Issue #282, opened 2015-07-19
Add .pylintrc and .pep8


kaidokert:
after creating the project, i tried to run `prospector -s veryhigh`  ( pip install prospector ) which runs pylint, pep8 and a bunch of other tools.
Had to add .pylintrc and .pep8 to match the project conventions, here is what i got in `.pep8`
> [pep8]
> max-line-length = 120
> exclude=*/migrations/*

and `.pylintrc`
> [MASTER]
> ignore=migrations
> [FORMAT]
> max-line-length=120
> [MESSAGES CONTROL]
> disable=missing-docstring,invalid-name
> [DESIGN]
> max-parents=13

I would probably add disable=no-init,too-few-public-methods too, to eliminate most of the false positives.

kaidokert:
Actually i found prospector was not loading all the plugins properly, see https://github.com/landscapeio/prospector/pull/134

Once this is addressed, the following `.pylintrc` is sufficient, and results in a  clean check when run with `prospector -F -s veryhigh`

> [MASTER]
> load-plugins=pylint_common,pylint_django,pylint_celery
> [FORMAT]
> max-line-length=120
> [MESSAGES CONTROL]
> disable=missing-docstring,invalid-name,locally-disabled
> [DESIGN]
> max-parents=13

pydanny:
I can do this unless you want to open a pull request and get credit?

kaidokert:
I have a PR open here https://github.com/pydanny/cookiecutter-django/pull/283

audreyr:
Merged. Thanks much @kaidokert.

PR #283, opened 2015-07-20
Add .pep8 and .pylintrc, for #282

kaidokert:
Here is what the defaults would look like to make the generated project pass most of the strict checking.
With these .rc files, prospector shows only 8 violations currently, all minor and easy to clean up.

### DIFF
--- a/{{cookiecutter.repo_name}}/config/settings/common.py
+++ b/{{cookiecutter.repo_name}}/config/settings/common.py
@@ -256,11 +256,14 @@ LOGGING = {
     }
 }
 {% if cookiecutter.use_celery == "y" %}
-########## CELERY
+# ######### CELERY
 INSTALLED_APPS += ('{{cookiecutter.repo_name}}.taskapp.celery.CeleryConfig',)
 # if you are not using the django database broker (e.g. rabbitmq, redis, memcached), you can remove the next line.
 INSTALLED_APPS += ('kombu.transport.django',)
 BROKER_URL = env("CELERY_BROKER_URL", default='django://')
-########## END CELERY
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_TASK_SERIALIZER = 'json'
+CELERY_RESULT_SERIALIZER = 'json'
+# ######### END CELERY
 {% endif %}
 # Your common stuff: Below this line define 3rd party library settings
--- a/{{cookiecutter.repo_name}}/config/settings/common.py
+++ b/{{cookiecutter.repo_name}}/config/settings/common.py
@@ -256,14 +256,11 @@ LOGGING = {
     }
 }
 {% if cookiecutter.use_celery == "y" %}
-# ######### CELERY
+########## CELERY
 INSTALLED_APPS += ('{{cookiecutter.repo_name}}.taskapp.celery.CeleryConfig',)
 # if you are not using the django database broker (e.g. rabbitmq, redis, memcached), you can remove the next line.
 INSTALLED_APPS += ('kombu.transport.django',)
 BROKER_URL = env("CELERY_BROKER_URL", default='django://')
-CELERY_ACCEPT_CONTENT = ['json']
-CELERY_TASK_SERIALIZER = 'json'
-CELERY_RESULT_SERIALIZER = 'json'
-# ######### END CELERY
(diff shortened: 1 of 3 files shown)


theskumar:
Are we supporting prospector in any ways. These seem to be me duplicate configurations and will also add blot to project. @pydanny @luzfcb ?

I'm happy adding things to `setup.cfg`, where flake8 rules are already added[1]. 

[1] https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/setup.cfg#L1-L4

kaidokert:
The two files are not specifically for prospector, they are for pep8 and pylint respectively, in their default locations ( prospector is just a convenient wrapper to run a long suite of of the linting tools easily, but is not related to these config files )
Maybe a configuration-time question to which linting tools to pre-configure, to avoid bloat ? "use linting:" and a choice of flake8, pylint and pep8 ?

pydanny:
@theskumar Normally I'm against any kind of duplication, but since this one is designed for coders using a set tools to check code-smell, it might be worth including. 

As for including a question about which files to include at generation time, I would be all for that BUT windows breaks when you use logic in `cookiecutter` template or directory names. :unamused:

luzfcb:
I not use prospector directly, but only via landscape.io
I still prefer flake8.
the configuration I'm using now is https://github.com/luzfcb/django_documentos/blob/master/setup.cfg

and to landscape.io ( prospector ) https://github.com/luzfcb/django_documentos/blob/master/.landscape.yml

{{cookiecutter.repo_name}}/config/settings/common.py
```diff
@@ -256,11 +256,14 @@
...
-########## CELERY
+# ######### CELERY
 INSTALLED_APPS += ('{{cookiecutter.repo_name}}.taskapp.celery.CeleryConfig',)
 # if you are not using the django database broker (e.g. rabbitmq, redis, memcached), you can remove the next line.
 INSTALLED_APPS += ('kombu.transport.django',)
 BROKER_URL = env("CELERY_BROKER_URL", default='django://')
-########## END CELERY
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_TASK_SERIALIZER = 'json'
+CELERY_RESULT_SERIALIZER = 'json'
```
theskumar: Please remove this part, you may like to have it in a different PR.

kaidokert: oops, done. Not sure how to keep two PRs open against the same branch in parallel

pydanny:
Any strong opinions about this one? Just want to know if we should merge or not. I lean yes, but want input from @luzfcb and @theskumar.

audreyr:
+1 on this. Even with the overlap, I like the ability to run more linting tools. :shipit:

But before merging, we ought to document this in a `docs/linters.rst` file. (We should document our current flake8 setup as well, but that's a separate issue)

audreyr:
I merged this and then:
* Moved the rules from `.pep8` into `setup.cfg`. This didn't affect the workings of pep8 or prospector.
* Documented it in `docs/linters.rst` (feel free to improve this)

Anyone know if/how it's possible to put pylint rules into setup.cfg as well? @theskumar? I left the pylint rules in `.pylintrc` because I couldn't find any info on this.

luzfcb:
@audreyr apparently, the idea was rejected: https://bitbucket.org/logilab/pylint/issues/617/search-pylint-configuration-in-setupcfg

audreyr:
@luzfcb Oh, that's too bad. I guess we'll stick with having a `.pylintrc` then.

luzfcb:
i suppose that flake8 wrap all pylint configuration, then .pylintrc may be
not necessary.

@luzfcb <https://github.com/luzfcb> Oh, that's too bad. I guess we'll stick
with having a .pylintrc then.

—
Reply to this email directly or view it on GitHub
<https://github.com/pydanny/cookiecutter-django/pull/283#issuecomment-143502752>
.

audreyr:
The additional .pylintrc gives us the ability to:
* Run prospector
* Run pylint

Which means we get a lot more visibility into code quality. I think it's alright to have another dotfile in the project.

Issue #260, opened 2015-07-07
Separate/isolate per platform setup/deployment to different repositories


theskumar:
Continuing from here: https://github.com/pydanny/cookiecutter-django/pull/258#issuecomment-119078354

We should start to think about separating out, platform specific setups to a different repo and while supporting on 1 or 2(at max.) in the core cookiecutter.

Currently we have Heroku, dokku, Docker and Ubuntu and we had a PR for Fedora.

theskumar:
Quoting @burhan from https://github.com/pydanny/cookiecutter-django/pull/258#issuecomment-119169813 
> I recommend adding them as `feature/` branches; and then giving people access to maintain those branches (ie, sponsors for certain platforms).
> 
> My personal order of preference would be:
> 
> 1. Docker (with docker-compose) - as this is most agnostic and has widest support (osx, linux, windows, Amazon, GAE, etc.)
> 2. Ubuntu
> 3. Heroku
> 
> The others, based on interest/support from the community. I personally use the following; and can support these:
> 
> 1. AWS BeanStalk
> 2. Docker on AWS
> 3. Oracle
> 4. RHEL

burhan:
I recommend adding them as feature/ branches; and then giving people access to maintain those branches (ie, sponsors for certain platforms).

My personal order of preference would be:

1. Docker (with docker-compose) - as this is most agnostic and has widest support (osx, linux, windows, Amazon, GAE, etc.)
2. Ubuntu
3. Heroku

The others, based on interest/support from the community. I personally use the following; and can support these:

1. AWS BeanStalk
2. Docker on AWS
3. Oracle
4. RHEL

jpatel3:
+1 AWS BeanStalk

pydanny:
In my experience vendors won't provide support here. Sure, they might do a drive-by pull request, but they aren't going to maintain things. That falls upon us, the community and especially the core contributors. 

That's why we don't provide more options, Redhat, Amazon, et al, they just aren't going to help us maintain tools to access their systems.

pydanny:
As much as I want to support separate repos, I don't think its feasible. None of us have the time to execute this feature.

PR #284, opened 2015-07-21
Enhanced Docker Support using docker-compose

jayfk:
There are two environments, one for local development and one for production use.

**dev.yml** contains the configuration for local development, consisting of two services:

- A service called `django` that basically just runs the `runserver` command.
- postgres

In development, the code folder is mounted into the container so that `runserver`s auto code reloading works. As of now, I had no time to take a look at the code reloading using node.

**docker-compose.yml** is for production use, consisting of 4 or 7 services if celery is installed:

- Nginx as a reverse proxy
- Gunicorn as an upstream server for nginx running the django app
- postgres
- Redis 
- RabbitMQ as celery broker (optional)
- Celery Worker (optional)
- Celery Beat (optional)

Environment variables for production are defined using an environment file called `env.production`.

I'm currently working on a fork that goes beyond the scope of this PR here: https://github.com/jayfk/cookiecutter-django-docker 

The idea is to experiment with all docker and docker-compose features in a seperate project and to contribute to upstream as soon as the changes are mature enough. In case someone is interested, join me there.


pydanny:
Thanks for all this hard work. Would it be possible to sketch out some basic documentation of usage of Docker like what we have for Heroku and Dokku? It would go in the deployment section: https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/README.rst#deployment

burhan:
Amazing work, I was working on something similar.  I'll help with the documentation.

stepmr:
:+1: :+1: :+1: 

For reference, there's a super basic walkthrough at https://realpython.com/blog/python/django-development-with-docker-compose-and-machine/

martinblech:
This looks really good! It's great to have a working Docker configuration, for both development and production.

I like this PR overall, but I have a couple of comments:

1. Is `entrypoint.sh` really necessary? If it's only used to set environment variables, I would argue it's better to keep them in the compose `.yml` files. It's more explicit, there's fewer places to check if something goes wrong. Same for `gunicorn.sh`.
2. You could add `USER django` to the `Dockerfile` and avoid having to write `su -m django -c "<cmd>"` for every command. I think it's less error-prone this way.
3. The development configuration should support Celery too.

jayfk:
> Is entrypoint.sh really necessary? If it's only used to set environment variables, I would argue it's better to keep them in the compose .yml files. It's more explicit, there's fewer places to check if something goes wrong. Same for gunicorn.sh.

The purpose of `entrypoint.sh` is to make sure that the environment variables `cookiecutter` is using don't have to be set multiple times in different formats.

If you use the postgres container with a different user/password combination, you'd have to set the environment variables like this:

For the database:

    POSTGRES_PASSWORD=mysecretpass
    POSTGRES_USER=postgresuser

For Django:

    DATABASE_URL=postgres://postgresuser:mysecretpass@postgres:5432/postgresuser

I really didn't wan't to change too much about the current implementation so I created `entrypoint.sh` as a compat layer.

A pure docker solution would look like this: 

    DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql_psycopg2',
        'NAME': env("POSTGRES_ENV_POSTGRES_USER", default="postgres"),
        'USER': env("POSTGRES_ENV_POSTGRES_USER", default="postgres"),
        'PASSWORD': env("POSTGRES_ENV_POSTGRES_PASSWORD", default=""),
        'HOST': 'postgres',
        'PORT': 5432,
     }
  }

This could be added to the docs for people that prefer a cleaner solution.

> You could add USER django to the Dockerfile and avoid having to write su -m django -c "<cmd>" for every command. I think it's less error-prone this way.

That's a good point.  This can be put in the `Dockerfile` or the `.yml` file. The `.yml` might be the better place, because more people are looking at it. This allows everyone to see the the user at a quick glance. What do you think? 

> The development configuration should support Celery too.

I always like to keep the `dev` environment relatively simple, but that's a matter of taste.

pydanny:
If this works, and no one objects, I would like to merge it in. It doesn't have to be perfect, it just has to work. We can always refactor and improve on our Docker support as it is used.

burhan:
I would like to put my $0.02 and ask that this be merged; I am waiting to give this is a try as we exclusively deploy on docker in production.

theskumar:
@jayfk Thanks for this. I'm going to give it a try on my machine and comment accordingly (Maybe this weekend or meantime). I don't want to hold up, but I would like to try it out once before merging it in. 

There are couple of things, before that:

1. `env.production` – can we have sensible defaults hard-coded and then use `.env` (not tracked) to override or provide the values as needed. We would not like to track anything that stores credentials.
2. Is there any specific reason to `rabitmq` when we already have one `redis` instance.
3. There are some existing documentation for docker https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/docs/docker_ec2.rst , it needs to be cleaned up/modified. 
4. I think it would be very helpful, if we can improvise the documentation with relevant hyperlinks where applicable. We can/should also use wiki for documentation that might look noise in the codebase and provide a hyperlink to it.
5. When we say you can use docker setup in production, can we have documentation where someone would like to have redis/postgres running on a independent machines. Imagine a scenario of scaling up the stack against a load balancer, else we should just say use this docker setup in development and it's not for production explicitly. __I would not prefer someone get stuck while trying to scale horizontally in production, because they used docker as suggested here.__
6. Please rebase your branch on top of `master` branch of this repo, currently it has merge conflicts.  

PS: 
- That said, adding things to a single repository makes things monolithic & little hard to maintain, a separate discussion thread for it has already been initiated (https://github.com/pydanny/cookiecutter-django/issues/260), please help out with your feedback. 
- In the case of docker, looking at it's acceptance and demand, I think we should not have a major issue maintaining it in this project, but before merging I would be interested to know reasonable answer to point 5. 
- Like @pydanny code doesn't have to be perfect, but I would not like to compromise on documentation part. Please provide reasoning for anything done where needed as comments like in the case of `entrypoint.sh`.

Cheers,

jayfk:
1. What about a `.env.example` with sensible defaults (tracked) and `docker-compose.yml` linking to `.env` (not tracked). We could then add documentation that the file should be renamed *and* that it is wise to not track the file in version control.

2. No, other than that you have to set some additional configurations using redis: http://celery.readthedocs.org/en/latest/getting-started/brokers/redis.html

3+4. I totally agree with you. I'm currently working on more documentation

5.

Scaling to more than one host with `docker-compose` is currently being worked on at docker. They have to get the host to host network part in place, before they can introduce the functionality to `docker-compose`. 

Once this is merged in, it should work out of the box for the current setup. You'd spin up a swarm cluster and run the whole stack on multiple hosts. You may have to add a flag to your postgres service that no other service should run on the same host, but that's about it.

To your main concern: They'd not get stuck, it's just currently not supported out of the box. They'd have to use external tools for host to host networking like weave. There's a guide here: http://weave.works/guides/weave-and-docker-platform/chapter4/compose-scalable-swarm-cluster-with-weave.html

We could (and probably should) add this to the documentation.

jayfk:
Related issues:

https://github.com/docker/compose/issues/1803
https://github.com/docker/compose/pull/1676

theskumar:
@jayfk thanks for updates. There are a lot of new information coming, which are important too. Let's get things documented and we should be good. ;-)

1. yes. Please.
2. Let's go with additional configuration and use redis only for celery. If there are no other technical requirements/issues to rabitmq. This will make one less application to maintain across all the deployment setups.. e.g. Heroku, Ubuntu. (We have been using redis as broker in production, and didn't face any issue as such)

pydanny:
We should focus on focusing the local development version of Docker for cookiecutter-django. Once we get that looking really good, we can focus on production deployments. The advantage there is:

1. Narrows the scope of this pull request.
2. Allows us to see how it works for many users and adjust accordingly before providing an 'official' docker-deployment solution.

Also, I want to say @jayfk, this is great work and great dialogue.

theskumar:
@pydanny +1

luzfcb:
@pydanny +1 . I think that until we have, over time, adjusted everything to work well in production, it would be good to leave extremely explicit that this docker setting was not meant to be used in a production environment. Done that, I have no objection to include configuration for docker.

some random thoughts :smiley: : :

I still have my doubts about security, specifically on what is included via dependencies of Docker containers. eg.: https://github.com/pydanny/cookiecutter-django/pull/238#issuecomment-103665409

How can I be sure that any of the dependencies does not include any security breach caused by lack of knowledge about techniques of safety or security hole purposely included?.

I still have very little experience with docker, perhaps that question already has been answered somewhere

pydanny:
@luzfcb My thought has been to have docker for cookiecutter-django only for local development. For local development it shines, and security isn't that much of an issue.

As for production-readiness, I'm not keen on endorsing that until we've had Docker for local development for some time. In other words, it's a way to lower the barrier of entry for people on systems where installing our requirements is hard (like Windows users).

jayfk:
> I still have my doubts about security, specifically on what is included via dependencies of Docker containers. eg.: #238 (comment)

I'm always looking from the perspective of what we currently use to install software on our servers. While some people install from scratch and compile everything on their own, it's relatively safe to say that most people use package managers like apt or yum.

When using a package manager you trust the people maintaining the package to make the right decisions regarding dependencies, security and sane default settings. That's the same thing for docker images. You'll have to trust the people that have put the image together.

The cool thing for docker image is, you can go down the whole chain to see what they have done. If you look at the dockerfiles, you can see a chain of images they build up on and what they are doing on their part (mostly apt installing software and providing a default config that runs in the foreground).

By following the chain and looking at the individual docker and config files, and, more important what it runs at the end, I found it easier to to understand what the image does exactly.

In the end, it's just a new technology (or an old technology reinvented and tweaked) you have to get your head around at first to understand all the implications on the different layers involved.

Imagine someone in the 90's popping op in the linux mailing list, saying that he invented a great technology to install software. All one had to do is to run `apt-get install whatnot` and they are ready to run. I think the reactions would have been like today with docker. It's not secure, what if someone screws up with a config along the way etc.?

I think once a technology evolves and more and more people put their eyes on it most of these questions go away because people build trust in it.

This is not to defender docker, there are absolutely horrible examples for insecure docker images. If you look at the official django image that runs `runserver` by default, you get the idea.

That's why I think the ongoing discussion on this issue is a vital thing, because the end result is more quality.

jayfk:
Catched up with upstream.

Commit c7ea475 addresses all the points discussed here. It'd nice if a native speaker could take a quick look at the documentation.

pydanny:
LGTM.

I'm merging this in. As we use it we can fine tune things as needed. Excellent work @jayfk and many thanks to @martinblech for making the initial pass!

theskumar:
Awesome work! :+1: 

Kudos to @jayfk @martinblech and everyone else for making this happen. :blush:

Issue #279, opened 2015-07-17
Documentation is inaccurate for running celery


tarkatronic:
As noted in [the original pull request](https://github.com/pydanny/cookiecutter-django/pull/266#commitcomment-12102184), celery needs to be run with `celery -A proj.taskman` (now taskapp). However, the README states that you need to run `celery -A {{cookiecutter.repo_name}} worker -l info`. The command given in the README will just produce a very unhelpful, confusing error.

kaidokert:
I was about to pullrequest this .. and then i stepped through celery loader line by line and got even more confused. By default its looking for {{appname}}.celery as a fallback,  as per celery.app.utils.find_app()

>                    try:
>                        return find_app(
>                            '{0}.celery'.format(app),
>                            symbol_by_name=symbol_by_name, imp=imp,

That code is also looking for found = sym.app and found = sym.celery as an attribute of loaded module.

So if i rename taskapp->celery then it somehow works with `celery -A {{appname}} worker`, but completely does not pick up the correct config and starts looking for amqp:// default broker..
Running with `celery -A proj.taskapp` works for now though.
Would be awesome to know what the best practice actually should be.

PR #285, opened 2015-07-21
Reflect Celery taskapp refactoring in the docs

jayfk:
Fixes https://github.com/pydanny/cookiecutter-django/issues/279

### DIFF
--- a/{{cookiecutter.repo_name}}/README.rst
+++ b/{{cookiecutter.repo_name}}/README.rst
@@ -122,7 +122,7 @@ To run a celery worker:
 .. code-block:: bash
 
     cd {{cookiecutter.repo_name}}
-    celery -A {{cookiecutter.repo_name}} worker -l info
+    celery -A {{cookiecutter.repo_name}}.taskapp worker -l info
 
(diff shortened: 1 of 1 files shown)


pydanny:
LGTM. Anyone else?

jayfk:
@mwdiers might want to take a look on it. I'm still not entirely sure how celery import magic works in the background, but it should do for now.

mwdiers:
Hmm. This doesn't seem right to me. There is really no import magic going on that I am aware of. Celery will look for the named module in the PYTHONPATH, just like any other import.

If the project_root has been added to your PYTHONPATH, then your start command is just this:

```
celery -A taskapp worker -l info
```

mwdiers:
Perhaps I am misunderstanding Two Scoops recommendations, but I never add the repository root to my PYTHONPATH. Only the project root. Obviously, it one adds the repository_root instead, then the proposed doc change is correct.

pydanny:
@mwdiers, we aren't super-explicit in TSD18 where one should set the Python path. We do provide it as a tip for users who know their operating system well enough to get it right, but that's not everyone (for example, it's rather poorly documented for Windows users).

Tell you what, if this turns out not to be super-ideal, we'll refactor and improve the celery part(s) of cookiecutter-django. :smile:

PR #289, opened 2015-07-24
Added vagrant support with automatic provisioning of dependencies and PostgreSQL

javipalanca:
I've added a Vagrantfile that allows developers to automatically run a vagrant image (utopic64) wich is provisioned with all dependencies (os and python), creates a virtualenv and installs PostgreSQL with the default database for local development.
It also installs python3. I've updated libtiff4-dev to libtiff5-dev. I had to comment gevent since it is not python3 compatible.
Hope this helps!

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -51,6 +51,7 @@ Burhan Khalid / @burhan
 Jannis Gebauer / @got_nil
 jayfk / @jayfk
 stepmr / @stepmr 
+Javi Palanca / @javipalanca
 
 * Possesses commit rights
 
--- a/{{cookiecutter.repo_name}}/.gitignore
+++ b/{{cookiecutter.repo_name}}/.gitignore
@@ -54,3 +54,6 @@ node_modules/
 
 # virtual environments
 .env
+
+# Vagrant
+.vagrant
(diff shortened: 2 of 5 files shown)


pydanny:
Sorry @javipalanca for the slow response. 

In any case, is there any reason to use this when there is docker? Note that I'm not opposed, just asking the question. :smile:

javipalanca:
Hi Danny!
My only reason is that I've not learned docker yet... :=)
btw, does docker run in osx?

pydanny:
Hi @javipalanca! Yes, docker runs very nicely on OSX. I'm using docker-machine and docker-compose and it's very elegant and quick. Our instructions for it in the README aren't quite there yet, something I plan to address in the next few days.

What's nice is besides one small flag I can literally use the same commands locally for provisioning new virtualboxes as I do on Digital Ocean. I don't ever remember having that superpower with Vagrant.

pydanny:
@javipalanca I'm going to go ahead and close this PR. If you think it should get in, let's talk. We always welcome contributions and I don't want to dissuade from making more in the future.

javipalanca:
Hi @pydanny,
Don't worry. I understand. I'm not dissuaded from making more contributions ;-)
In fact, I'm going to learn Docker!

Issue #194, opened 2015-02-26
Add optional sentry configuration


pydanny:
`cookiecutter.json` must have a prompt that says, "Sentry for monitoring? (y/n)". If the user clicks yes, then template adds in the Raven package and configuration.

theskumar:
Closed via #292

PR #292, opened 2015-07-28
Feature/sentry support

burhan:
Adding sentry support for #194 

Note that I have left the URL as an argument, in case the Sentry server is local and not the hosted version at getsentry.com

All other settings have also been integrated, including the middleware.

### DIFF
--- a/{{cookiecutter.repo_name}}/config/settings/production.py
+++ b/{{cookiecutter.repo_name}}/config/settings/production.py
@@ -6,14 +6,18 @@ Production Configurations
 - Use Amazon's S3 for storing static files and uploaded media
 - Use mailgun to send emails
 - Use Redis on Heroku
-'''
 {% if cookiecutter.use_sentry == "y" %}
-import raven
+- Use sentry for error logging
 {% endif %}
+'''
 from __future__ import absolute_import, unicode_literals
 
 from boto.s3.connection import OrdinaryCallingFormat
 from django.utils import six
+{% if cookiecutter.use_sentry == "y" %}
+import raven
+import logging
+{% endif %}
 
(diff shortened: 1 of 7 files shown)


{{cookiecutter.repo_name}}/config/settings/production.py
```diff
@@ -5,11 +5,13 @@
 - Use djangosecure
 - Use Amazon's S3 for storing static files and uploaded media
 - Use mailgun to send emails
-- Use MEMCACHIER on Heroku
+- Use Redis on Heroku
 '''
+{% if cookiecutter.use_sentry == "y" %}
+import raven
+{% endif %}
```
luzfcb: __future__ import must be the first import the module

luzfcb:
please, update the tests https://github.com/pydanny/cookiecutter-django/blob/master/tests/base.py#L18-L28 and README.rst usage section

burhan:
Hi @luzfcb - README.rst was updated and tests also pass; I am working on a different PR to update the tests for the remaining features. There is some refactoring that can be done.

{{cookiecutter.repo_name}}/config/wsgi.py
```diff
@@ -36,6 +39,9 @@
 # See: https://whitenoise.readthedocs.org/
 application = DjangoWhiteNoise(application)
 {%- endif %}
+{% if cookiecutter.use_sentry == "y" -%}
+application = Sentry(application)
```
luzfcb: is this really necessary? if use_whitenoise is enable, this overwrites the previous `application`
some documentation about it?

luzfcb:
@pydanny @theskumar ok for me.

PR #295, opened 2015-08-01
Added support for the Hitch integration testing framework.

crdoconnor:
See http://hitchtest.com for details and videos.


pydanny:
@crdoconnor rdoconnor, thanks for this PR! Hitch appeared on my radar recently, and I've been looking for a chance to check it out. I do have some questions:

1. Does it work on OSX? I see a lot of libraries in `requirements.apt` that I don't think are in most OSX installations.
2. Does it work on Windows? `cookiecutter-django` provides minimal support for Windows and I doubt all the libs work there either.
3. What about Travis, Jenkins, AppVeyor (Windows), and other CI tools? How does Hitch interact with them?

If the answer is 'no' to any of those questions, that doesn't mean Hitchtest can't be included. Instead, what it means:

1. The dependency on Linux will need to be specified in the documentation. 
2. Instructions for running on CI tools should also be specified. I've debated adding `.travis` and `.appveyor` settings files, and this seems like the perfect time to focus on it.

crdoconnor:
1. Yes, it's regularly tested and it runs fine on OS X (as well as Arch, Debian, CentOS and Fedora). Cookie cutter did not seem to have an equivalent of requirements.apt for Mac OS X, though (or for various other linux distributions). What should I do about this?

2. No, unfortunately it will not work on Windows and it would be very difficult to make it do so :( I will add a note to the documentation if you like.

3. I haven't done much work on testing integrations with CI tools yet, but that's my next top priority. I'll try make sure I have instructions for jenkins and travis in the next few days.

pydanny:
Any thoughts @luzfcb or @theskumar? We're supposed to dally with the bleeding edge, and Hitch looks promising.

pydanny:
How well does it work in virtual environments a'la Docker/Vagrant?

crdoconnor:
The tests I did on other linux distributions apart from my own (Debian, Arch, CentOS, Fedora) were done in Vagrant. Works just fine.

I have not tried it in Docker yet. I'll add that to my to do list.

crdoconnor:
I have also made it work in Travis CI, incidentally. An example of that is here: https://travis-ci.org/hitchtest/django-remindme

I'll update the pull request to include an appropriate config for travis tonight.

pydanny:
@luzfcb, @theskumar Any comments? I like what I see, since end-to-end is a really nice feature. Also, it's totally opt-in: you can use it any time you want. Barring any further comments or objections I want to merge this PR into cookiecutter-django.

There is the Windows issue (it won't work there). But once we get this merged in and I get my Windows laptop running again (used solely for testing), I'll add in logic for windows users to not get quite so many hitch-focused files.

Again, let me know your thoughts.

{{cookiecutter.repo_name}}/README.rst
```diff
@@ -152,8 +152,32 @@ The email server listens on 127.0.0.1:1025
...
 
 
+Running end to end integration tests
+------------------------------------
+
+N.B. The integration tests will not run on Windows.
+
+To install the test runner::
+
+  $ sudo pip install hitch
```
luzfcb: I suppose that sudo is not necessary, unless you really want to install hitch globally

crdoconnor:
No, not strictly speaking necessary. Actually I use pipsi to install it, much like I do with cookiecutter.

crdoconnor:
@pydanny I have tested it in docker too now. It seems to run fine.

pydanny:
Grats and thanks @crdoconnor for giving cookiecutter Django end-to-end testing functionality. :ship:

crdoconnor:
Thanks for accepting it!

luzfcb:
@crdoconnor a good candidate for plugin is Splinter [1][2]. 

[1] http://splinter.readthedocs.org/en/latest/
[2] https://github.com/cobrateam/splinter

crdoconnor:
@luzfcb Actually you wouldn't even need a plugin to use splinter, since it can use a selenium remote webdriver (see here https://splinter.readthedocs.org/en/latest/drivers/remote.html). Just do "hitch install splinter", and in a step definition you can do the following (taken from the docs above):

    from splinter import Browser

    def step_definition(self):
        with Browser(driver_name="remote",
                     url=self.services['Firefox'].logs.json()[0]['uri']) as browser:
            print("Link to job: https://saucelabs.com/jobs/{}".format(
                  browser.driver.session_id))
            browser.visit("http://splinter.readthedocs.org")
            browser.find_link_by_text('documentation').first.click()

luzfcb:
@crdoconnor a question:
You have any plans to include cache of non-python packages, something like `pip` currently does.

For example, I already have a project with Hitch. after execute `hitch test stub.test`, Hitch download the Python, Postgresql and Redis packages.

I start a new project. When running Hitch in the new project, it will download again all non-python packages. This takes a lot of time on slow connections.

I sometimes suffer with bad internet connection. 

I suppose several other people may have the same problem

crdoconnor:
@luzfcb It already does this. The second time you run stub.test and python, postgres & redis are already built, it won't try to re-download or recompile them.

luzfcb:
@crdoconnor yes, he will not re-download if I'm in the same project, but if I create a new project in a totally different directory, add Hitch with exactly the same dependencies.
Hitch will download the non-python dependencies again, ( the python dependencies will not be downloaded again because pip already caches automatically)

crdoconnor:
Got it. I've created an issue: https://github.com/hitchtest/hitchtest/issues/1

Will fix it over the next few days.

PR #306, opened 2015-08-20
WIP: Pycharm Support (including debugging in Docker!)

noisy:
The goal of this PR is to introduce a "ready to go" configuration for PyCharm.

Moreover, this PR introduce debuggable dockercontainer, which can be build by debug.yml using docker-compass. Take a look on description in file:`{{cookiecutter.repo_name}}/docs/docker_remote_debugging.rst`

**Please Read**: [docker_remote_debugging.rst](https://github.com/SpisTresci/cookiecutter-django/blob/pycharm_support/%7B%7Bcookiecutter.repo_name%7D%7D/docs/docker_remote_debugging.rst)

For PyCharm users, this PR provides more than 20 *Run/Debug Configurations*, which can be very easily activated in two easy steps.

Instructions for other IDEs are more than welcome!

~~This PR has still status "Work in progress", because few other things has to be done:~~

## ToDo

* [x] - pycharm and docker are now optional
  * [x] - if user decide not to use pycharm or docker, proper files are removed in `post_gen_project.py` hook
* [x] - posibility of login into docker container with `compose/pycharm/.ssh_keys_to_docker/id_rsa`
* [x] - user can compile scss file inside a docker container. Compass doesn't have to be installed on host machine
* [x] - cleaning in `/compose/pycharm/` directory. Directory should be renamed to `/compose/debug/`, pycharm related stuff should be moved somewhere else.
* [x] -  ~~move node to separate container, and building assets process to entrypoint script~~
* [x] - fix problem with `Fatal error: Unable to find local grunt.` inside docker
* [x] - dev.yml and debug.yml should use this same user. **Note**: django user is renamed (only in debug container!) to docker_{{ cookiecutter.repo_name }}, to improve visibility during debugging multiple containers.
* [x] - improve .gitignore
* [x] - rebase
* [x] - write documentation in section *Configure Remote Python Interpreter based on deployment settings* in `docs/docker_remote_debugging.rst`
  * [x] - Add screenshots
* [x] - ~~some tests~~, PoC prepared: https://github.com/pydanny/cookiecutter-django/issues/309


{{cookiecutter.repo_name}}/compose/pycharm/.ssh_keys_to_docker/id_rsa
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
pydanny: Why provide an `ssh` key?

{{cookiecutter.repo_name}}/compose/pycharm/sshd_config
```diff
@@ -0,0 +1,82 @@
+# TODO: this could use some cleanup...
```
pydanny: English would be nice. :wink

pydanny:
The docker support in this PR is building off the current docker support in cookiecutter-django, right?

{{cookiecutter.repo_name}}/compose/pycharm/.ssh_keys_to_docker/id_rsa
noisy: this is a feature. Because this debug container with sshd is used only for debbuging, we do not care, that everyone have access to it. Typing password all the time is just annoying.

noisy:
> The docker support in this PR is building off the current docker support in cookiecutter-django, right?
@pydanny exactly. With very strong cooperation with @jayfk :)

{{cookiecutter.repo_name}}/config/settings/common.py
```diff
@@ -99,7 +99,11 @@
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
 DATABASES = {
     # Raises ImproperlyConfigured exception if DATABASE_URL not in os.environ
-    'default': env.db("DATABASE_URL", default="postgres://{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.repo_name}}"),
+    {% if cookiecutter.use_docker == 'y' %}
```
jayfk: This won't work if the user has changed username/password with env variables. `entrypoint.sh` already populates all needed env variables, so I see no use case to add this explicitly.

cookiecutter.json
```diff
@@ -13,5 +13,7 @@
     "use_celery": "n",
     "use_maildump": "n",
     "use_sentry": "n",
+    "use_docker": "n",
```
jayfk: I'm -1 on introducing on introducing the `use_docker` question here. People will end up with all the docker specific files anyway because there is no way to exclude files.

They will have the files but no further documentation for it.

The current implementation integrates nicely in the eco system we already have. A user can switch between all available options (virtualenv or docker) without the need to change anything. For the time being, we should make sure that we keep these side by side without introducing too much switches.

noisy: > People will end up with all the docker specific files anyway because there is no way to exclude files.

but I exclude them :wink: 

Please read description of PR carefully:

> if user decide not to use pycharm or docker, proper files are removed in post_gen_project.py hook

----

> For the time being, we should make sure that we keep these side by side without introducing too much switches.

This is a though topic. I believe, that user should use defaults answers (for most feautes right now, default is `n`), and only type `y`, when knows what exactly this mean. Moreover I want to keep final product of generation as clean as possible, and those docker-related files, IMO actually makes a difference:

with and without docker

@pydanny what do you think?

pydanny:
I much prefer having Docker in by default. It should be a viable path for development and eventually the defacto deployment target of cookiecutter-django created projects. Let's face it, while the tool chain for deployment isn't 100% there, Docker itself is arguably ready for production. 

Furthermore, by going with Docker as a default eventually for production, that makes the defacto endorsement of one PAAS over others (Heroku) go away. Considering Heroku doesn't sponsor this project, (and no longer sponsors djangopackages.com) ending their defacto endorsement as the primary deployment target for cookiecutter-django is a good thing.

Note: Making this project ready for Docker deployment will require a completely separate pull request and shouldn't happen for a while.

pydanny:
LGTM. @jayfk, what are your thoughts?

jayfk:
I hate to say this because I know how much effort is behind this work, but I think it adds too much complexity.

I'm always trying to follow the KISS principle and - when it comes to docker - to tailor an image that is perfect to run one task and nothing more. This is not the case here. The base image in this PR contains a full `node`, `python` and `ruby` stack. Furthermore, `entrypoint.sh` runs `npm install` the `grunt` task and adds a user on every container startup (TBH: I don't think this even works since the bash script uses `set -e` and afaik `adduser` returns a non zero exit code if the user already exists).

Don't get me wrong @noisy, I'm using hacks like this for my personal projects all the time to make it work, but I wouldn't want other users to fiddle around my hacks in a widely adapted open source project the people are referring to when it comes to best practices.

I'd love to see this PR tailored **just** for the task to add enhanced Pycharm support and nothing more. I have a issue for `grunt` already open [here](https://github.com/pydanny/cookiecutter-django/issues/310) so I think it would be better to separate these two things and to continue the discussion there.

Sidenote: I wasn't able to run the `dev` environment:

    export COMPOSE_FILE=dev.yml
    docker-compose up
    ...
    Step 7 : ENTRYPOINT /app/compose/django/entrypoint.sh
     ---> Running in f0b4ae65dc0d
     ---> cefd938a8bb6 
     Removing intermediate container f0b4ae65dc0d
     Successfully built cefd938a8bb6
     Creating dpy_django_1...
     Cannot start container 4bb14021e74d29ac892b13e291dd898837fb8b994fe978ef8b75afcac8cd5f38: [8] System error: exec: "/app/compose/django/entrypoint.sh": stat /app/compose/django/entrypoint.sh: no such file or directory

noisy:
@jayfk I checked one more time. And "on my machine" #306 works correctly. I know, that is is not excuse and we should be sure, that this works everywhere without problems, so please take a look on https://asciinema.org/a/eyjhuzhllmcwro8dem0qszz3q. Maybe we could figure out why this doesn't work on yours.

> Sidenote: I wasn't able to run the dev environment:

are you sure, that you rebuild image before calling `docker-compose up`?

> TBH: I don't think this even works since the bash script uses set -e and afaik adduser returns a non zero exit code if the user already exists

I use `useradd`, and in our case, this never return other value than zero, because Dockerfile is a guarantee, that any other user besides root does not exist. Remember, that `entrypoint.sh` is run only on fresh container.

Even if you will run `docker-compose up` in one shell, and `docker-compose run django python manage.py migrate` in another shell,  everything will be still working. That is because `docker-compose run` runs new container everytime. Changes in database are preserved, because of linked database container, which has volume set to `/data/`

Nevertheless, I can agree, that part of my fixes are beyond scope of this task. I my opinion, there is no much sense having docker, If I still need install node, grunt, compass on my local machine. Without that fixes, current `django` image is not ready for production at all.

> Don't get me wrong @noisy, I'm using hacks like this for my personal projects all the time to make it work, but I wouldn't want other users to fiddle around my hacks in a widely adapted open source project the people are referring to when it comes to best practices.

I am not sure, to which "hacks" are you referring to. Could you point me them inline in the code, and then we could think about alternatives.

jayfk:
> @jayfk I checked one more time. And "on my machine" PR works correctly. I know, that is is not excuse and we should be sure, that this works everywhere without problems, so please take a look on:

I'm relatively short on time right now so I can't give you a detailed reply, I just wanted to let you know that I've managed to get it working when rebuilding with `--no-cache`. 

I'll have more time over the weekend.

jayfk:
> I use useradd, and in our case, this never return other value than zero, because Dockerfile is a guarantee, that any other user besides root does not exist. Remember, that entrypoint.sh is run only on fresh container.

No, `entrypoint.sh` is run on every startup: https://docs.docker.com/articles/dockerfile_best-practices/#entrypoint

You can easily check this out for yourself. If you run:

    docker-compose up
    CTRL-C
    docker-compose up
    django_1   | useradd: user 'django' already exists

As of `1.4.0`, `docker-compose` re uses the old container and runs `adduser` again if there are no changes.

Other than that, there are a couple more problems: You can't run the project as root, or over `sshfs` from a mac. Haven't tried `vboxfs` on windows, but my guess is that we might run into problems there, too.


> Nevertheless, I can agree, that part of my fixes are beyond scope of this task. I my opinion, there is no much sense having docker, If I still need install node, grunt, compass on my local machine.

I can understand that, but I'd tackle this problem from a different angle. Instead of having one gigantic image with all batteries included, I'd create separate images for each task and link them together or run one off commands before deploying.

Grunt is extremely unfriendly here because it has to be installed in the current working directory as pointed out [here](https://github.com/pydanny/cookiecutter-django/issues/310). That's like shipping your whole virtual environment with your python code.

As I said before, I think the whole Pycharm support is fantastic and I'd merge this in an instant but without the changes regarding `node`, `compass` and `grunt`. We have to design this more carefully In my opinion.

pydanny:
Paging @pauleveritt. Could you take a look at this pull request and weigh in? We really want cookiecutter-django to have PyCharm support!

For reference, @pauleveritt is an old mentor and friend of mine. He was also just hired as developer evangelist for PyCharm.

noisy:
@pydanny Sorry for lack of update. I will be working on that PR on Saturday & Sunday

pydanny:
@noisy Hey! No worries. You have an excellent (and awesome) reason why you stepped back from this pull request. When you are comfortably able to work on this effort, then step back in.

Just remember that this is open source work. Most of us don't get paid for it, so other priorities come first. :)

jayfk:
I've done a lot of groundwork in this branch here: https://github.com/jayfk/cookiecutter-django/tree/SpisTresci-pycharm_support

Your debug ssh container is now basically a first class citizen in the `dev.yml` configuration :)

Could you pull from my branch and rebase on top of the latest commit? I could open a PR against your fork, but you'd have to solve all merge conflicts since the two branches have diverged significantly. This might be the easiest thing to to.

I've done some minor tweaks to the `ssh` configuration (users & host_key). This is still undocumented, but I'm sure you'll notice the difference by going over the diff.

pauleveritt:
HI @pydanny this ticket is some seriously impressive stuff. I haven't done Docker for more than trivial stuff, so I'll need some time to get my brain wrapped around the various moving parts of this ticket.

burhan:
Amazing work @noisy 

I'm with @jayfk - I think this should be split into two in order to ensure that both ends (Pycharm and Docker) are properly tested and vetted and do not dependent. For example, in the Pycharm configuration, it is assumed that the project will be using docker.

There are also some other issues (for example, the uuid should be random generated in the `webServers.xml`) etc.

pydanny:
@jayfk Go ahead and work off your branch. Once we merge that in we'll make triply sure @noisy gets tons of credit for his incredible work. :smile: 

@noisy I can't say enough how good your work has been. I'm glad that you have your priorities straight, family always comes first!

burhan:
Quick question - what software was used for the screen capture in @noisy's comment?

luzfcb:
@burhan https://asciinema.org/

```bash
# login first
asciinema auth
# after
asciinema rec
# or
asciinema rec file_to_record_session.json
# or
asciinema rec -t "Title to this recorded session" file_to_record_session.json
# or
# --max-wait 2 starts recording session in which all pauses longer than 2 seconds are replaced with exactly 2 second ones.
asciinema rec --max-wait 2 -t "Title to this recorded session" file_to_record_session.json
# you can edit the file json and after upload
asciinema upload file_to_record_session.json
```

luzfcb:
PyCharm 5 may be shipped with integrated support for docker: http://blog.jetbrains.com/pycharm/2015/10/announcing-pycharm-5-eap-143-165-docker-integration/

jayfk:
Yeah,they are talking directly to the docker daemon. When looking at the screenshots it seems this is a half-assed solution with docker-machine only that runs an exclusive python container. I can be completely wrong here, so I'll have to test it first ;)

Not exactly what we looked for, but it's a start.

noisy:
@luzfcb, @jayfk: PyCharm 5 EAP 143.165 looks very promising! It would simplify this PR a lot.

jayfk:
Works pretty good so far. 

All you have to do is fill in `{{cookiecutter.repo_name}}_django` as image name (has to be build with docker-compose before) and your default machine. Also, make sure to put in `/usr/local/bin/python` in python interpreter path.

@noisy You are linux, right? Do they offer support for native docker, or do you have to use `docker-machine`?

yunti:
Note pycharm has added further docker support to their latest version which may affect the implementation here.  http://blog.jetbrains.com/pycharm/2015/10/announcing-pycharm-5-eap-143-165-docker-integration/

noisy:
FYI:

I have checked the possibilities of [latest pycharm EAP version ](http://blog.jetbrains.com/pycharm/2015/10/announcing-pycharm-5-eap-143-165-docker-integration/#comment-252432)  with docker support and for now it doesn't allow to use docker-compose. Right now it supports only "single image mode"

 So, there are three choices:
1) correct and adjust current implementation
2) wait for docker-compose support in PyCharm (there is no ETA)
3) prepare single debug image with all stuff in it (all images combined into one)

I would prefer to wait for docker-compose support, mostly because other options have too many disadvantages.

**[EDIT]**: PyCharm Product Manager says that docker-compose support will not be added in current release: 
https://twitter.com/filippovdmitry/status/659171042083127296

jayfk:
Yup, let's go with 2. Even if it takes two years for them to support it. It was a cool thing to work on nevertheless.

noisy:
> Even if it takes two years for them to support it.

I doubt that this will take so long ;) Dmitry wrote:

> We’ll consider support for docker-compose for the next release.

I have also designed suggested solution: http://i.imgur.com/ZJpfnJU.png

pauleveritt:
As an FYI, I’ve started talking to Dmitry about CookieCutter in general. (As FYI, I’m new as the PyCharm and WebStorm Developer Advocate.) I haven’t pushed too much, as the PyCharm 5 release cycle is…consuming. :)

I’d like though to step back, both in PyCharm and Python in general, and see how we can improve the getting started experience. Obviously CookieCutter has the right idea (move scaffolds out of each web etc. project.) I still lament the virtualenv part, but perhaps in something like PyCharm, that part can be handled out-of-band.

None of this has anything to do with Docker, though, so I’ll shut up.

—Paul

burhan:
@pauleveritt I would really appreciate if you could have a look at https://youtrack.jetbrains.com/issue/PY-14237 as a possible feature addition for sometime soon in PyCharm.

pauleveritt:
Hi Burhan. Yes indeed, I have been talking to Dmitry about cookiecutter support directly in the Create Project wizard. WebStorm now support Yeoman templates, so it is a little bit like that.

—Paul

noisy:
FYI: https://youtrack.jetbrains.com/issue/IDEA-137765 just changed status to "in progress". I will adapt this PR, whehn this feature be released in EAP version.

mimischi:
The newest PyCharm EAP 2016.1 RC2 (145.259) finally works for me with the current cookiecutter-django configuration using docker-compose v2 on OS X.

Here are two screenshots on how I "configured" it (didn't really change much..):

Preferences -> Project -> Project Interpreter -> Add Remote..
Remote configuration

Run -> Edit Configurations -> Defaults -> Django server 
Run configuration

**Note: ** the host needs to be set to 0.0.0.0, the "Run browser" IP is my docker-machine env IP, as well as the "Custom run command" is set to "runserver_plus"

Attaching the debugger also works without any changes to the current cookiecutter code base.

Only thing that does not work is running manage.py commands from inside PyCharm. This seems to restart the django service again. I just ran "docker-compose -f dev.yml run django python manage.py migrate" via Terminal on the first start up.

noisy:
@mimischi thanks for information. I'm on it. I have t sync with @jayfk and I will update this PR.

jayfk:
@noisy I'd strongly suggest to start with a fresh branch from master. There have been a lot of changes in the last 10 months that make it hard to follow the changes this way.

pydanny:
I think I agree with @jayfk. This does represent a huge amount of work, and I'm willing to help @noisy  with creating a new branch from master.

noisy:
> I'd strongly suggest to start with a fresh branch from master. There have been a lot of changes in the last 10 months that make it hard to follow the changes this way.

I am closing this ticket. New Pull Request https://github.com/pydanny/cookiecutter-django/pull/507 was created.

noisy:
@mimischi so far almost everything is working as expected. Right now I have only an issue with degguging. 

selection_204

Is it working for you?

mimischi:
It does work for me and I did not change anything to get the debugger working (besides the settings mentioned above). I noticed you are running "runserver" instead of "runserver_plus" as specified normally by cookiecutter-django. Is this maybe the culprit? I can't get it to work without runserver_plus on my machine.

bildschirmfoto 2016-03-23 um 22 08 01

noisy:
@mimischi I think this is something else. Moreover I would like to have a possiblility not only debugging main process run by runserver, but also tests or django managment tasks.

Could you join our chat on https://gitter.im/pydanny/cookiecutter-django - it will be easier to figure out toghether on chat where the problem is.

PR #315, opened 2015-08-30
Upgrade to Bootstrap 4 alpha

audreyr:
People use cookiecutter-django to create new Django projects, and I think there's no point starting a project with Bootstrap 3 when Bootstrap 4 will be ready soon. This makes it easier for them to upgrade to Bootstrap 4 final when it comes out.

The main changes here are the updated stylesheet and navbar HTML.


luzfcb:
+1
Cool, Bootstrap 4 seems to be coming with a lot of good improvements ... 
Another css framework interesting that i discovered recently was the Semantic-ui  http://semantic-ui.com/ 
https://github.com/semantic-org/semantic-ui/

luzfcb:
@audreyr I was updating dependencies, and I was in doubt if the current bootstrap3 template of django-crispy-forms will work without changes with bootstrap 4

PR #319, opened 2015-09-01
Allowed hosts

bogdal:
For security reasons, we should set explicitly list of allowed hosts.


pydanny:
I have a beef with `ALLOWED_HOSTS`, and that's the terrible error messages it provides when it isn't set correctly. What it does is:

1. Throws 500 errors on page loads.
2. Reports a nigh useless `SuspiciousOperation` in the logs. Doesn't explain what is causing the problem. Unless you are experienced at Django, this is tremendously annoying to debug. Even if you are experienced at Django, sometimes you stumble into it anyway.

Therefore, if we're going to change the default from `["*"]` to something else, we should do at least one of the following:

A. Document that this has weak messages on error. Maybe in a deployment troubleshooting section?
B. Figure out a way to improve the logging of the exception.

bogdal:
@pydanny I have just added `django.security.DisallowedHost` to the loggers list.
An example from my console:
```
ERROR 2015-09-02 20:57:51,235 base 18882 4419268608 Invalid HTTP_HOST header: 'localhost:8000'. You may need to add u'localhost' to ALLOWED_HOSTS.
```

jayfk:
Could you add the environment variable to the docker specific `env.example` file in the root directory?

PR #332, opened 2015-09-17
Fix for wsgi.py for Raven in dev

yunti:


pydanny:
Grats on your first pull request @yunti! Welcome to the club! :smiley:

yunti:
Thanks - I really appreciate that! Silly question - this repo is a bit different to others as cookie cutter needs to be run to then spit out the project files - but then the resultant project is quite different to the original repo. What's the best way to deal with that? Clone django-cookiecutter -> make changes-> do pull request -> and then really on travis to verify the changes are ok.  (however in that workflow I can't do normal run server as I would still have all the cookie cutter switches in place - as. I haven't run cookie cutter )

Issue #322, opened 2015-09-11
Option for python version on deployment (and docker dev)


yunti:
It would be useful to have an option for which version of python is used for the current project.  Although  all used dependencies are targeted to be compatible with both python 2 and python3. The default deployment option (or current docker dev setup) is for python 2. 
  
Eg for heroku deployment a runtime file needs to be added or the heroku build pack will default to python 2.7.  
Additionally the docker file default is python 2.7.

Implementation:
This would probably be best done as a cookie cutter option (generally python version for a project is decided at the outset).  
For heroku we could include a runtime.txt file with something like:

{% if cookiecutter.use_python3= 'y' -%}
python-3.4.3
{%- endif %}

Assuming that a blank runtime file for Heroku would default back to python 2? (or include an else if not)

The docker file could be altered similarly:
{% if cookiecutter.use_python3= 'y' -%}
FROM python:3.4
{% else %}
FROM python:2.7
{%- endif %}

pydanny:
Would you like to submit a pull request?

jayfk:
I was thinking of something like this, too. What about flipping the question so that it asks for `python2` and make `python3` the default?

yunti:
That would be the preference for me too making python 3 the default - but it still feels like I'm in the minority there, but as this is more a bleeding edge template - perhaps that's not a bad idea here.

jayfk:
I'm closing this because the changes have been merged. Thanks @yunti

PR #336, opened 2015-09-18
Add python version option for deployment

yunti:
Fix for #322

### DIFF
--- a/{{cookiecutter.repo_name}}/Dockerfile
+++ b/{{cookiecutter.repo_name}}/Dockerfile
@@ -1,4 +1,8 @@
+{% if cookiecutter.use_python2= 'n' -%}
+FROM python:3.4
+{% else %}
 FROM python:2.7
+{%- endif %}
 ENV PYTHONUNBUFFERED 1
 
 # Requirements have to be pulled and installed here, otherwise caching won't work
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/Dockerfile
```diff
@@ -1,4 +1,8 @@
+{% if cookiecutter.use_python2= 'n' -%}
```
luzfcb:
```python
{% if cookiecutter.use_python2 == 'n' -%}
```

{{cookiecutter.repo_name}}/runtime.txt
```diff
@@ -0,0 +1,3 @@
+{% if cookiecutter.use_python2= 'n' -%}
```
luzfcb:
```python
{% if cookiecutter.use_python2 == 'n' -%}
```

yunti:
Hmm bit of a mess in terms of commits. Just let me know if you would rather I delete and resubmit.

crdoconnor:
Should I amend this too https://github.com/pydanny/cookiecutter-django/pull/340/files#diff-5f8fc5f3de1f2873b3cc21cf3f6df870R21 ?

audreyr:
Awesome. I went over this with @pydanny and am going to merge it in. Thanks @yunti and don't worry about the commits.

We should see about if 3.5 will work in Dockerfile and Heroku runtime rather than 3.4, but that can be a later pull request.

yunti:
Thanks , saw pydanny added 3.5 to docker file.  3.5 build pack isn't in Heroku yet but from this pull request should hopefully be soon: https://github.com/heroku/heroku-buildpack-python/pull/250

PR #340, opened 2015-09-22
Updates to the hitch tests

crdoconnor:
* Added the .hitch directory to gitignore.
* Added README specifically for running the tests.
* Removed the jinja2 parameterization in the stub test (it's an advanced feature; probably not really suitable for a quickstart project).
* Directly specifies the version of python to download, install and build the virtualenv with in engine.py - 2.7.10 if python 2 is specified by cookiecutter, 3.5.0 otherwise.
* Upgraded hitch requirements (includes a lot of bug fixes).
* Override the default settings with an environment variable so that it sends an actual email to HitchSMTP on port 1025 during tests rather than using the console back end.
* Now uses unixpackage to install prerequisite packages. Despite ubuntu packages being specified in settings.yml, the system will automatically try and install the correct equivalent for the platform it is on when running the test for the first time. On Arch it uses pacman, Fedora it uses yum, Mac OS X it uses brew, etc.
* There's finally a one step set up for tests.

It's currently been tested on Debian, Ubuntu, Arch and Mac OS X Mavericks.


luzfcb:
@crdoconnor one question about `unixpackage`:
when I added [`install_os_dependencies.sh`](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/install_os_dependencies.sh) script in https://github.com/pydanny/cookiecutter-django/pull/200, was commented the question that it is not prudent run, by default,  `apt-get install` without `--no-upgrade` option, because, by default, if a package is already installed and there is an update for it in the repository, `apt-get install` also the updates it.
This can result in breakage of existing legacy applications

https://github.com/pydanny/cookiecutter-django/pull/200#discussion_r26436046

crdoconnor:
Yep, that's a bug I'd say.

It checks in groups so if everything in the group is installed it won't run apt-get, but it will if just one of the packages is missing. Which is bad. It should only run apt-get for the missing package.

PR #346, opened 2015-09-24
Remove commented-out flake8 ignore rule.

audreyr:
Would it be okay to remove this commented-out line from our flake8 config?
Or is it there for a reason?


luzfcb:
to me is ok to remove. this comment was included in https://github.com/pydanny/cookiecutter-django/commit/10e3e8f6ab29b704ed2903fe53e04a18555dfab3. I could not test to verify if it's still valid.

PR #350, opened 2015-09-27
Fixed reStructuredText (RST) format

andor-pierdelacabeza:
The lack of a space between the '::' tag and the text did that the block around the code didn't appear.


audreyr:
Thanks for catching this, @andor-pierdelacabeza.

PR #351, opened 2015-09-28
Enviromental variables not being read.

sebastian-code:
I created a project recently, but the DB connection wasn't working, reviewing I found on django-environ this piece of code. I'm not sure if this is intended.


pydanny:
Hi @sebastian-code, according to django-environ, the way we currently are using the tool is correct. Reference:

* https://github.com/joke2k/django-environ/blob/master/environ/environ.py#L48-L49

The `Env.read_env` static method you suggest is used for reading in `.env` files. While this is useful to those who use `.env` files, for those of us who don't use `.env` files outside of the odd docker deployment, it will break the settings configuration. Reference:

* https://github.com/joke2k/django-environ/blob/master/environ/environ.py#L557-L558

Now that said, if using `Env.read_env` works for you, then by all means do so. :smile:

sebastian-code:
Thanks @pydanny for the answer, I didn't notice that part. That explain some things. Thanks again

PR #354, opened 2015-09-30
String view arguments to url() is deprecated in django 1.9

siauPatrick:
Hello. I tried django 1.9 and found
```bash
config/urls.py:29: 
RemovedInDjango110Warning: Support for string view arguments to 
url() is deprecated and will be removed in Django 1.10 
(got django.views.defaults.bad_request). Pass the callable instead.
  url(r'^400/$', 'django.views.defaults.bad_request'),
```


luzfcb:
@siauPatrick thank's :+1:

PR #358, opened 2015-10-03
escape admin url usage example in comment

luzfcb:


luzfcb:
oh, very fast merge @pydanny ... but, there's more places with that comment ...

personal note: Avoid creating pull-request via github web interface :see_no_evil:

Issue #356, opened 2015-10-01
Docker dev enviroment not working: python: can't open file '/app/manage.py': [Errno 2] No such file or directory


cepinos:
I am trying to run the docker dev environment and this throw me this error: python: can't open file '/app/manage.py': [Errno 2] No such file or directory

I am using ubuntu 14.04 LTS as host machine.

I ran this command $docker-compose -f dev.yml up

jayfk:
This is probably a permission problem. Do you run ubuntu on bare metal with a native docker installation, or is this on a VM with a docker remote daemon or something similar?

Can you run 
    
    pwd
    docker run -v $(pwd):/test -it ubuntu ls -la test

in you project folder and post the output here?

cepinos:
Hi @jayfk Thanks for respond! 

I am not sure what bare metal means, sorry :$. but my machine have Ubuntu installed native. 

I am following this instructions http://cookiecutter-django.readthedocs.org/en/latest/developing-locally-docker.html, so I guess docker will run in a VM, but also, I have docker running in the host machine. 

I ran the commands you asked me to run,

$pwd                                                                                                                                                                                                       /home/expedienteweb/inqbation/django-508q

docker run -v $(pwd):/test -it ubuntu ls -la test                                                                                                                                                         
WARNING: Error loading config file:open /home/expedienteweb/.dockercfg: permission denied
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
d3a1f33e8a5a: Pull complete 
c22013c84729: Pull complete 
d74508fb6632: Pull complete 
91e54dfb1179: Pull complete 
library/ubuntu:latest: The image you are pulling has been verified. Important: image verification is a tech preview feature and should not be relied on to provide security.
Digest: sha256:73fbe2308f5f5cb6e343425831b8ab44f10bbd77070ecdfbe4081daa4dbe3ed1
Status: Downloaded newer image for ubuntu:latest
total 4
drwxr-xr-x  2 root root   40 Oct  1 15:17 .
drwxr-xr-x 33 root root 4096 Oct  1 16:21 ..

Thanks a lot for you support!

cepinos:
Hi @jayfk, I was able to run it without the VM, I think maybe it could be a permissions problem, as you said. My partners are experimenting the same problem, they have mac as a host machine.

jayfk:
[...]

Issue #335, opened 2015-09-18
Reorganize documentation


pydanny:
cookiecutter-django is sophisticated enough to warrant its own top-level documentation served by Read the Docs, as well as generated project-level documentation. Unfortunately right now we have a mish-mash of organically grown documentation in the following areas:

* README.rst
* {{ cookiecutter.repo_name }}/README.rst (mix of features, config, and instructions)
* {{ cookiecutter.repo_name }}/docs/  (mix of features, config, and instructions)

What we are implementing now is the following:

* /docs/ (All configuration, setup, and testing documentation lives here)
* /README.rst (feature list, basic usage, links to other documentation levels)
* {{ cookiecutter.repo_name }}/README.rst (basic instructions, links to repo-level docs/)
* {{ cookiecutter.repo_name }}/docs/  (Content moved to this location from /docs/ in post_gen_hook)

This makes it possible for us to have one documentation location for all the technical options and use readthedocs.org trivially.

TASK: Modify the post_gen_hook to copy files from /docs/ to {{ cookiecutter.repo_name }}/docs/. This will require using the `.cookiecutters` directory to find the files in /docs/.  This may require using a `pre_gen_hook`.

luzfcb:
@pydanny this fits somewhere: https://gist.github.com/luzfcb/59cf1d3cf24bf801dd70

pydanny:
Yes! Absolutely! I think it should go into a Mac OS X specific section in https://github.com/pydanny/cookiecutter-django/blob/master/docs/developing-locally.rst :smile:

jayfk:
You moved the part about the persistent directory on boot2docker into the deprecated section.

> If you are using boot2docker to develop on OS X or Windows, you need to create a /data partition inside your boot2docker vm to make all changes persistent. If you don't do that your /data directory will get wiped out on every reboot.

> To create a persistent folder, log into the boot2docker vm by running:

> `$ bootdocker ssh`
> And then:

> `$ sudo su`
> `$ echo 'ln -sfn /mnt/sda1/data /data' >> /var/lib/boot2docker/bootlocal.sh`

This might lead to the conclusion that this is no longer the case on recent docker releases. Unless I have missed something, this is still a problem using docker-toolbox (because it uses boot2docker under the hood).

> If you are using OS X or Windows, you need to create a /data partition inside the virtual machine that runs the docker deamon in order make all changes persistent. If you don't do that your /data directory will get wiped out on every reboot.

[...]

PR #359, opened 2015-10-03
Docker docs cleanup

audreyr:
My work-in-progress attempt to tidy up the Docker docs for #335 and #356.

Still working on it, but I wanted to start a PR to give everyone a heads up and make sure I don't accidentally remove something important.

Suggestions welcome here or via Gitter.

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -16,7 +16,6 @@ Prerequisites
 If you don't already have these installed, you can get them at:
 
 * https://github.com/docker/toolbox/releases
-* https://www.virtualbox.org/wiki/Downloads
 
 Go to the Root of your Project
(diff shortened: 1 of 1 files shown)


audreyr:
Alright. I unified the parts that were redundant and improved the instructions as best I could. 

The diff is a bit complicated to follow, but if you go through the commits one by one, you can see my process. It may be best to just accept this as an incremental improvement and then iterate on it further in separate PRs. Thoughts?

jayfk:
Other than that, LGTM.

audreyr:
OK. I undid that last commit which was incorrectly renaming `django` to `dev1`: https://github.com/pydanny/cookiecutter-django/commit/08aea25424e889b57cda7dc5eb1fb726bd2236a0

Issue #360, opened 2015-10-04
Multiple Dockerfiles


audreyr:
If I understand the current Docker setup, we install both the requirements from `requirements/local.txt` and the Heroku `requirements.txt` into the Docker container since we just have 1 Dockerfile.

Shouldn't we have separate Dockerfiles for local, test, prod? For example, I feel uncomfortable installing local debugging tool requirements into production.

jayfk:
That's correct, these parts should really be separated. I'll continue to work on this PR.

jayfk:
This should work as expected. There are two dockerfiles for dev and prod. On `dev`, `local.txt` is installed. On `prod`, `production.txt`.

audreyr:
Merged #361.

PR #361, opened 2015-10-04
Multiple Dockerfiles. Implementation of #360.

audreyr:
Here's how having multiple Dockerfiles would work. 

I still want to play around with this a bit -- just sending the PR as a proposal for discussion.


audreyr:
@jayfk Awesome! That worked for me.

Any thoughts on having `dev.env` / `prod.env` as well?

audreyr:
Merging this in because it seems to work fine, and because we may need it for #369.

Issue #353, opened 2015-09-30
Opinions on (optional) New Relic support?


jayfk:
I'm using New Relic in most of my projects. Even the free tier gives a lot of helpful metrics. Since the initial setup is a little bit annoying, it might be helpful to have it as an optional thing that does everything for us.

luzfcb:
:+1: to me

yunti:
Would love this too

jayfk:
Merged in #362

PR #362, opened 2015-10-04
Add newrelic for monitoring.

amjith:
Please use this PR as a starting point for addressing https://github.com/pydanny/cookiecutter-django/issues/353. 

I have not tested this since I don't have various credentials (AWS keys) required to setup a production instance of the generated app. 

This PR also doesn't add logging for the newrelic agent because the [logging setup for sentry](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/config/settings/production.py#L167) seems to override everything else, which will have to change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,7 @@ Features
 * Pre configured Celery_ (optional)
 * Integration with Maildump_ for local email testing (optional)
 * Integration with Sentry_ for error logging (optional)
+* Integration with NewRelic_ for performance monitoring (optional)
 * Docker support using docker-compose_ for dev and prod
 * Procfile_ for deploying to Heroku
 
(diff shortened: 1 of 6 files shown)


jayfk:
thanks!

amjith:
flake8 check is failing because of extra blank lines. But the extra blank lines are introduced by the template system when the `if` conditions are false. I'm not sure what's the workaround for it.

audreyr:
@amjith You can use Jinja2's whitespace control: http://jinja.pocoo.org/docs/dev/templates/#whitespace-control

burhan:
Why not use the [heroku instructions for newrelic](https://docs.newrelic.com/docs/agents/python-agent/getting-started/django-heroku-quick-start)?

jayfk:
> Why not use the heroku instructions for newrelic?

That's a heroku specific thing that only works on heroku.

amjith:
The PR is made flake8 compliant. Sorry about the delay, I'm in India visiting family so computer time is limited.

Tests are passing. But I'd prefer if someone tries deploying an app using this template with newrelic enabled. As I mentioned in the original comment, I don't have a production instance to test this with.

jayfk:
Tested, works! Thanks @amjith

PR #364, opened 2015-10-04
Fixes #363 - Remove stray closing tags, fix navbar margin

### LINKED ISSUE #363: Extra </li> tags in base.html without <li> or <ul> tags?
Not quite sure what's going on here?

For example in base.html you will see:

    <nav class="nav navbar-nav">
         <a class="nav-link nav-item" href="{% url 'home' %}">Home</a></li>
         <a class="nav-link nav-item" href="{% url 'about' %}">About</a></li>
    </nav>

Presumably bootstrap 4 doesn't need the < li > tags any more and someone just went through and removed all the < li > tags without removing < /li > or is there something more to it?

hairychris:
Removes stray </ul> tags that don't seem like they should be there. Reduces topbar bottom margin because it looked bad.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.html
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/templates/base.html
@@ -31,7 +31,7 @@
 
   <body>
     
-    <div class="m-b-lg">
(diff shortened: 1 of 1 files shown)


audreyr:
Good catch @hairychris. And thanks for the bottom margin tweak.

Issue #365, opened 2015-10-05
Crispy Forms horizontal layout broken with Bootstrap 4


pusateri:
Trying to use horizontal forms and Bootstrap 4 requires "row" class for horizontal forms in form-group div (see http://www.quackit.com/bootstrap/bootstrap_4/tutorial/bootstrap_forms.cfm) but crispy-forms doesn't know about Bootstrap 4 yet so the current combination breaks horizontal forms.

jayfk:
Related: https://github.com/maraujop/django-crispy-forms/issues/524

pydanny:
For the time being, we might want to do a template override to make Bootstrap 4 templates work. As I've got a project that just encountered this problem, I'll try and get this done today.

audreyr:
Looks like this was fixed in upstream crispy-forms, but there isn't a new crispy-forms release out yet.

jayfk:
This should be working now.

PR #366, opened 2015-10-05
Adds Bootstrap 4 classes as crispy-forms override

pydanny:
While we're waiting on a crispy-forms update I added a template override for that library's bootstrap3 implementation. What it does is apply the missing classes to the form. Addresses #365.

I've flagged the relevant crispy-forms issues so when (maraujop/django-crispy-forms#524), so when this is addressed in that package, we can remove this template.


areski:
It seems that this has been merged https://github.com/maraujop/django-crispy-forms/pull/528
:+1:

audreyr:
Yep, just waiting for the next crispy-forms release so we can update the requirements. Closing this in the meantime.

PR #369, opened 2015-10-06
WIP: docker-controls

jayfk:
Adds utility functions for docker. Would love to see some testing before giving it a go.

Please take a look at the code to see what it does. Most of it is pretty straightforward and documented.

To do:
- [ ] More documentation
- [ ] Implement all database controls
- [ ] Fix PEP issues
- [ ] Refactor behemoth sql function
- [ ] Add a few lines about fabrics code completion

### DIFF
--- a/{{cookiecutter.repo_name}}/fabfile.py
+++ b/{{cookiecutter.repo_name}}/fabfile.py
@@ -1,0 +1,249 @@
+# -*- coding: utf-8 -*-
+"""
+This is a collection of useful utility functions when working with docker on different environments.
+
+In order to use these functions, install fabric on your local machine with::
+
+    pip install fabric
+
+Please note: Fabric is a remote code execution tool, NOT a remote configuration tool. While you can copy files
+from here to there, it is not a good replacement for salt or ansible in this regard.
+
+There is a function called `production` where you need to fill in the details about your production machine(s).
+
+You can then run::
+
+    fab production status
+
+to get the status of your stack
+
+To list all available commands, run::
+
+    fab -l
+"""
+
+from __future__ import absolute_import, print_function, unicode_literals
+from fabric.operations import local as lrun, run, sudo
+from fabric.api import *
+from fabric.colors import green, red, yellow
+from fabric.contrib.console import confirm
+
+
+def local():
+    """
+    Work on the local environment
+    """
+    env.compose_file = "dev.yml"
+    env.project_dir = "."
+    env.run = lrun
+    env.cd = lcd
+
+
+def production():
+    """
+    Work on the production environment
+    """
+    env.hosts = [""]  # list the ip addresses or domain names of your production boxes here
+    env.port = 22  # ssh port
+    env.user = "root"  # remote user, see `env.run` if you don't log in as root
+
+    env.compose_file = "docker-compose.yml"
+    env.project_dir = ""  # this is the project dir where your code lives on this machine
+
+    # if you don't use key authentication, add your password here
+    # env.password = "foobar"
+    # if your machine has no bash installed, fall back to sh
+    # env.shell = "/bin/sh -c"
+
+    env.run = run  # if you don't log in as root, replace with 'env.run = sudo'
+    env.cd = cd
+
+
+def rollback(commit="HEAD~1"):
+    """
+    Rollback to a previous commit and build the stack
+    :param commit: Commit you want to roll back to. Default is the previous commit
+    """
+    with env.cd(env.project_dir):
+        env.run("git checkout {}".format(commit))
+
+    docker_compose("build")
+
+
+def deploy():
+    """
+    Pulls the latest changes from master, rebuilt and restarts the stack
+    """
+    with env.cd(env.project_dir):
+        env.run("git pull origin master")
+
+    build()
+    restart()
+
+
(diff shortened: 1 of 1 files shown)


audreyr:
:+1: Trying this out right now. Amazing work!

{{cookiecutter.repo_name}}/fabfile.py
```diff
@@ -0,0 +1,249 @@
...
+
+def build(cache=True):
+    """
+    Builds the the stack
+    :param cache: Use docker cache. Default is True
+    """
+    docker_compose("build" if cache else "build --no-cache")
+
+
+def status():
```
audreyr: I think `ps` would be easier to type/remember. But this is minor; I'm fine with leaving it as-is if you prefer.

{{cookiecutter.repo_name}}/fabfile.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    Work on the local environment
+    """
+    env.compose_file = "dev.yml"
+    env.project_dir = "."
+    env.run = lrun
+    env.cd = lcd
+
+
+def production():
```
audreyr: How about just `prod` for easier typing?

{{cookiecutter.repo_name}}/fabfile.py
```diff
@@ -0,0 +1,249 @@
...
+
+def deploy():
+    """
+    Pulls the latest changes from master, rebuilt and restarts the stack
+    """
+    with env.cd(env.project_dir):
+        env.run("git pull origin master")
+
+    build()
+    restart()
```
audreyr: I don't know if it's just a problem with my setup or understanding, but `fab local deploy` didn't work when I tried it. I had to run `fab local docker_compose:'up -d'` (which seems like it should do the same thing, so it's a bit confusing to me).

jayfk: You are right, `deploy` is a bit misleading here and makes little sense in a local environment. The function is intended to be used on remote servers only (comes in handy if you have multiple remote machines, e.g `staging`/`testing`).

What do you think about adding a switch that checks if the command is run on `local`? If so, we could tell the user to run a second function instead that is meant to be run on `local` only. Something like `up` or `start`.

This would make this inconsistent for different environments but it might be the most intuitive.

What do you think?

{{cookiecutter.repo_name}}/fabfile.py
jayfk: :+1:

{{cookiecutter.repo_name}}/fabfile.py
audreyr: I'm on the fence. Perhaps just a separate `up` command intended for `local` might be enough. I have no strong opinion here.

mjsisley:
I have been using .bash_profile/.bashrc for a simpler version of this, but want better controls. 

- Fabric is python2 only. 
- I'm not sure how close Fabric 2.x (with python3 support) is. 
- There is a Fabric3 fork, but tests with this script fail. I didn't look too hard, but problems with the import. Throwing Attribute Errors.

I'll try to finish these but it is not clear what the best option for a tool that works on python2 and python3. 

Any opinions on:

- Waiting until python3 support arrives for Fabric? 
- A replacement tool? 
- Use Make?

jayfk:
Thanks again for picking this up. As discussed in the chat we recently, using fabric is not a good idea at this point because of the missing python 3 support.
  
> A replacement tool?

Maybe invoke? Any other ideas?

> Use Make?

Or a shell file?

I'd say use whatever works for you and is readable :)

ssteinerx:
Where is "unknown repository" and what happened to this idea?

mjsisley:
A version of the controls that I have used, made using (Click)[https://github.com/pallets/click], are here https://github.com/mjsisley/cookiecutter-controls/tree/master/%7B%7Bcookiecutter.project_slug%7D%7D/controls/ccc and could be pulled out of the repo. You are going to need to look into installing a click package to use them currently though. These commands can only be run locally in their current state. I haven't tried to extend them to work over ssh. It may be better to reimplement them with Make or use Fabric if you are on python 2 (maybe it is available on 3 now, haven't checked).

@jayfk Let me know if you want me to try to pull the controls out and send a PR, not sure if it is worth it.

As a background on the cookiecutter-controls repo:

I went a little crazy here. I made it in such a way so you could have a namespaced CLI for each project... which is probably unnecessary. I was experimenting with composing multiple cookiecutter projects. I think something like audreyr/cookiecutter#770 would help with making it slightly easier to compose multiple cookiecutter projects.

The idea was that a standard cookiecutter-cli, namespaced to each project, could be incorporated into any other cookiecutter project by copy and pasting some code into the post-gen hooks. The vision was to try to work towards using cookiecutter in a more package-like way. I have realized that the reality of the maintenance of cookiecutter projects makes that probably more trouble than it is worth.

jayfk:
@mjsisley Thanks for sharing, looks interesting. Are you actively using that in a project right now?

As for this PR and related tools: I think at this point it's better to focus on the docker specific documentation than to introduce yet another tool for convenience.

Once we have a good documentation we can maybe wrap that up in a neat script to make things easier.

Issue #311, opened 2015-08-25
issue when repo_name does not follow naming convention


cdvv7788:
I guess this was added to avoid name clashes, but there can be issues with some imports:

```Python
from my-report-name-with-invalid-name.models import something
```
It could be useful to force the repo_name to follow python naming conventions or mention that in the installation steps.

pydanny:
Would you like to submit a pull request for a documentation change?

cdvv7788:
Yeah, i can submit a little PR mentioning this.

jayfk:
This was fixed with #375, thanks @cdvv7788!

PR #375, opened 2015-10-13
Readme update

cdvv7788:
Added a warning about repo_name, which needs to be a valid python module name. https://github.com/pydanny/cookiecutter-django/issues/311


luzfcb:
merged. Thanks @cdvv7788

PR #377, opened 2015-10-14
Added clarification on building for local or production

MathijsHoogland:


audreyr:
Thanks much @MathijsHoogland. We particularly appreciate documentation pull requests like this!

PR #378, opened 2015-10-14
Added sudo: true to the travis file

MathijsHoogland:
Sudo isn't true for default for some people in travis (people with a student license for example).


audreyr:
Thanks for catching this and submitting a fix @MathijsHoogland.

PR #381, opened 2015-10-18
Add factories for model users.User

ad-m:


pydanny:
This functionality is encapsulated by django-test-plus' `self.make_user()` function. Which means one of the following:

* We need to document why we don't include a `UserFactory`
* There might be a good reason to include this factory.

@frankwiles, @ad-m Do you have any comments or thoughts?

ad-m:
I agree that UserFactory alone can be replaced by ```self.make_user()```, but cookiecutter-django is starter for other projects, where ```UserFactory``` is powerfull to use as subfactory (see ```factory.SubFactory``` from factoryboy). Using ```self.make_user()``` to populate subfactories (or (sub)^x factories) sounds awful.

ad-m:
I added default password for ```UserFactory``` like ```self.make_user()```, so you can log-in freely by ```self.client.login(username=UserFactory().username, password='password')``` too.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/users/tests/factories.py
```diff
@@ -0,0 +1,12 @@
+from feder.users import models
```
Eraldo: Just out of curiosity..
Where does `feder` come from?

frankwiles:
I agree with @ad-m ```self.make_user()``` is fine for simple tests, but having a true User factory is more powerful, but often project specific.  In my experience, if you're making a custom user model it's very likely for things like removing username and using email login only or other major surgery that ```make_user()``` isn't ever going to be able to support.

{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/users/tests/factories.py
```diff
@@ -0,0 +1,12 @@
...
+import factory
+
+
+class UserFactory(factory.django.DjangoModelFactory):
+    username = factory.Sequence(lambda n: 'user-{0}'.format(n))
+    email = factory.Sequence(lambda n: 'user-{0}@example.com'.format(n))
+    password = factory.PosteGnerationMethodCall('set_password', 'password')
```
pydanny: `PosteGnerationMethodCall` appears to be a typo. :wink:

pydanny:
@ad-m Great explanation! @frankwiles, thank you for your commentary. :smile: 

@ad-m Can you fix the typos so we can accept this pull request? :sunglasses:

PR #382, opened 2015-10-20
Hitch : Updates to the default end to end test framework template

crdoconnor:
I've made some improvements.

The framework is now a lot less bleeding edge and should be more stable now. I've been a lot of bugfixes and some new features to make things easier, especially set up. The APIs won't be changing as much either, and I'll be deprecating stuff rather than just removing it from now on.

Changes in cookiecutter-django:

* hitchreqs.txt is updated to include all a multitude of bug fixes (and some new features).
* There are now three settings files - one for CI, one for test driven development and a global one: all.settings.
* all.settings contains the version of python the user chose - either 2.7.10 or 3.5.0.
* Django / Celery versions are no longer checked. They are just assumed to be correct now.
* I removed stub.test and replaced it with an actual test that does something now - it follows the sign up form, clicks on link in email and logs in.
* To accomodate that test and to keep it readable I added some HTML IDs on the elements it was interacting with which didn't have elements. I hope that's okay.
* There's a system.packages file containing a unix-independent list of package names required to run the code - e.g. libjpeg8-dev. Hitch should ask politely to install these during initialization if they are not already installed, along with any packages it might need.
* Added .hitch directory to gitignore.

You can try all of this out just with the following five commands::

    cookiecutter git@github.com:crdoconnor/cookiecutter-django.git
    pip install --upgrade hitch (works fine with sudo/pipsi as well)
    cd my_project/tests/
    hitch init
    hitch test register-and-log-in.test

Note:

* hitch init may try to install some packages with apt-get, yum, pacman or brew. If it needs sudo it should ask nicely first.
* hitch test will download and compile its own packages (python, postgres, redis) on the first test run and put them in ~/.hitchpkg, but it won't ask for sudo like hitch init. First test run takes 25 minutes in my VM. Subsequent test run takes < 45 seconds. A powerful machine will be much faster.
* you can remove all of those downloaded packages (python, postgres, redis) with the command "hitch cleanpkg" and remove the .hitch directory with "hitch clean". No other directories are used.

I'm still doing a lot of regression testing on the latest version of the framework under a variety of conditions but this all seems to be working for now. I've tested on Mac OS X Yosemite and Ubuntu so far. I'm doing other Linux distros tonight.

Nonetheless, there might still be bugs lurking. Please tell me if you find any.

Additional environment tests are all very welcome. If anybody can try this on El Capitan and tell me if it works I'd be very grateful. I can't test that unfortunately.

Comments, advice, requests for changes, etc. all welcome.

Thanks once again @pydanny, @audreyr, @luzfcb and anybody else I've missed for creating cookiecutter-django.


luzfcb:
@crdoconnor Thanks again for your hard work.

Some minor questions:
1 - if now Hitch uses [`unixpackage`](https://github.com/unixpackage/unixpackage), I believe that [this](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/requirements.apt#L26-L44) is no longer necessary.

( I believe that at some point in the future when `unixpackage` is sufficiently stable, we can remove my [install_os_dependencies.sh](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/install_os_dependencies.sh) bash script . I think that `unixpackage` solves pydanny/cookiecutter-django/issues/34 in a better way ) 

I personally do not like the package python-pip and python3-pip, because linux distributions send very old pip (Ubuntu 15.04 delivery pip 1.5.6), because very confusing for beginners (and not so beginners as well) able to upgrade it correctly and some related problems: pypa/pip/issues/2936, pypa/pip/issues/2973 pypa/pip/issues/3164

2 - What is the behavior of `unixpackage` when a package is already installed, however, there is an update for it in the repository?
a - unixpackage install a new version of the already installed packages?
b - informs the user that the package is already installed, and asks if he can be upgrade to the new version
c - does nothing. Install another packages but do not install the new version of the package can be upgraded.

I would expect the default `unixpackage` behavior would be option c, and informs the user that the package has a new version available in the repository and informs the user that he can update re-run `unixpackage` with `--foo` flag (I suggest  `--upgrade` ). This case has reported by @dupuy on https://github.com/pydanny/cookiecutter-django/pull/200#discussion_r26436046


@pydanny, @audreyr, @theskumar @jayfk LGTM :+1:

crdoconnor:
>if now Hitch uses unixpackage, I believe that this is no longer necessary.

Is it used elsewhere?

Incidentally, unixpackage is stable but it has some quirks that I'm not necessarily fond of that I put in for expediency's sake. For instance, for firefox, virtualenv and node-less it always presumes that they are already installed on the mac since none of them can actually be installed via brew.

Similarly, there are system package managers missing. I don't have macports, emerge, zypper, urpmi or pkg. Also it doesn't deal with pip or npm.

To be honest, though, I'm reluctant to open this can of worms up. At least not now. It does at least solve the main problem of installing those 20 core packages you need but which are named differently on every system.

>I personally do not like the package python-pip and python3-pip, because linux distributions send very old pip

Yeah, me neither, although I'm loath to actually upgrade it in case that violate's my system's expectations somehow. This is partly why hitch circumvents system pip entirely except for the bootstrapper.

>What is the behavior of unixpackage when a package is already installed, however, there is an update for it in the repository?

It's c (thanks for reporting that bug a few weeks ago, by the way). I'm still on the fence about whether I want a warning.

However, there's a related problem I have yet to solve but is on the roadmap: what if the software installed on the system is the wrong version for the test?

Take libjpeg for instance. The latest ubuntu version is 8. The debian stable version is probably 5 or something. Currently version 9 is available but not in any package managers (except maybe arch). Your software could require it to be >=5, ==8, or even >=6 *and* <= 9. Ideally I guess you should have the same version running which you do in production. For some projects it might matter that you don't.

This is the kind of thing I designed the environment checkers to handle. They run at the beginning of each test. They explicitly don't require root. They don't do anything except cause tests to fail:  https://hitchtest.readthedocs.org/en/latest/api/environment.html

For checking the version of software installed I would probably bypass the system package manager entirely and use ldconfig, or "packagecommand --version".

PR #385, opened 2015-10-22
Use COPY instead of ADD in Dockerfiles

audreyr:
I think https://docs.docker.com/articles/dockerfile_best-practices/#add-or-copy is suggesting to use COPY instead of ADD if it suffices.

Asking for a review rather than committing directly because I'm not 100% sure.


jayfk:
Is caching working with `COPY`? As far is I remember it was broken for a specific version but I can't find the issue on docker upstream to see if it has been fixed.

jayfk:
Works!

jayfk:
Merging this in. I have a couple of docker related changes in mind that can build up on this. Will update the changelog accordingly.

PR #391, opened 2015-10-31
Fix gevent requirement for Python 3

mcho421:
gevent==1.0.2 isn't compatible with Python 3. 
gevent==1.1b6 is a beta version but it is compatible. 
I used a conditional to replace gevent==1.0.2 with gevent==1.1b6 if use_python2 = 'n'.


audreyr:
Great, thanks @mcho421.

PR #392, opened 2015-11-02
Add setup_venv.sh ; Helper script to setup virtual env for cookiecutter based project

webanz:
Add  setup_venv.sh script . 
The script helps to setup virtualenv for django-cookiecutter based project.

### DIFF
--- a/{{cookiecutter.repo_name}}/setup_venv.sh
+++ b/{{cookiecutter.repo_name}}/setup_venv.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+
+# NAME parameter of each environment is used to define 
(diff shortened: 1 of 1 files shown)


pydanny:
Thank you @webanz, but with my apologies we have to turn down this pull request. Setting up `virtualenv` is very dependent on which Python you are running, how you install pip/virtualenv, and the operating system. Maintaining this would be extremely challenging for the maintainers, as we would have to accommodate a shell script that would only grow with complexity.

webanz:
Yes, I understand the problem, I am looking for a python library for managing  virtualenvs consistently across platforms.  Thanks for Django Cookie Cutter.

PR #394, opened 2015-11-03
Fix python version issue in .travis.yml

show0k:


noisy:
@show0k it seesm that this right now cause some problems: https://travis-ci.org/SpisTresci/SpisTresci/builds/122075842

PR #397, opened 2015-11-06
changed ADMIN_URL to DJANGO_ADMIN_URL in env.example

ChrisPappalardo:
DJANGO_ADMIN_URL is an expected environment variable which is used to set ADMIN_URL.  The current env.example fails due to the missing DJANGO_ADMIN_URL variable.


jayfk:
thanks!

PR #403, opened 2015-11-11
fix display update button & change name to nickname

areski:
Few things here:
a) Fixing the display of the Update button
b) Changing new field `name` to `nickname`, I found it quite confusing as this is very close to username field and so in order to people how to extend the User model it might be good to consider making it clearer.
c) Add first_name and last_name to the user update form


jayfk:
First name and last name is a specific thing for western countries, but there are cultures where there is no such thing as a last name or first name. That's why the more generic `name` is used:

```
First Name and Last Name do not cover name patterns	 around the globe.
```

PR #404, opened 2015-11-12
Update django-aullauth to latest release

yunti:


luzfcb:
thanks again @yunti 
:dancers: :+1:

yunti:
There seems to be an issue with one of the templates, so far using 0.24.1 of django all-auth.
..    
    Exception Type:	TemplateSyntaxError
    Exception Value:	'url' is not a valid tag or filter in tag library 'future'
    in /templates/account/signup.html
    This tag ...  {% load url from future %}

yunti:
Yup ok .. the templates also need to be updated from django-allauth.  I'll update when I get the chance later.  At least it's now compatible with django 1.9 (b1)

luzfcb:
@yunti from [Django 1.9 release notes](https://docs.djangoproject.com/en/dev/releases/1.9/#features-removed-in-1-9): "ssi and url template tags are removed from the future template tag library."

simply remove any occurrence of ``{% load url from future %}`` in templates, solve the problem.

yunti:
Yup I had just been testing that and it works ok.  (And realised that the templates in allauth also had no change since the last update).

PR #408, opened 2015-11-16
cleanup of main readme

burhan:
Wanted to cleanup a bit the main readme to clarify which components are optional vs. those that are enabled by default.

It gets rid of the `(optional)` that kept repeating.


luzfcb:
Good. Thanks again @burhan

PR #409, opened 2015-11-16
Replace Maildump with MailHog

keybits:
Since Maildump is not compatible with Python 3, I've replaced it with [MailHog](https://github.com/mailhog/MailHog).

MailHog is written in Go so the binary has no dependencies and releases are available for most platforms.

I've tested this and it works well. I've also updated the docs and readme files where relevant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [2015-11-17]
 ### Added
 - initial configuration to support opbeat (@burhan)
+### Removed
+- Took *.pyc out of .gitignore, because it's already covered by *.py[cod] (@audreyr)
 
 ## [2015-11-16]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All enhancements and patches to cookiecutter-django will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2015-11-17]
+### Added
+- initial configuration to support opbeat (@burhan)
+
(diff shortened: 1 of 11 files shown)


burhan:
It has my vote; but why not use the docker file for a simpler installation and setup? As we are already using docker-compose, it might be a more inclusive option (rather than having to download yet-another-thing).

keybits:
@burhan Good point! I've not been using Docker with cookiecutter-django so didn't consider it. Looks like it should be easy enough to add with MailHog's provided Dockerfile.

I'll update my branch when I get a moment to include support for this.

keybits:
On second thoughts, I'll leave adding Docker support to someone more familiar with how 'Developing locally with Docker' is configured here.

Did Maildump ever work with Docker for local development? Both it and my implementation make use of the provided Grunt tasks (specifically grunt-bg-shell) to run various processes in parallel.

burhan:
I think we should get it in; and then do another PR to see how we can better utilize docker.

pydanny:
LGTM :+1: 

When the merge conflicts are resolved I'll merge it in.

keybits:
Thanks for the comments everyone. @pydanny should I do something about the merge conflicts, or leave that to you?

burhan:
I think your branch just needs a rebase.

keybits:
Eugh! Sorry - my rebasing went a bit haywire... so now have horrible history / inaccuarete commit messages. Not sure how best to tidy up. Any suggestions greatly appreciated.

pydanny:
@keybits no worries. While I believe people deserve credit for their efforts, I'm not religious about commit history. Especially when it blocks an important change to a project. :cat2:

PR #411, opened 2015-11-16
initial configuration to support opbeat

burhan:
I would like to propose including [opbeat](https://opbeat.com/languages/django/) as an alternative to newrelic for performance monitoring.

I personally find them a bit easier to get setup; and their free tier is also a lot more accommodating.

I am not against keeping both (yaay for choice!) if someone would like to test both options and then decide which one they want to keep.

The initial commit does not add logging configuration; which is still left up to sentry.


pydanny:
:+1: for getting us off newrelic and onto @opbeat

burhan:
@jayfk done

PR #432, opened 2015-12-05
[IMP] Allow enable or disable user registration using ACCOUNT_ALLOW_REGISTRATION setting.

ddiazpinto:
The intention of this improvement is to allow enable and disable user registration through `django-allauth`, using `ACCOUNT_ALLOW_REGISTRATION` setting but without loosing other characteristics like authentication and account management.

Projects that may benefit from this improvement are those that do not require user registration at all  but need the authentication of current users through other methods like social networks, and other projects that needs temporary disable the registration process, for example, not-ready projects.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -69,6 +69,7 @@ Code Contributors
 * Jan Van Bruggen / @jvanbrug
 * Jon Miller / @jondelmil
 * Thomas Korrison / @failsafe86
+* David Díaz / @DavidDiazPinto *
 
 \* Possesses commit rights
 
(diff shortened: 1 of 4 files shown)


pydanny:
Nicely done @ddiazpinto! Before I merge this, can you let me know what `[IMP]` means?

ddiazpinto:
Thank you @pydanny.
[IMP] means improvement. Normally I use [IMP] for improvements,  [ADD] for new content, [FIX] for fixes and [I18N] for translations or internationalization changes. ;)

audreyr:
Great work @ddiazpinto, thanks!

PR #433, opened 2015-12-08
Update common.py

thisismsreddy:
Adding a Sqlite db in comments if any case PostgreSQL not available


pydanny:
@thisismsreddy Thank you for your contribution, we very much appreciate it, but per https://github.com/pydanny/cookiecutter-django#constraints cookiecutter-django is PostgreSQL only.

PR #435, opened 2015-12-09
Bump Django to v1.9

areski:
Bump Django version to v1.9


pydanny:
Today I'll review this, cut a last tag for 1.8.x, then accept it in.

areski:
:+1:

PR #445, opened 2015-12-15
Fix celery worker app name in Procfile

stepmr:
- Celery worker was using incorrect app name


luzfcb:
@stepmr Thanks. :smiley:

Issue #446, opened 2015-12-16
In env.example, including DJANGO_SENTRY_DSN `if cookiecutter.use_whitenoise`


dsclementsen:
I feel like this ought to be `if cookiecutter.use_sentry`.

Is this an overlooked copy-paste, or is there a sentry-whitenoise relationship that I am unaware of?

PR #447, opened 2015-12-16
DJANGO_SENTRY_DSN env variable if use_sentry

dsclementsen:
Closes pydanny/cookiecutter-django#446


luzfcb:
@dsclementsen :ship:  thanks for catch this issue.

PR #456, opened 2015-12-31
added external resource to readme

mjhea0:


pydanny:
@mjhea0 I really like your article! However, I think links to external resources should be placed in their own `articles` section, not unlike how it is done in requests: http://docs.python-requests.org/en/latest/community/out-there/#articles-talks

If you could add header for `Articles` above your link, then we could make it the first curated article. :smile:

mjhea0:
@pydanny updated - thanks!

PR #457, opened 2016-01-02
Added a cdn link to Tether, which is a new requirement for Bootstrap 4

EricZaporzan:
More information [here](http://v4-alpha.getbootstrap.com/components/tooltips/#overview). Looks like Tether is needed for proper positioning of Bootstrap tooltips.


luzfcb:
Thanks @EricZaporzan

PR #460, opened 2016-01-05
Add dockerignore file

bogdal:
This PR adds `.dockerignore` file. Hidden files/dirs (especially `.git` directory) are unnecessary in the docker image.


jayfk:
Won't this also match the default `.env` file and thus break the production setting? Furthermore, some people maybe have a `.coverage` file for their test coverage that runs inside the container.

I'm totally on your side that we need a `.dockerignore` file, but I think it's better to be more precise about what we want to exclude.

bogdal:
@jayfk you're absolutely right. I've just added a few exceptions.

audreyr:
:+1: I think it looks good now

jayfk:
Yup, thanks!

PR #462, opened 2016-01-06
initial fixes for docker issue on windows 10

burhan:
Please test; I have done the least destructive fix - the previous fixes with `.gitignore` mungering were not reliable during testing.

I also realized that the `.editorconfig` is also forcing line endings to `LR` which will help if the editor being used supports editorconfig (most popular ones do via plugins).

### DIFF
--- a/{{cookiecutter.repo_name}}/Dockerfile
+++ b/{{cookiecutter.repo_name}}/Dockerfile
@@ -17,7 +17,7 @@ RUN chown -R django /app
 COPY ./compose/django/gunicorn.sh /gunicorn.sh
 COPY ./compose/django/entrypoint.sh /entrypoint.sh
 RUN sed -i 's/\r//' /entrypoint.sh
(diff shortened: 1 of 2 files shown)


{{cookiecutter.repo_name}}/Dockerfile
```diff
@@ -16,7 +16,8 @@ RUN chown -R django /app
 
 COPY ./compose/django/gunicorn.sh /gunicorn.sh
 COPY ./compose/django/entrypoint.sh /entrypoint.sh
-
+RUN sed -i 's/\r//' /entrypoint.sh
+RUN set -i 's/\r//' /guincorn.sh
```
jayfk: typo?

burhan: Yup, you are right. Late night edits.

jayfk:
LGTM. I've tested this on OS X and it works!

PR #463, opened 2016-01-08
[Experimental] Channels Support

jayfk:
I've been playing around with channels recently and thought this would be a good starting point to add channels to cookiecutter-django sometime in the future.

This is not intended to be merged into master right now. Consider it more like a playground for now until we have something stable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,6 +46,7 @@ Optional Integrations
 * Integration with Sentry_ for error logging
 * Integration with NewRelic_ for performance monitoring
 * Integration with Opbeat_ for performance monitoring
+* Experimental Channels_ support
 
 .. _alpha: http://blog.getbootstrap.com/2015/08/19/bootstrap-4-alpha/
 .. _Hitch: https://github.com/hitchtest/hitchtest
@@ -64,7 +65,7 @@ Optional Integrations
 .. _NewRelic: https://newrelic.com
 .. _docker-compose: https://www.github.com/docker/compose
 .. _Opbeat: https://opbeat.com/
-
+.. _Channels: https://github.com/andrewgodwin/channels
 
 Constraints
 -----------
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -16,5 +16,6 @@
     "use_newrelic": "n",
     "use_opbeat": "n",
     "windows": "n",
-    "use_python2": "n"
+    "use_python2": "n",
+    "use_channels": "n"
 }
--- a/docs/channels.rst
+++ b/docs/channels.rst
@@ -1,0 +1,29 @@
+Channels
+========
+
+Cookiecutter-django comes with (experimental) channels support. A basic skeleton of a Django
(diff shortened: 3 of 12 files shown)


luzfcb:
:smile_cat: \o/

audreyr:
This is exciting @jayfk! 😹

pydanny:
Having talked to @andrewgodwin, I think it's in our best interest to continue keeping this one on hold.

jayfk:
This one has diverged from master so much that there's no reason to keep it open any longer. Once channels is ready for prime time we can continue to tackle this beast.

ssteinerx:
@jayfk, @pydanny, @audreyr I think the best approach on this would be as a `mixin`, similar to the way React/Webpack was added (and then removed), except without the adding.

A `cookiecutter-django mixin, the way I see it, adds into the `cookiecutter-django` structure (and possibly modifies configuration files etc.) but is *not* included in the project.  

You cookiecutter in your main project, then *mixin* other features  They stay smaller, better focused, and don't bloat cookiecutter-django with stuff most projects do not need.

I'm using Channels for a current project and thinking about this mixin approach quite a bit...

jayfk:
I think that has to be decided on a case by case basis. The webpack integration as a plugin (or, as you call it, mixin) was a good idea because it was loosely coupled and pulled in a lot of unrelated files.

ssteinerx:
A plugin implies an architecture that supports integrating functionality into a product in a *call out* sort of way.  E.g. a *markdown mixin* could allow you to write blog posts using Markdown instead of HTML and would translate for display in HTML.

A mixin adds to e.g. a class directly ala the class django.views.generic.base.ContextMixin.

The architecture I'm proposing mixes things into a `cookiecutter` project, it does not require the original project to *support anything*.

It's more like making chocolate chip cookies from the same dough than going crazy and making [these](http://www.loveandoliveoil.com/2016/04/marzipan-stuffed-almond-sugar-cookies.html)

ssteinerx:
The more I think about this, the more I like the idea of pulling things out of this project into mixins and leaving this project as pure scaffolding with a consistent *best practice* approach.

I like this idea so much that I'm pulling this off to a [separate project](https://github.com/ssteinerx/cookiecutter-django-mixin) right now.

adminq80:
Hi guys, I am working on a project with `channels` and I am using `cookiecutter docker containers` for the deployment. Currently, I changed Nginx configuration to route Websocket to the server, but django still doesn't recognize the incoming connections as websocket. 

I really need help with this,
Abdulrahman

ssteinerx:
@jayfk -- since the branch is deleted, it's probably time to close this.  We'll open a new one when there's something concrete to discuss.

@adminq80 Since Channels isn't even integrated into cookiecutter-django yet, your request for help would be much better directed at the Channels IRC chat, Stack Overflow, etc.

adminq80:
@ssteinerx thanks

PR #468, opened 2016-01-13
Update setup.py

Collederas:
Just updating the version number as stated in the comment. The procedure says you want to branch and tag so I leave this pull request as a simple reminder ;)


audreyr:
Thanks @Collederas.

PR #469, opened 2016-01-13
Adds celerybeat db to gitignore

originell:
When testing "production mode" locally, celery of course generates it's task scheduling database for `celery beat`. It doesn't really make much sense to track this with any VCS :)

Might also be nice to add this to #460 ?


audreyr:
Thanks @originell!

originell:
Yay :yellow_heart: so nice :D

PR #470, opened 2016-01-15
Modified gruntfile.js and dev.yml to use grunt when using docker in development

mjsisley:
Slight change so that grunt is used when developing with docker container.

### DIFF
--- a/{{cookiecutter.repo_name}}/Gruntfile.js
+++ b/{{cookiecutter.repo_name}}/Gruntfile.js
@@ -110,7 +110,7 @@ module.exports = function (grunt) {
         bg: true
       },
       runDjango: {
-        cmd: 'python <%= paths.manageScript %> runserver'
+        cmd: 'python <%= paths.manageScript %> runserver_plus 0.0.0.0:8000'
       },
       {% if cookiecutter.use_mailhog == "y" -%}runMailHog: {
         cmd: './mailhog'
--- a/{{cookiecutter.repo_name}}/dev.yml
+++ b/{{cookiecutter.repo_name}}/dev.yml
@@ -8,7 +8,7 @@ postgres:
 django:
   dockerfile: Dockerfile-dev
   build: .
-  command: python /app/manage.py runserver_plus 0.0.0.0:8000
+  command: grunt serve
   volumes:
     - .:/app
   ports:


jayfk:
Thanks for this pull request, but I'm afraid this needs more thought.

If you don't use docker and run a project based on this pull request where `runserver_plus` listens on all interfaces, you open your machine for remote code execution.

That may be okay if you do development at home where you have your network in a *DMZ*, but if you sit in a cafe and are connected via wifi that's not so good.

Besides that, what is `grunt serve` supposed to do in `dev.yml`? `grunt` is currently not installed in the docker image, so this will yield an error.

If you are interested in bringing `grunt` to docker, I'd like to invite you to join the [cookiecutter-django channel](https://gitter.im/pydanny/cookiecutter-django?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) at gitter where we can discuss this and work on a solution together.

mjsisley:
Thank you for the explanation.  I currently have grunt working within docker for myself (explained below).  If you think it is worth it, then I'm more than happy to help arrive at the best solution for the general case.

I did this for myself a while ago, and foolishly missed these lines within Dockerfile-dev for the pull request. 

```
RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
RUN apt-get install -y nodejs
RUN npm install -g grunt-cli  
```

The way I currently use it is as follows:

- Start new cookiecutter project.
- Follow the process here http://cookiecutter-django.readthedocs.org/en/latest/live-reloading-and-sass-compilation.html locally one one computer:
- git commit with node_modules.
- Clone repo on other development computers and run ```'docker-compose -f dev.yml up'```.

I originally set it up so http://cookiecutter-django.readthedocs.org/en/latest/live-reloading-and-sass-compilation.html was done within the container (which also requires downloading ruby within the container), but it seemed unnecessary to make the container that heavy when the above procedure works.  Version tracking the node_modules is not ideal, perhaps there is a better approach.

I was not aware that runserver_plus was causing this issue, thank you.  I had assumed specifying the 8000 port meant that only that port was exposed, but I should have looked into that more.  I also have only been using it inside docker, and obviously that can't be assumed here.  The easiest solution would seem to me to be to add a 'serve-docker' task to gruntfile.js to be called in dev.yml.

mukgupta:
Grunt works out of the box even while developing with docker. You only need to run `grunt serve ` separately along-with `docker-compose -f dev.yml up` .

PR #487, opened 2016-02-16
Renamed adapter file to match django convention.

Eraldo:
Renamed the `adapter.py` file to `adapters.py` to match the _django_ naming convention.
It also makes sense since there are two different adapters in that file (normal and social).


jayfk:
:+1: 

Could you take a look at #488 again? Once you have that updated I can merge both PRs. That makes it easier when editing the history.

Thank you!

Eraldo:
Sure thing..
✓ Added account registration switch to environment example file.

PR #495, opened 2016-03-02
typos and cleanup

highpost:
- changed the "heroku create" syntax to match what's used in the buildpack doc
- removed the first DJANGO_ADMIN_URL, it's probably a typo and it's overridden by the second def
- changed DJANGO_SETTINGS_MODULE to use local because that's more appropriate for an example and production is further down the road (given that, I would also remove DJANGO_ADMIN_URL since it has a [logical default for development](http://cookiecutter-django.readthedocs.org/en/latest/settings.html), but I left it intact)
- removed PYTHONHASHSEED since random is the default

### DIFF
--- a/docs/deployment-on-heroku.rst
+++ b/docs/deployment-on-heroku.rst
@@ -7,7 +7,7 @@ You can either push the 'deploy' button in your generated README.rst or run thes
 
 .. code-block:: bash
 
-    heroku create --buildpack https://github.com/heroku/heroku-buildpack-python
+    heroku create --buildpack heroku/python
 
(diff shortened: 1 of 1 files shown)


pydanny:
Thanks for the pull request.

We intentionally randomize the admin URL as a security measure. We won't accept a non-random setting for that URL.

jayfk:
> We intentionally randomize the admin URL as a security measure. We won't accept a non-random setting for that URL.

Right. Furthermore the deployment settings are replaced with those used for development.

Again, thanks for the pull request but I'm afraid we have to close it.

Issue #254, opened 2015-06-26
The MEDIA_URL and STATIC_URL settings must have different values (Production settings issue)


cdvv7788:
Django is complaining about the paths being the same (which is the default on the template). I fixed this on my settings with:

MEDIA_URL = 'https://s3.amazonaws.com/%s/media/' % AWS_STORAGE_BUCKET_NAME
STATIC_URL = 'https://s3.amazonaws.com/%s/static/' % AWS_STORAGE_BUCKET_NAME

instead of:

MEDIA_URL = 'https://s3.amazonaws.com/%s/' % AWS_STORAGE_BUCKET_NAME
STATIC_URL = MEDIA_URL

What is the purpose of treating them as the same as default? Is this issue showing up for everybody?

luzfcb:
Oh, really. Thanks for reporting this.

@theskumar  this was introduced in https://github.com/pydanny/cookiecutter-django/commit/7fb6f94d5e8b41e4f407c5aacac6fbe452b28a7e#diff-26beb9938fe6c9635e2c57dd6b0f1c5bL87 .
Are there any special reason for this configuration?

or a small mistake which went without anyone see

![Your Unit Tests are Passing ](http://cdn.makeagif.com/media/6-26-2015/jpdwZQ.gif)

theskumar:
> What is the purpose of treating them as the same as default? 
To answer your question, there is no specific reason. 

> Is this issue showing up for everybody?
I don't see it. @cdvv7788 which version of Django are you using and how is it possible to see that complain. `manage.py check --deploy`?

theskumar:
> @theskumar  this was introduced in https://github.com/pydanny/cookiecutter-django/commit/7fb6f94d5e8b41e4f407c5aacac6fbe452b28a7e#diff-26beb9938fe6c9635e2c57dd6b0f1c5bL87 .
> Are there any special reason for this configuration?

It was introduced to make sure that `MEDIA_URL` and `STATIC_URL` are different when using whitenoise. When using whitenoise STATIC_URL should point to a relative `/static/` url which is inherited from `common.py`, instead of fully qualified domain name to `http://s3.amazon.com/...`

cdvv7788:
I am using django1.8.2...that setting is enforced since 1.4 i think.

pydanny:
Is this still an issue?

cdvv7788:
[...]

PR #496, opened 2016-03-02
Make sure that STATIC_URL != MEDIA_URL

cdvv7788:
https://github.com/pydanny/cookiecutter-django/issues/254
This should ensure that STATIC_URL stays the same as it was if using whitenoise, but if whitenoise is not being used it will point to an specific folder inside S3. MEDIA_URL will change in this case too.
If there is a better approach to this, that would be nice, but currently i need to modify some access per folder and with the current setup everything is mixed at the root of the bucket, which is not good.


hilltops:
+1

andresgz:
+1

jlariza:
+1

jayfk:
LGTM

Issue #483, opened 2016-02-08
Move to new docker-compose format


jayfk:
`Docker 1.10` is out and there's a new [compose file format](https://docs.docker.com/compose/compose-file/#versioning:91de898b5f5cdb090642a917d3dedf68). This brings a lot of goodies, so we should switch as soon as possible.

PR #497, opened 2016-03-03
updates docker-compose format to version 2 #483

aeikenberry:


aeikenberry:
Darn. Looking at why the test fails.

jayfk:
:+1: 

The test probably fails because travis isn't using docker 1.10 yet.

Adding this to `before_install` should do the trick:

    - sudo apt-get update
    - sudo apt-get install -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" docker-engine=1.10.1-0~trusty

aeikenberry:
:100: Thanks that fixed it.

jayfk:
Thanks!

PR #506, opened 2016-03-23
Update LICENSE template

romanosipenko:
Fix condition check in LICENSE file, which prevent cookiecutter to generate this file.


luzfcb:
@romanosipenko thanks for catch this. :ship:

PR #507, opened 2016-03-23
Pycharm Support (including debugging in Docker)

noisy:
New implementation of https://github.com/pydanny/cookiecutter-django/pull/306

* [x] proper settings in _Language & Frameworks > Django_
* [x] Default Python Template Language set to Django instead of Jinja2 in _Language & Frameworks > Python Template Language_
* [x] `node_modules` directory added to excluded (will not be indexed)
* [x] Example Run/Debug Configurations for runserver, runserver_pluse, migrate
* [x] Example Run/Debug Configurations for tests (all, from module, from file, from class, specific)
* [x] `requirements/local.txt` marked as default for development in _Tools > Python Integreated Tools_
* [x] documentation - [pycharm/configuration.rst](https://github.com/SpisTresci/cookiecutter-django/blob/pycharm_support_2/%7B%7Bcookiecutter.repo_name%7D%7D/docs/pycharm/configuration.rst)

PR #517, opened 2016-04-10
Deleting unnecessary .idea dir from MAIN directory

noisy:
It seems that this file was committed by mistake in #507. Probably was confused with correct file `{{cookiecutter.repo_name}}/.idea/vcs.xml`


audreyr:
Good catch 👍

Issue #476, opened 2016-01-22
Make Grunt optional?


jayfk:
If we want to bring grunt to docker (#310), it would be good to make grunt an option first.

Why?
The grunt task comes with a full `ruby` and a `node` stack that needs to be run alongside the Django container. If we made grunt an option, we could make sure that ruby/node only have to be installed when necessary.

Thoughts?

yunti:
Yes this would be very good to have as an option. Tooling probably varies the most at the moment and it would be to have this optional to cover a wider range of options. Particularly as having ruby and node included that is a lot of unnecessary  requirements that wouldn't be needed by many.

theskumar:
Yes. Grunt should be optional.

burhan:
Why not replace it with a Python task runner?

pydanny:
Yes, would love to have grunt made an option. Especially considering #538.

In fact, I think a list of tooling options would be ideal. Something like:
```
Choose a tooling option:

1. Webpack
2. Grunt
3. ???
4. No front end tooling

audreyr:
This was done in #516.

PR #516, opened 2016-04-08
Major Project Generation Cleanup

jayfk:
This is a major cleanup regarding project generation. 

It introduces four new `cookiecutter.json` options:

| option  | default  |
|---|---|
| use_docker  |  yes |
| use_heroku  |  no |
| use_grunt  |  no |
| use_angular | no |

## Option use_docker
If docker isn't going to be used it removes:
  - dev.yml
  - docker-compose.yml
  - .dockerignore
  - compose/

Furthermore, it displays a warning during project generation if docker and mailhog/grunt are selected. Both options are not implemented, yet.

## Option use_heroku
If heroku isn't going to be used it removes:
  - app.json
  - Procfile
  - requirements.txt (discuss?)
  - runtime.txt

## Option use_grunt
If grunt isn't going to be used it removes:
  - Gruntfile.js
  - package.json

## Option use_angular
Removes `ng-app` and the `angular` block in `templates/base.html` if angular isn't going to be used.

## Random String Generation

Instead of failing with an absurd error message that the `django` config can't be imported if no pseudo-random number generator is available, it now warns the user to change the `SECRET_KEY`.

## OS Dependencies

Moved `install_os_dependencies.sh`, `install_python_dependencies.sh` and `requirements.apt` into `utiliy/`.

Addresses:
 - [#476: Make Grunt optional?](https://github.com/pydanny/cookiecutter-django/issues/476)
 - [#474: Make Docker and Heroku cookiecutter.json options?](https://github.com/pydanny/cookiecutter-django/issues/474)
 - [#439: Consideration: Move all config files to a specific location](https://github.com/pydanny/cookiecutter-django/issues/439)
 - [#310: Grunt support for docker (partly by displaying a warning)](https://github.com/pydanny/cookiecutter-django/issues/310)
 - [#308: Consider removal of dev.yml](https://github.com/pydanny/cookiecutter-django/issues/308)
 - [#307: Make Angular.js an option, rather than a default](https://github.com/pydanny/cookiecutter-django/issues/307)
 - [#475: Shouldn't the dockerfile take care of installing mailhog? (partly by displaying a warning)](https://github.com/pydanny/cookiecutter-django/issues/475)

PR #518, opened 2016-04-10
Mailhog as docker-container

noisy:
I created this PR, because I don't like the idea, that I have to download additional binaries:

from https://github.com/pydanny/cookiecutter-django/blob/master/docs/developing-locally.rst

> In development you can (optionally) use MailHog for email testing. MailHog is built with Go so there are no dependencies. To use MailHog:
> 
> Download the latest release for your operating system
> Rename the executable to mailhog and copy it to the root of your project directory
> Make sure it is executable (e.g. chmod +x mailhog)
> 

I think it is also better to not use grunt for running mailhog, and because also there are plans to [make grunt optional](https://github.com/pydanny/cookiecutter-django/issues/476) (https://github.com/pydanny/cookiecutter-django/issues/476), this will remove unnecessary dependency

### DIFF
--- a/docs/developing-locally.rst
+++ b/docs/developing-locally.rst
@@ -42,14 +42,12 @@ django-allauth sends an email to verify users (and superusers) after signup and 
 
 .. _configure your email backend: http://docs.djangoproject.com/en/1.9/topics/email/#smtp-backend
 
-In development you can (optionally) use MailHog_ for email testing. MailHog is built with Go so there are no dependencies. To use MailHog::
+In development you can (optionally) use MailHog_ for email testing. MailHog is added as docker-container. To use MailHog::
 
-1. `Download the latest release`_ for your operating system
-2. Rename the executable to ``mailhog`` and copy it to the root of your project directory
-3. Make sure it is executable (e.g. ``chmod +x mailhog``)
+1. Make sure, that ``mailhog`` docker container is up and running
+2. Open your browser and go to ``http://127.0.0.1:8025``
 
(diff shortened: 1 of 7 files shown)


mjsisley:
+1 for using this in master. I am using this in development and it works well. 

One thing is that if docker is going to be an option (per #516) , maybe all of the changes you made should also check if docker has been selected so that if someone elects to use mailhog and doesn't elect to use docker, they still can use mailhog the old way.

jayfk:
LGTM. Could you do a proper rebase?

noisy:
@jayfk rebased
@mjsisley I have added use_docker guards...

...but I am wondering what we should do if someone will choose:

    "use_mailhog": "y",
    "use_docker": "n",
    "use_grunt": "n",

burhan:
Grunt is not really a requirement here; its just there to facilitate running jobs and can easily be replaced with a `README` note.

Regarding the requirement for using docker, again I feel documentation can handle this.

noisy:
@burhan 
> Regarding the requirement for using docker, again I feel documentation can handle this.

I am not sure, what you mean by that. Should I add something to documentation?

jayfk:
> I am not sure, what you mean by that. Should I add something to documentation?

I think a small hint what to do when using mailhog w/o grunt/docker should be enough.

pydanny:
Should we merge this in and then document it shortly? I would love to see this added.

shireenrao:
This is would be a really nice enhancement. Let me know If I can help out in anyway to get this added.

PR #519, opened 2016-04-12
Quote consistency

blopker:
The PR is to clean up the use of double vs. single quotes. I know it's not very interesting, but I use this project a lot and I kept going OCD on all the double quotes! There was also a case of using single quotes for docstings.

I also added unicode_literals to a couple files where I thought it would be useful and cleaned up the `u`s.

Cheers! 😄

### DIFF
--- a/{{cookiecutter.repo_name}}/config/settings/production.py
+++ b/{{cookiecutter.repo_name}}/config/settings/production.py
@@ -1,23 +1,23 @@
 # -*- coding: utf-8 -*-
-'''
+"""
 Production Configurations
 
 - Use djangosecure
 - Use Amazon's S3 for storing static files and uploaded media
 - Use mailgun to send emails
 - Use Redis on Heroku
-{% if cookiecutter.use_sentry == "y" %}
+{% if cookiecutter.use_sentry == 'y' %}
 - Use sentry for error logging
 {% endif %}
-{% if cookiecutter.use_opbeat == "y" %}
+{% if cookiecutter.use_opbeat == 'y' %}
 - Use opbeat for error reporting
 {% endif %}
-'''
+"""
 from __future__ import absolute_import, unicode_literals
 
 from boto.s3.connection import OrdinaryCallingFormat
 from django.utils import six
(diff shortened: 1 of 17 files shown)


{{cookiecutter.repo_name}}/config/settings/production.py
```diff
@@ -118,7 +118,7 @@
 
 # URL that handles the media served from MEDIA_ROOT, used for managing
 # stored files.
-{% if cookiecutter.use_whitenoise == 'y' -%}
+{% if cookiecutter.use_whitenoise == "y" -%}
```
noisy: why here double quotes?

blopker: When it comes to the cookiecutter template this project uses double quotes. I just changed these so it's consistent.

pydanny: The mix of quotes and double-quotes is by me. Accidentally. I process those the same way, hence why my projects mix them. If you want to make the project consistent, then those should remain as single quotes.

blopker: My main contention is that the cookiecutter templates are used throughout the codebase in other non-Python files. For instance in a JSON file it would be consistent on a file level to use double quotes everywhere instead. 

I guess every file in this repo is a cookiecutter template first and another file type second. Like there can be a consistency 'boundary' between the two file types where as a template, double quotes make sense, and as a Python file single quotes make sense. 

Maybe I'm thinking about this too much. 😄 I'd like to get this merged though. Do you want me to just make all quotes in Python files single or keep the template statements double quoted?

pydanny: You are overthinking it. :D

blopker: Fair enough! Double quotes be gone.

pydanny:
Want to rebase this so I can merge it in?

blopker:
Rebaseness. Cheers!

PR #530, opened 2016-04-21
Remove duplicate clickjacking protection

mcoconnor:
Currently this project includes both Django's builtin `XFrameOptionsMiddleware` as well as enabling django-secure's `SECURE_FRAME_DENY` feature.  As noted in [django-secure's documentation](https://django-secure.readthedocs.org/en/latest/settings.html#secure-frame-deny), including both options is unnecessary (and is problematic for views that need to be exempted).


audreyr:
Great catch, appreciate this @mcoconnor.

PR #536, opened 2016-04-30
Small fixes to utility scripts

scast:
Hello there,

This PR basically fixes two little bugs:

- `utility/install_python_requirements.sh` was trying to install `requirements.txt` even if project was created `use_heroku == 'n'`

- both utility scripts depended of being called from undocumented and inconsistent paths. Particularly, `install_python_requirements.sh` only worked if it was called from the top of the project, and `install_os_dependencies.sh` only worked if called from inside `utility/` folder. Now you can call them from wherever you want. 

Cheers!


pydanny:
Thanks for this! 😆

PR #540, opened 2016-05-08
Updated whitenoise configuration to match changes in version 3.0

trungdong:
(see http://whitenoise.evans.io/en/stable/changelog.html)


audreyr:
Thanks @trungdong

PR #541, opened 2016-05-09
Added a comment on how to use mailhog

shireenrao:
As somebody new to mailhog, I was a confused on how to use it. I first assumed, it was going to be used by the django app itself. After reading more on the internet I found it needed to be executed outside of the django application. I feel the lines I added may help other new users.

Thanks
-Srini


luzfcb:
@shireenrao thanks. :+1:

PR #551, opened 2016-05-23
Install python3-dev when not using python2

bloodpet:
By default, ubuntu-based distros would install python2-dev if you install python-dev.
The package python3-dev should be installed explicitly if we're not using python2.

I've encountered this problem in two cases:
- When spinning up a new distro and starting a cookiecutter-django project with python2="no"
- When cookiecutter-django in a distro where python2-dev is already installed, but I've selected the option python2=no


pydanny:
Many thanks @bloodpet!

PR #552, opened 2016-05-24
Update developing locally docs on how to setup .env file

shireenrao:
Even though a link to the django-environ is there, it may help to see that .env file needs to be present for it to work.


audreyr:
Thanks @shireenrao, we love documentation improvements like this.

PR #553, opened 2016-05-24
Setting up new project with mailhog without docker breaks use of mailhog 

shireenrao:
The recent addition of mailhog as a docker container broke the template if one wanted to use mailhog without docker locally. As EMAIL_BACKEND was set, mailhog was not getting the emails from the django app. I fixed it by adding special cases in {{cookiecutter.project_slug}}/config/settings/local.py. I have now tested all 4 combinations of mailhog (y or n) and docker (y or n) and can confirm that they all work. As mailhog now works with docker, I also removed the warning message in hooks/post_gen_project.py which said mailhog and docker were not compatible.
The last change I made was a cosmetic one in dev.yml to remove unnecessary whitespace.

PR #555, opened 2016-05-24
Added a comment on mailhog for local docker development

shireenrao:
Now that mailhog is available as a container for local docker development, this note should help people on how to use it

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -86,6 +86,20 @@ To migrate your app and to create a superuser, run::
 
 Here we specify the ``django`` container as the location to run our management commands.
 
+Setup your email backend
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+django-allauth sends an email to verify users (and superusers) after signup and login (if they are still not verified). To send email you need to `configure your email backend`_
+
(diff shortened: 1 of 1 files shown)


shireenrao:
closing this PR as it is merged with PR #553

docs/developing-locally-docker.rst
```diff
@@ -86,6 +86,20 @@ To migrate your app and to create a superuser, run::
 
 Here we specify the ``django`` container as the location to run our management commands.
 
+Setup your email backend
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+django-allauth sends an email to verify users (and superusers) after signup and login (if they are still not verified). To send email you need to `configure your email backend`_
```
Afrowave: Towards the end of this line, I would swop, `To send email ...` with line 96, `In development ...`.

That way the idea would flow from development to production. I found that setting up e. g. Anymail with Mailgun as a default for development to be too much of a hassle. I am imagining that this is what it seems to say here since it is under *Development Mode*.

shireenrao: Hello Afrowave
All instructions in this document are only for local development using docker.

PR #557, opened 2016-05-26
Update common.py

thisismsreddy:
Adding a Sqlite db in comments if any case PostgreSQL not available


thisismsreddy:
nothing

pydanny:
Much as i appreciate the pull request, PostgreSQL everywhere. Runs on every operating system, including windows. I would rather have PostgreSql installation instructions then deal with people who had problems with either deploying sqlite or RAM into database engine descrepencies.

Issue #342, opened 2015-09-22
Django Compressor and Cookiecutter-django


andresgz:
Hi there,
I remember in previous versions "Django Compressor" used to com by default. Is there any reason to don't be included anymore?
I configured the Django-Compressor in a recent project, but It did not work  in production. It requires some additional configuration to make it work with Django Storages and S3.

Makes sense to include compressor again?

jayfk:
Absolutely, take a look at #310, #300 and #286 which more or less all come down to `django-compressor`. 

Maybe [django-pipeline](https://django-pipeline.readthedocs.org/en/latest/) can be another option. Haven't tried it yet, but the features look promising.

luzfcb:
I believe this is indirectly related to #301 .

andresgz:
Thanks @jayfk for responding.
I have checked django-pipeline and seems that configuration could be more complex and could require more modifications to the Cookiecutter. In general terms, django-compressor would be more intuitive.

andresgz:
I configured my last project including Django Compressor, and works great. If you think It's useful, I could make a pull request to include it by default. What do you think?

pydanny:
@andresgz, if it's not too late, please open a pull request! 😄

andresgz:
Sure @pydanny I will open a pull request

pydanny:
@andresgz Are you working on this today? If not someone here at the [PyCon 2016 Cookiecutter sprints](https://github.com/audreyr/cookiecutter/wiki/PyCon-2016-Cookiecutter-Sprint) says they are willing to do it. Either way, you get credit for participating.

andresgz:
Yes, I'm on that! 
I will post any question I find during the process.

pydanny:
Thanks @andresgz!

andresgz:
I have created the pull request #568 
Please let me know if you found any issue.

pydanny:
Merged.

PR #568, opened 2016-06-02
Add django-compressor suport

andresgz:
This PR is intended to solve issue #342 to add support to django-compressor.
It includes a new option to choose it (disabled by default).
According to the selection it will inject the necessary configuration in settings files and base template.

This PR has not been fully tested for all the cases, so if you find any issue, please let me know.


{{cookiecutter.project_slug}}/config/settings/common.py
```diff
@@ -185,6 +185,8 @@
     'django.contrib.staticfiles.finders.AppDirectoriesFinder',
 )
 
+# 'compressor.finders.CompressorFinder',
+
```
luzfcb: put inside the `{% if cookiecutter.use_compressor == 'y'-%}`

andresgz: thanks @luzfcb,
That was a redundant line, just removed it

luzfcb:
since we use Amazon S3 to staticfiles, I think this is also necessary: 
https://django-compressor.readthedocs.io/en/latest/remote-storages/#django-storages

probably it would also be useful to expose (via environment variable) some of these settings:

https://django-compressor.readthedocs.io/en/latest/settings/#base-settings

since I haven't used it so far, I do not know which of these settings would be good

if any new environment variable is added, It is also necessary to include an example of use in: https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/env.example


another question: django-compressor has no non-python dependency?
If have any non-python dependencies, it is also necessary to include them in [requirements.apt](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/utility/requirements.apt) and [requirements.apt.xenial](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/utility/requirements.apt.xenial)

andresgz:
Thanks @luzfcb 
I will add those options for S3 and will include the environment variables.
According to what i read in the documentation, it does not have system requirements.

pydanny:
I believe the Amazon S3 stuff is already defined in https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L139-L152. 

Therefore, all that needs to happen is to test that those values (or others in production.py) fulfill the needs of this pull request.

chrisdev:
@andresgz @pydanny in ``production.py`` we probably need to enable  offline compression see https://django-compressor.readthedocs.io/en/latest/scenarios/#offline-compression. 
Especially so if assets are in a CDN. 

We specify need to  ``COMPRESS_OFFLINE = True```

``COMPRESS_STORAGE = 'storages.backends.s3boto.S3BotoStorage'```

In this case we got to run ``manage.py compress``

chrisdev:
@andresgz @pydanny compressor can also handle the SASS work flow. No 
Something like this should be added to the base template.

```<link rel="stylesheet" type="text/x-scss" href="{% static 'css/app.scss' %}">```


and something like this  to ```production.py```

```
COMPRESS_CSS_FILTERS = [
    'compressor.filters.css_default.CssAbsoluteFilter',
    'compressor.filters.cssmin.CSSMinFilter',
]
```

chrisdev:
@andresgz @pydanny here are couple of additional dependencies that will leverage the ```django-compressor``` and get rid of Grunt from the point of view the SASS work flow https://github.com/torchbox/django-libsass and of course libsass

andresgz:
Thanks @chrisdev !
As a personal opinion I would like to leave the compressor configuration as simple possible. And after project is cooked, the user can extend the capabilities with other options.

pydanny:
@andresgz, simplest can be good, but we need an approach tied to tools already in this project, specifically SASS and other options @chrisdev mentions. 

In other words, we want/need an opinionated approach.

andresgz:
Hi guys! 
As a personal approach I prefer to use Gulp tasks for front-end development and compression of SASS and JS assets, to also to take advantage of tools like [Browsersync](https://www.browsersync.io/). Then I use compressor at the end, to compress the final assets. But I think It would be great if @chrisdev can help us to include those options in another PR, that could be very useful when not using these tools in the front side. What do you think?

andresgz:
BTW, just added the missing options in a last commit and also merged the last changes from master. It seems working fine

andresgz:
Hi guys,
I think this is ready to merge. Do you want me to add something extra?

pydanny:
@andresgz I'll be testing your pull request today both locally and in production and will give feedback and we'll hopefully get this merged in ASAP. 😄

pydanny:
I'm going to merge this in so users can try it. Considering that this touches static media and involves configuration, there are going to be a few issues, but nothing that shouldn't be easy to walk people through.

Thanks so much @andresgz!

andresgz:
Thanks @pydanny 
I will be pending in case there is any issue.

Issue #556, opened 2016-05-24
git pre/post commit hooks


jeffshek:
Thoughts on a pull request on a standard out of the box git pre-commit/post-commit hooks? I've used ones where we catch for any pdb in production / run pyflakes on code prior to commit, etc.

I'm looking at something similar below for inspiration
https://github.com/pre-commit/pre-commit-hooks

pydanny:
Interesting idea!

:bell: @burhan @theskumar @luzfcb @audreyr @jayfk

theskumar:
It's going to be a lovely edition.  It would be nice have some documentation on how to disable it temporarily.

djch:
@theskumar From my experience you can just do add the `--no-verify` argument to your `git commit`

jayfk:
Would love to have that as an optional thing :)

luzfcb:
+1 to to have that as an optional thing.
flake8 pre-commit hook and isort pre-commit hook[1] with `--check-only` and/or `--diff`

[1] https://github.com/FalconSocial/pre-commit-python-sorter

pydanny:
I think the consensus is that we would really like this feature. Please submit @jeffshek!

audreyr:
Or if anyone wants to work on adding some of these during the PyCon sprints, go for it. 👍

jeffshek:
@pydanny Sorry! Got a little caught up, but I'll make a PR for this tonight!

PR #570, opened 2016-06-02
Add pre-commit hooks

jeffshek:
Referencing https://github.com/pydanny/cookiecutter-django/issues/556
```
cookiecutter git@github.com:jeffshek/cookiecutter-django.git --checkout precommit_hooks
```
cd into the folder
``` bash
git init
git add .
git commit -m "first commit"
# install requirements in virtualenv / preference
pre-commit install
touch batman.py
git add .
git commit -am "testaroo"
```

Also wanted to mention, I tried out using subprocess with ['pre-commit', 'install'] arguments, but that requires to create the initial .git folder (where hooks are stored by default). It's not the end-all and git init would still work, but the debate was

1. Would users expect a .git folder to be automatically created prior to git init? No obvious downsides, just a little bit of surprise.
2. If a user did choose yes on precommit-hooks, then yes they should expect the folder .git to be created. Would (or could ??) he/she want to have the "initial commit" pass commit hooks?

This decision struck me as a core contributors decision and not mine to make 😰


pydanny:
In my opinion, we won't want to create the .git folder, as that puts too much of the burden of git maintenance on the core committers. But I'll 🔔 this on the other core contributors and get their thoughts @jayfk @burhan @luzfcb @theskumar @audreyr

pydanny:
When I try this, I get the following error:

``` bash
$ pre-commit install
$ bash: pre-commit: command not found
```

pydanny:
Okay, discovered you need to `pip install pre-commit`

pydanny:
D'oh. Maybe if I installed your requirements that would have worked.

pydanny:
This is pretty interesting, but there are a few issues:

1. Flake8 doesn't seem to be working properly. I've test files that  violate the heck out of PEP8 and yet aren't being flagged except for imports that aren't being used
2. Django really prefers if test files are prefixed with `test_`, not suffixed with `_test.py`. Can this be reversed.
3. Documentation! This should be mentioned in the README and in the formal docs.

As for the git stuff, I've decided that the control of git needs to be in the hands of the developer, not Cookiecutter Django. So you don't have to worry too much about git project instantiation except for providing it in the documentation for using pre-commit.

jeffshek:
1. I'll check it out again, I did try Flake8 and it did catch one of the errors, but it might be something else. Otherwise, might take out flake8 and leave it up to to the developer to choose the plugin of choice? I originally chose pyflakes, but that's being depreciated.
2. I did throw an extra param to consider for Django in there, so that it's not playing nice with tests is surprising/sad. Going to give it another try now.
3. Will do.

I agree about the .git folder instantiation! That would be a bit too surprising.

luzfcb:
I discovered that it is possible to overwrite the test message, adding a key `name`
```yml
    -   id: name-tests-test
        name: Django Tests should start with test_ prefix
        files: tests/.+\.py$
        args: ['--django']  # Use args: ['--django'] to match test*.py instead.
```

perhaps it would be useful to provide messages a little more explicit, but I do not know if it would be a good practice with pre-commit.

I do not know if it would be good to include by default, any hook that makes modifications on the code, like `end-of-file-fixer`, `trailing-whitespace` hook

```yml
-   repo: git://github.com/pre-commit/pre-commit-hooks
    sha: "6b1aa19425b35f8c34f43ada8f1045b028dc19a5"
    hooks:
    # keep ids alphabetical! see http://pre-commit.com/hooks.html
    -   id: debug-statements
        name: Check if not contains debug statements (pdb, ipdb, pudb, q)
        files: \.py$
    -   id: end-of-file-fixer
        files: \.(py|sh)$
    -   id: flake8
        files: \.(py|sh)$
    -   id: name-tests-test
        name: Django Tests should start with test_ prefix
        files: tests/.+\.py$
        args: ['--django']  # Use args: ['--django'] to match test*.py instead.
    -   id: trailing-whitespace
        files: \.(py|sh|yaml|txt)$
-   repo: git://github.com/FalconSocial/pre-commit-python-sorter
    sha: 1.0.4
    hooks:
    -   id: python-import-sorter
        args: ['--check-only']
        name: Test import order with isort

```

jeffshek:
@pydanny If you have an example of where it's failing on Pyflakes, that might be useful for me, otherwise, all comments addressed and recently pulled from master. Thanks!

luzfcb:
@pydanny another minor issue: pre-commit-hooks depends of pyyaml : https://github.com/pre-commit/pre-commit-hooks/blob/master/setup.py#L33

by consequence, he may suffer from the same problems that led to want to remove support for yml of cookiecutter

jayfk:
@luzfcb @pydanny pyyaml all over the place 😞

pydanny:
pyyaml? 😭 

Also, this error:

``` bash
git commit -am "testaroo"
An error has occurred: InvalidConfigError: File /Users/danielgreenfeld/code/posters/project_name/.pre-commit-config.yaml does not exist
Check the log at ~/.pre-commit/pre-commit.log
```

jeffshek:
@pydanny 
*scratches head* how does that ... NOT get copied over when someone makes a template the first time?

@luzfcb @jayfk 
😢 just looked up https://github.com/audreyr/cookiecutter/issues/743

Taking a step back, my feelings aren't hurt if we decide the dependency of "pre-commit" and thus pyyaml unfortunately hampers progress ...  there are other ways to include default pre-commit hooks (aka providing a configuration that can be copied/sym-ed) or decide maybe not at all too!

audreyr:
I tried this out and really, really like it a lot, so it's frustrating to not be able to include it because of the pyyaml issues.

audreyr:
Experimenting some more, when I tried removing `.pre-commit-config.yaml` and doing `pip uninstall pre-commit`, I got this error:

```
$ git commit -am "Some commit"
`pre-commit` not found.  Did you forget to activate your virtualenv?
```

Looks like `pre-commit` leaves some cruft around that breaks your git repo even when you uninstall it. I'm not a fan of tools that are easy to install but hard to uninstall.

So I'm 👎  on merging this into core Cookiecutter Django. But this was an awesome effort and exploration @jeffshek, so thank you for this.

pydanny:
Hey @jeffshek, really sorry, but the problems we've encountered trying to use pre-commit have been troubling. For me its been buggy, for others the intrusive nature of the library is distressing. That sucks because I really like the idea of pre-commit, but the implementation just doesn't seem to work for us.

However, I really appreciate the idea of this commit, and don't want you to give up contributing because it didn't get through. This kind of exploration is awesome, and needs to be done. I don't want to leave you with just a failed pull request on this project, so I really encourage you to submit more efforts in the future. 😄

jeffshek:
No problem :) I had high expectations for pre-commit too, the modular aspect of it looked great (better than the old-school pre-commit hooks regex search style), just quite didn't match expectations.

PR #587, opened 2016-06-04
Include Licensing options from cookiecutter.json in LICENSE template

### LINKED ISSUE #581: Dynamically generate license file
This should follow the same pattern as https://github.com/audreyr/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/LICENSE. Dependent on #578.

xpostudio4:
Fixes #581


audreyr:
Looks great @xpostudio4!

PR #600, opened 2016-06-07
automatically update INTERNAL_IPS

sladinji:
This trick is nice to automatically have debug toolbar when working with Docker.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -9,6 +9,8 @@ Local settings
 """
 
 from .common import *  # noqa
+import socket
+
 
 # DEBUG
 # ------------------------------------------------------------------------------
@@ -49,7 +51,6 @@ INSTALLED_APPS += ('debug_toolbar', )
 
 INTERNAL_IPS = ('127.0.0.1', '10.0.2.2',)
 # tricks to have debug toolbar when developing with docker
-import socket
 ip = socket.gethostbyname(socket.gethostname())
 INTERNAL_IPS += [ip, ip[:-1]+"1"]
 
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -48,6 +48,10 @@
 INSTALLED_APPS += ('debug_toolbar', )
 
 INTERNAL_IPS = ('127.0.0.1', '10.0.2.2',)
+# tricks to have debug toolbar when developing with docker
+import socket
+ip = socket.gethostbyname(socket.gethostname())
+INTERNAL_IPS += [ip, ip[:-1]+"1"]
```
theskumar: This looks good. I didn't get why are you using `ip[:-1]+"1"`. 

I think it would be nice, if you put the import statement on the top. My 2 cents!

sladinji:
ip = socket.gethostbyname(socket.gethostname())
=> ip value is the container IP and we want to host IP, which is usually the same but with a 1 at the end :
 INTERNAL_IPS += [ip, ip[:-1]+"1"]

sladinji:
Well, if finally took the time to read what I wrote and the last commit should be OK...

INTERNAL_IPS += [ip[:-1]+"1"]

We append docker host ip at the end of INTERNAL_IPS \o/

audreyr:
Thanks for working on this @sladinji! Looks like the error is due to the generated `local.py` file not passing flake8 linting. If you could add a fix for that to your pull request, that would be great. Let us know if you need any help with that.

sladinji:
@audreyr my pleasure, this project is really nice. 
I actually start a new project and I integrate django-oscar on the top of this cookiecutter. Basically, I apply add oscar config (http://django-oscar.readthedocs.io/en/latest/internals/getting_started.html) to current config, and I adapt templates to inherit from oscar templates. What's the best way to put this on github ? Should I create another fork named cookiecutter-django-oscar ?

pydanny:
How well does this work if you are not in a Docker container?

sladinji:
Well on my ubuntu, ip it's equal to "127.0.1.1", so it's fine. But it could be nice to check if we're in a linux container, you're right. It can be done by checking an environment variable set in Docker env file. I'll do that tomorrow if you want :)

mjsisley:
I would recommend adding the cookiecutter jinja if use_docker == y template tag

sladinji:
@mjsisley I used an env var instead of a template tag. This way, IP list is updated at runtime and you can switch between docker run and local run without worries.

Issue #466, opened 2016-01-11
Add React setup and configuration


pydanny:
I've hacked `React` into a project or two, but **I'll be the first to admit I'm not React setup expert**. I would love if someone who knows and understands the `React` ecosystem submitted a pull request that included the necessary elements in our front end tooling for building projects using React. 

Requirements:

1. This would have to be an option definined in `cookiecutter.json`. I would like to default to React and make Angular an option.
2. Documentation: How to on glue React into a template or two. I want more than just one, because per https://github.com/petehunt/react-howto, the first step to getting good is to replace individual components in React, not build a single-page app from scratch.

burhan:
+1 to this.

hanneshapke:
+1

onelson:
Do you have feelings about the asset pipeline?
Normally, I don't tie into `staticfiles` and treat my client source as a separate project (driven by grunt, gulp, webpack, et al).

pydanny:
`cookiecutter-django` currently uses Gulp for front end asset management, and it's bundled in. We're not adverse to switching to something else, but will want to keep the bundling in.

onelson:
I'm wondering if it'll even be possible to come up with a de facto implementation for this out of the gate. My projects, again, split client and backend entirely. As such, I handle identity and authorization via something similar to JWT (provided by django rest framework). This also relies on CORS being configured.

DRF has competitors, so you'll potentially want flavors for different backends. Additionally, there's a spectrum of ways to go about SPAs with Django, my preference being one of the extremes.

That all said, it's easier to decide what you like or don't like about a thing when you have the thing already in front of you. Maybe we throw something against the wall and see what sticks?

pydanny:
I would prefer to have a really strong DRF setup for React on `cookiecutter-django` than provide alternative options for REST support. Less complexity, and keeps this already **large** project from growing more.

Copser:
+1

dkdndes:
+1

Eraldo:
+1

\o/,
Eraldo

-- 
/============================================
| *E r a l d o    E n e r g y*
| *Coach*: Personal Development
| *Dancer*: Salsa, Bachata, Kizomba & Co.
| *Developer*: Python/Django
| *Legend*: coLegend.org <http://www.colegend.org>
|--------------------------------------------
| M: +49 176 97 99 44 80 <004917697994480>
| E: eraldo@eraldo.org
| W: www.eraldo.org
[...]

PR #610, opened 2016-06-11
WIP: Add webpack as an option

### LINKED ISSUE #538: Add Webpack as an Option
I've been interested in it as a way to maintain resources. And this is an AWESOME approach: https://github.com/hzdg/cookiecutter-webpack

Integrating it into Cookiecutter Django would be easy: https://github.com/hzdg/cookiecutter-webpack#integrating-into-existing-projects

goldhand:
This adds `Webpack` as an option in `js_task_runner`.

Still a work in progress but aims to resolve #538 

Works now but will need some work to be accepted. I just wanted to present this and collect feedback.

Unlike the grunt and gulp task runner options, this will clone a webpack config from [cookiecutter-webpack@pydanny-django](https://github.com/hzdg/cookiecutter-webpack/tree/pydanny-django)

* The `pydanny-django` branch of `cookiecutter-webpack` is dedicated to supporting this project and any changes that need to happen for `cookiecutter-webpack` to integrate well into here can be added in that branch (if there's a discrepancy with `master`).
* A package.json file will be generated that can be used to run the python and webpack hot reload server
* The webpack configuration files will be inserted into `/config/` and follow a similar pattern to the django settings. (`base`, `local` and `production`).
* Currently, this will also make react, redux and karma a part of the project (probably need to make those optional. Can consider make them options in the `cookiecutter-webpack` repo. This will also solve  #466 ).
* ~~There's a `less` loader in the webpack config that needs to be removed along with the less file in the demo.~~
* The app is currently injected into a `<div id="main"></div>` node in the base template but that can be changed if there's a better idea for how it should be integrated with django templates.
* Currently puts all webpack related static files into the dir `{{ project_slug }}/static/{{ project_slug }}`
* This webpack build doesn't go very well with the current static file dirs. Need to figure out a better structure.

To test this out, just select `3` (`Webpack`) as the option for `js_task_runner` and cd into the new project.

run `pip install -r requirements/local.txt && npm install`

to start the server with hot reloading enabled and a demo react / redux counter app run `npm start` and open up `localhost:8000`.

To run karma tests on static files run `npm test`.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -122,6 +122,7 @@ Listed in alphabetical order.
   Travis McNeill           `@Travistock`_               @tavistock_esq
   Vitaly Babiy
   Vivian Guillen           `@viviangb`_
+  Will Farley              `@goldhand`_                 @g01dhand
   Yaroslav Halchenko
 ========================== ============================ ==============
 
@@ -156,6 +157,7 @@ Listed in alphabetical order.
 .. _@eyadsibai: https://github.com/eyadsibai
 .. _@garry-cairns: https://github.com/garry-cairns
 .. _@garrypolley: https://github.com/garrypolley
+.. _@goldhand: https://github.com/goldhand
 .. _@hackebrot: https://github.com/hackebrot
 .. _@hairychris: https://github.com/hairychris
 .. _@hjwp: https://github.com/hjwp
--- a/README.rst
+++ b/README.rst
@@ -152,10 +152,12 @@ For development, see the following for local development:
 
 * `Developing locally`_
 * `Developing locally using docker`_
+* `Developing locally using webpack`_
 
 .. _options: http://cookiecutter-django.readthedocs.io/en/latest/project-generation-options.html
 .. _`Developing locally`: http://cookiecutter-django.readthedocs.io/en/latest/developing-locally.html
 .. _`Developing locally using docker`: http://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html
+.. _`Developing locally using webpack`: http://cookiecutter-django.readthedocs.io/en/latest/developing-locally-webpack.html
 
 Community
 -----------
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -18,7 +18,7 @@
     "use_docker": "y",
     "use_heroku": "n",
     "use_compressor": "n",
-    "js_task_runner": ["Gulp", "Grunt", "None"],
+    "js_task_runner": ["Gulp", "Grunt", "Webpack", "None"],
     "use_lets_encrypt": "n",
     "open_source_license": ["MIT", "BSD", "Apache Software License 2.0", "Not open source"]
 }
--- a/docs/developing-locally-webpack.rst
+++ b/docs/developing-locally-webpack.rst
@@ -1,0 +1,144 @@
+Developing Locally with Webpack
+===============================
+
+.. index:: Webpack, React, Redux, Karma, HMR
+
+The steps below will get you up and running with a super speedy local development environment with Webpack, React, Redux and Hot module Replacement.
+
+Before you begin, make sure you've created a python virtual environment, installed local python dependencies and created a postres database::
+
+    $ virtualenv env
+    $ source env/bin/activate
+    $ pip install -r requirements/local.txt
+    $ createdb [project_slug]
+    $ python manage.py migrate
+    $ python manage.py runserver
+
+These prerequisite steps are detailed more in the `developing locally section`_. If you've already done them, continue to installing javascript dependencies.
+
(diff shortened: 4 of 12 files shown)


pydanny:
Wow, this represents a lot of work (especially the branch of `cookiecutter-webpack`. I'm quite intrigued.

For this to be accepted we would have to:

1. Expand the documentation to better support Webpack. 
2. Figure out a way to include some comprehensive tests (:bell: @hackebrot and pytest-cookies). 
3. Come to some kind of understanding between Cookiecutter Django and Cookiecutter-webpack to ensure continued support and integration between projects. In short, I am concerned about the maintenance burden on you (@goldhand and @hzdg). May I assume that this is part of the toolchain used by @hzdg?)

goldhand:
Thanks for the feedback @pydanny 

1. That make sense, I'll see what I can do from both repositories.
2. I've been able to add [tests for linting](https://github.com/hzdg/cookiecutter-webpack/blob/master/tests/test_cookiecutter_generation.py#L56-L67) but it would be nice to run the project's karma tests. I'm not sure how to do that without adding a bunch of npm dependencies in the cookiecutter-django project. Maybe it would be better to isolate these tests in cookiecutter-webpack because they will mostly be javascript.
3. We do use cookiecutter-webpack it in our toolchain along with a variation of cookiecutter-django that uses EB for deployment. cookiecutter-webpack is consistently updated with the latest webpack 2 and react modules, for now at least. It may make sense to fork the project into one of your organizations and take in pull requests there. 

in addition:

4. Is the counter demo ok or should that be removed. It was added to give a baseline for structuring redux / react projects.
5. Should react / redux be options or can the be included by default?
6. karma, mocha and expect for testing, should they be included by default or optional?

goldhand:
I was just thinking, regarding the __3__ and the continued support of cookiecutter webpack.  Our current hzdg/cookiecutter-django has options for elastic beanstalk and django-cms, if those options were available with pydanny/cookiecutter-django we could probably just scrap our current one and use pydanny/cookiecutter-django in our toolchain.

jayfk:
I know this is going to be a lot of (additional) work, but I'd love to see a real world integration with react/redux. The counter example is nice for demo purposes, but isn't helpful when it comes to Django integration.

Cookiecutter Django comes with a couple of User/Profile views by default. Maybe that's a starting point to port them using react/redux. We don't even need to pull in something like DRF to make these views accessible via JSON.  We have `django-braces` that comes with a JSON response mixin which should be enough to get started.

@goldhand Is the Cookiecutter Django fork public? Personally, I'm in favor for option 3. It probably allows a tighter integration.

pydanny:
@goldhand In regards to merging in components of `hzdg/cookiecutter-django` into `pydanny/cookiecutter-django` the Elastic Beanstalk portion is something that could probably be added, especially if @hzdg has a reason for continued work on it.

However, the django-cms portion is tricky. We're agnostic about CMS decisions, which makes a lot of people happy. For what its worth, I know of at least one agency that uses vanilla Cookiecutter Django, but incorporates an in-house script that uses Cookiecutter's `work_in` function to easily allow them to make enough changes so that django-cms integration is easy. If you like, I can probably get a snippet of code out of them, or even write it myself.

goldhand:
@jayfk I like the idea of working with the existing user views, it might not be too much additional work especially if I can find something that fits this criteria in a django/react+redux project I'm working on.  Currently, @hzdg's cookiecutter-django-hzskeleton is not public though. It's similar to this one only it has additional options for django-cms and eb. I can look into making it public.

@pydanny I'm curious how the `work_in` function helped them do that, if we can use it to add a django cms configuration that would be awesome. All our projects are going on to eb now, so I don't see maintaining that feature an issue.

pydanny:
@goldhand `work_in` is a context manager that lets you easily change directories and then back out. This is great for writing post-Cookiecutter scripts. Which is what a certain large corporation that just got bought by Microsoft does for all their new Java and Python projects (albeit with their own internal Cookiecutter templates). It works at the most basic level like so:

``` python
from cookiecutter.utils import work_in

# Add Django CMS as a post-generate requirement
with work_in(project_name):
    with open("requirements/base.txt", "a") as f:
        f.write("\ndjango-cms==3.x")
```

goldhand:
Awesome, I think that will work fine for us! I mean... if it works for a big company that microsoft recently bought :laughing:

pydanny:
@goldhand If you can resolve the merge conflicts I'm willing to merge this in. 👍 

For reference, there are going to be some hiccups and post-merge cleanups. That's okay, it's expected on something that brings in a whole new set of features into the project.

goldhand:
Alright! :+1: I'm going to work on the merge conflicts and docs this weekend and hopefully have it ready by Monday

goldhand:
@pydanny webpack documented and conflicts resolved

pydanny:
Thank you so much @goldhand!

PR #613, opened 2016-06-13
Improve docs/developing-locally.rst

kappataumu:


audreyr:
Fantastic work @kappataumu.

PR #614, opened 2016-06-13
Added django-robots integration

msamoylov:
Google seems to require robots.txt for new sites, otherwise it's not going to crawl them returning "Google couldn't crawl your site because we were unable to access your site's robots.txt file" in Google Search Console. Please refer to https://support.google.com/webmasters/answer/2409682 for the details.

This pull requests contains a working and tested integration of https://github.com/jazzband/django-robots/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,7 @@ Optional Integrations
 .. _docker-compose: https://www.github.com/docker/compose
 .. _Opbeat: https://opbeat.com/
 .. _PythonAnywhere: https://www.pythonanywhere.com/
+.. _django-robots: https://github.com/jazzband/django-robots/
 
 
 Constraints
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -11,6 +11,7 @@
     "use_celery": "n",
     "use_mailhog": "n",
     "use_sentry_for_error_reporting": "y",
+    "use_robots": "y",
     "use_opbeat": "n",
     "use_pycharm": "n",
     "windows": "n",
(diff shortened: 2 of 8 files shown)


msamoylov:
I have to admit that I don't know how to fix that failing test.

luzfcb:
Jinja2 has a different syntax to strip whitespace
http://jinja.pocoo.org/docs/dev/templates/#whitespace-control

It will probably be something like: 
```django
{% if cookiecutter.use_robots == 'y' -%}

{%- endif %}
```

instead of

```django
{% if cookiecutter.use_robots == 'y' %}

{% endif %}
```

ps: the first time I had seen it, I thought it was a syntax error :-P

msamoylov:
@luzfcb thanks man! That wasn't obvious for sure.

pydanny:
Questions:

1. Why does this need to be in Cookiecutter Django? Shouldn't users add this of their own accord?
2. If django-robots is used, does it come with intelligent defaults?

msamoylov:
1. Well, it's just nice to have and it seems to be better than a static file. I've been using it for 3-4 years and include  it in every public project.
2. It allows everything by default, so your website is accessible.

```
User-agent: *
Allow: /
```

pydanny:
@msamoylov Thank you so much for your pull request. It is very much appreciated.

However, I don't believe this should be integral to Cookiecutter Django. Considering the simplicity of the integration, have you thought about writing a post-generate script to add this to your projects programmatically?

PR #638, opened 2016-06-21
makes the database backups compressed. restores compressed backups

jangeador:
The database controls in the postgres docker image are very nice, however the uncompressed output grows really fast. This pull request modifies the scripts in the postgres docker container to output gzip files and also to restore from them. In my case the result are backup files which are many times smaller. This is an advantage especially if you are storing your backups on a paid cloud.


jayfk:
Thanks!

PR #639, opened 2016-06-21
Improve `install_os_dependencies.sh`

ddiazpinto:
This PR improve some things of the `install_os_dependencies.sh` script. In particular:

- Checks if a requirements file exist for the current distribution (which means that is supported) instead of be implicitly defined inside the script.
- Uses HEREDOC for grouping text and improve readability, specially for the usage messages. Regrettably, HEREDOC can only be indented with tabs.
- Redirects all the error messages to the error standard output. I suppose that the above redirection `2>&1` was an error and it should actually be `>&2`. If there is a reason to redirect all the errors to the standard output, please tell me and I'll fix it.
- Fixed some minor typos

### DIFF
--- a/{{cookiecutter.project_slug}}/utility/install_os_dependencies.sh
+++ b/{{cookiecutter.project_slug}}/utility/install_os_dependencies.sh
@@ -2,39 +2,48 @@
 
 WORK_DIR="$(dirname "$0")"
 DISTRO_NAME=$(lsb_release -sc)
+OS_REQUIREMENTS_FILENAME="requirements-$DISTRO_NAME.apt"
 
-OS_REQUIREMENTS_FILENAME="$WORK_DIR/requirements-$DISTRO_NAME.apt"
+cd $WORK_DIR
 
-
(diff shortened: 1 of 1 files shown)


luzfcb:
I'll probably have time to review this , next weekend

{{cookiecutter.project_slug}}/utility/install_os_dependencies.sh
```diff
@@ -2,39 +2,48 @@
...
+OS_REQUIREMENTS_FILENAME="requirements-$DISTRO_NAME.apt"
+
+cd $WORK_DIR
+
+# Check if a requirements file exist for the current distribution.
+if [ ! -r "$OS_REQUIREMENTS_FILENAME" ]; then
+    cat <<-EOF >&2
+		There is no requirements file for your distribution.
+		You can see one of the files listed below to help search the equivalent package in your system:
+		`find ./ -name "requirements-*.apt" -printf "  - %f\n"`
```
luzfcb: use, `$( foo_program )` insteadof ``` `foo_program` ```:


```
$(find ./ -name "requirements-*.apt" -printf "  - %f\n")
```
insteadof 
```
`find ./ -name "requirements-*.apt" -printf "  - %f\n"`
```

luzfcb:
@ddiazpinto thanks again. :smile:

PR #657, opened 2016-06-28
Update project generation options docs

kappataumu:


kappataumu:
This should be closed since the project generation options have been updated :)

PR #681, opened 2016-07-21
[requires.io] dependency update on master branch

pydanny:


audreyr:
Closing since we're using pyup.io. requires.io for this repo has also been deactivated.

PR #684, opened 2016-07-23
Update deployment-on-heroku.rst

bionikspoon:
On linux this `openssl` command usually creates inappropriate strings and the 'ticks' don't guard against it.  My way, works everytime. Noob friendly.


pydanny:
Many thanks @bionikspoon!

PR #663, opened 2016-07-06
Fix config files readability noise

aleprovencio:
Those are mostly little bugs introduced from other features, plus linter issues. I've also put a separator line to improve readability.

PR #701, opened 2016-07-31
Hide django-compressor comment section

gopar:
If you choose to not use `django-compressor`, it will still leave a comment section for that library.
This should hide it if you don't choose that option.


aleprovencio:
Hello @gopar, I've already covered this here https://github.com/pydanny/cookiecutter-django/pull/663

gopar:
@aleprovencio Ah thanks for telling! I'll close this in favor of yours then.

Issue #602, opened 2016-06-08
Wait for postgres container to start


jayfk:
Django's `runserver` relies on a working database connection during startup. If this is not the case `runserver` fails entirely and the container has to be restarted.

To circumvent that, the `runserver` call should be wrapped in a function that checks if the postgres database is up and running.

This only affects the development environment. Production is fine.

rougeth:
There's an example on Docker Compose docs of a container waiting for postgres to start, that can be useful: https://docs.docker.com/compose/startup-order/

anthonysutardja:
This sounds like something that could be solved with modifying the command in the docker-compose file. 

https://github.com/vishnubob/wait-for-it

Maybe placing this file (`wait-for-it.sh`) in the `utilities` directory could be useful? The script waits for the host and port to spin up (with a timeout) before exiting. The command in docker-compose would then be changed to the following:

```
   command: bash -c "./utilities/wait-for-it.sh postgres:5432 && python /app/manage.py runserver_plus 0.0.0.0:8000"
```

yashinshekh:
docker-compose down --volumes

PR #736, opened 2016-08-16
Wait for the postgres container to start

jayfk:
This is a backport from https://cookiecutter-saas.com

It adds a small python script to `entrypoint.sh` that checks if the postgres container is up and running. If postgres is not up yet, it waits for a second and tries again.

Related issue: https://github.com/pydanny/cookiecutter-django/issues/602

We had a couple of issues with the entrypoint on windows in the past. Could someone confirm this works on windows before merging this in?


pydanny:
I like this one!

PR #749, opened 2016-08-21
Change to use_python3 [y] by default

### LINKED ISSUE #747: Change use_python2 [N] to use_python3 [Y]
I prefer to use Python 3. The current `cookiecutter.json` module presents the older version, but with a negative default. I would rather list Python 3 with a positive default.

ameistad:
Fixes #747


jayfk:
Thanks! 👍

PR #754, opened 2016-08-24
Use app registry instead of INSTALLED_APPS to discover celery tasks

### LINKED ISSUE #717: apps.UsersConfig prevents celery task autodiscovery
Celery tasks that are created in `users/tasks.py` are not picked up automatically.

Probably because the `users` app is added to `INSTALLED_APPS` by pointing to the `UsersConfig` class instead of loading the module.

dhepper:
Fixes #717


jayfk:
Great, thanks!

PR #761, opened 2016-08-29
Replace unicode-slugify with awesome-slugify

### LINKED ISSUE #758: Slugify function
Slugify function installed not permit to change space remplacement character by other than '-'

stanislavshabalin:
awesome-slugify: https://github.com/dimka665/awesome-slugify
Per @pydanny's comment.

Resolves #758 

That was _too_ easy, is there anything else I'm missing?
I've searched `slugify`, there's also `slugify.slugify` function in the settings — still valid with the new package.

Maybe add something in the docs, but where? I'm not familiar with the project yet, but docs look like they need some loving :–)

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -40,7 +40,7 @@ psycopg2==2.6.2
 {%- endif %}
 
 # Unicode slugification
-unicode-slugify==0.1.3
+awesome-slugify==1.6.5
 django-autoslug==1.9.3
 
 # Time zones support


theskumar:
This looks good to me. I would defer to @pydanny  for final merge. 

As per my knowledge the library is not used anywhere with the extra code generated, is here just as a recommended slugifier. Not sure, how much of value addition it is, but I might consider removing it altogether. @pydanny thoughts?

pydanny:
I'm delighted to "see" you again @theskumar. 😄 

I use internationalized slugification a lot, so keeping this in is very useful for me. This is a good merge candidate.

PR #767, opened 2016-08-31
Pep8 imports

aleprovencio:


jayfk:
What's the reason to do this?

theskumar:
@jayfk this seems to breakup of the work done here: https://github.com/pydanny/cookiecutter-django/pull/663#issuecomment-243847420. it's just pepification of imports (relative imports come at the end). I don't have issue merging or closing either. 

Thinking deep, iSort[1] with it's configuration[2] in setup.cfg/.editorconfig is what use so we don't face these kind of issues and modifications, so everyone has a standard guide and tool to sort & write imports. 

[1] https://github.com/timothycrosley/isort
[2] https://github.com/Fueled/django-init/blob/master/%7B%7Bcookiecutter.github_repository%7D%7D/.editorconfig#L18-L23

PR #712, opened 2016-08-04
#667 Newline Madness

nelsonlho:
Got rid of base.txt unnecessary newlines

PR #775, opened 2016-09-04
Correct header markup, reduce paragraph spacing

bittner:
This PR ensures

- that the header markup is always valid. (When the project title is longer than the current hard-coded length of the h1 markup then the rendering would break.)
- In addition, spacing between paragraphs is reduced to make the README file more consistent.

## Side note on line spacing

Every template block introduces an additional newline. This is only a default setting of Jinja2, though, and can be avoided by setting Jinja2's `trim_blocks` environment option to `True`. See the docs for details:

- [Whitespace Control](http://jinja.pocoo.org/docs/dev/templates/#whitespace-control)
- [jinja2.Environment](http://jinja.pocoo.org/docs/dev/api/#jinja2.Environment)

(Can that be set as a default in cookiecutter, or is there a reason for not doing that?)

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -2,7 +2,7 @@ Contributors
 ============
 
 Core Developers
-----------------
+---------------
 
 These contributors have commit flags for the repository,
 and are able to accept and merge pull requests.
@@ -28,7 +28,7 @@ Daniel are on the Cookiecutter core team.*
 .. _@jayfk: https://github.com/jayfk
 
 Other Contributors
--------------------
+------------------
 
 Listed in alphabetical order.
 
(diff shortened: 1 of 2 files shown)


jayfk:
👍🏼

bittner:
Thanks @jayfk!

Any comment on the  Jinja2 `trim_blocks` environment setting?

jayfk:
I'm not quite sure if it's possible to set that in cookiecutter. Either way, I think it's better to stay on the default option for now. Changing that has the potential to introduce a lot of bugs that are hard to debug since the templates were created with `trim_blocks==False` in mind.

aleprovencio:
@jayfk

aleprovencio:
@jayfk , maybe the bugs are already there with this option, see https://github.com/pydanny/cookiecutter-django/pull/712

Issue #706, opened 2016-08-01
Update docker docs


jayfk:
Docker for Mac/Windows is the new thing. We should update the docs accordingly.

ssteinerx:
I can help with this though the lack of good mapping between the OS X and Debian/Ubuntu makes it so you have to use weird paths with everything e.g. you can't use /var/... on OS X which is the 'normal' place for things on Deb/Ubuntu.

jayfk:
@ssteinerx `dev.yml` and `docker-compose.yml` both use named volumes, so this shouldn't be a problem.

ssteinerx:
Yes, but you have to either use non-standard locations or different files for OS X and e.g. Debian.

PR #786, opened 2016-09-11
Update to use Docker for Mac/Windows

ameistad:
Removed sections about using Docker Toolbox with virtualbox and added links to Docker for Mac/Windows.
Docker for Mac/Windows uses native virtual machines instead of virtualbox and introduces many improvements.
See: https://blog.docker.com/2016/03/docker-for-mac-windows-beta/

Resolves: #706

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -13,35 +13,12 @@ You'll need at least Docker 1.10.
 
 If you don't already have it installed, follow the instructions for your OS:
 
- - On Mac OS X/Windows, you'll need `Docker Toolbox`_
+ - On Mac OS X, you'll need `Docker for Mac`_
+ - On Windows, you'll need `Docker for Windows`_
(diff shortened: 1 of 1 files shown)


jayfk:
👍

ssteinerx:
"Docker for Mac/Windows uses native virtual machines instead of virtualbox and introduces many improvements."

It also requires moving a lot of stuff around since the "usual places" on Linux are restricted access on Mac.  I think this should be *added* to the doc and the stuff about the original setup retained.

I work on a Mac in Docker and use VirtualBox to avoid the artificial constructs required to use Docker Mac.

jayfk:
What "usual places" are you talking about? You can add pretty much anything via the GUI to be mounted on VM start.

ssteinerx:
I'm talking mostly about shared volumes.  OS X doesn't allow /var/www, for example -- it's owned by root:wheel, so you have to put things in non-standard places.

Issue #792, opened 2016-09-15
add LocaleMiddleware


willemarcel:
The settings comes with `USE_I18N` and `USE_L10N` set as `True`, but we need to add `'django.middleware.locale.LocaleMiddleware'` to make the localization and internationalization work.

Is there any reason why the `LocaleMiddleware` isn't in the `MIDDLEWARE_CLASSES` by default?

jayfk:
Thanks for the heads up, it's coming with 1.10: #793.

Issue #703, opened 2016-08-01
Update to Django 1.10


luzfcb:
Checklist:

- [x] Before update to Django 1.10, create a git tag to current version of cookiecuter-django
- [ ] Check if all dependencies officialy work with Django 1.10( if not, wait)
- [x] Check that the tests pass on Django 1.10
- [ ] Check that the tests pass on Django 1.10 with zero deprecation warnings messages: https://docs.djangoproject.com/en/1.10/howto/upgrade-version/#resolving-deprecation-warnings
- [ ] Update to Django 1.10 and update other dependencies and:
    - [x] Check if all possible generated project layout still works on deploy on Heroku.
    - [ ] Check if all possible generated project layout still works on deploy on Docker.
    - [ ] Check PyCharm 2016.2 compatibility, with and widout docker
    - [x] Update the README.rst
    - [x] Update [version on setup.py](https://github.com/pydanny/cookiecutter-django/blob/master/setup.py#L13)
    - [x] Update changelog.
    - [ ] after update, create a git tag

related: https://github.com/pydanny/cookiecutter-django/pull/702

ssteinerx:
Generated layout works on Docker other than #704 opened 18 hours ago by ssteinerx

ssteinerx:
Pycharm container also crashes.

audreyr:
Verified that the tests pass on 1.10

audreyr:
Dependencies that are Django version-dependent:

- [ ] django-environ
- [ ] whitenoise
- [ ] django-braces
- [ ] django-crispy-forms
- [ ] django-floppyforms
- [ ] django-model-utils
- [ ] django-allauth
- [ ] django-autoslug
- [ ] django-redis
- [ ] django-coverage-plugin
- [ ] django-extensions
- [ ] django-test-plus
- [ ] factory_boy
- [ ] django-debug-toolbar
- [ ] pytest-django

Check off each of the above when it officially works on 1.10.

jayfk:
django-model-utils is not ready, see the latest change log in #709

jayfk:
django-extensions is not ready, see https://github.com/django-extensions/django-extensions/pull/934

jayfk:
django-test-plus is not ready, see https://github.com/revsys/django-test-plus/pull/45

jayfk:
django-debug-toolbar is not ready, see https://github.com/jazzband/django-debug-toolbar/pull/864

pydanny:
[...]

PR #793, opened 2016-09-15
WIP - update to django 1.10

luzfcb:
changes:
- [x] rename MIDDLEWARE_CLASSES to MIDDLEWARE
- [x] add default AUTH_PASSWORD_VALIDATORS generated by django 1.10 startproject
- [x] update `manage.py` to use same code of `manage.py` from django 1.10
- [x] <s>add LocaleMiddleware to fix issue #792</s> (It will be included in a separated pull-request)
- [x] update django version on `setup.py`
- [x] make tests pass by passing new password auth rules
- [x] update python dependencies
- [x] update to bootstrap 4 alfa 4, jquery and tether.js. (i generate the integrity hash using https://www.srihash.org/ . this was indicated by documentation https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
- [x] sync `sites` app migrations with django 1.10, and fix aditional migrations to `sites` and `user` app

- [ ] please, not merge, still requires checks everything that was said in https://github.com/pydanny/cookiecutter-django/issues/703


{{cookiecutter.project_slug}}/config/settings/common.py
```diff
@@ -53,9 +53,10 @@
 
 # MIDDLEWARE CONFIGURATION
 # ------------------------------------------------------------------------------
-MIDDLEWARE_CLASSES = (
+MIDDLEWARE = (
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.locale.LocaleMiddleware',
```
pydanny: I know we should include it, but is `LocaleMiddleware` a requirement for Django 1.10?

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -50,7 +50,7 @@
 
 # django-debug-toolbar
 # ------------------------------------------------------------------------------
-MIDDLEWARE_CLASSES += ('debug_toolbar.middleware.DebugToolbarMiddleware',)
+MIDDLEWARE += ('debug_toolbar.middleware.DebugToolbarMiddleware',)
```
pydanny: Sigh. I know this is necessary, but I'm not the biggest fan of the middleware changes. 😩

{{cookiecutter.project_slug}}/manage.py
```diff
@@ -5,6 +5,19 @@
 if __name__ == '__main__':
     os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'config.settings.local')
 
-    from django.core.management import execute_from_command_line
-
+    try:
```
pydanny: I love this whole section!

{{cookiecutter.project_slug}}/config/settings/common.py
luzfcb: ok, I'll will remove and send in a separate pull request

ssteinerx:
What is code coverage after this?  I know we had to kill the template coverage gizmo (though I think I figured out how to fix it).  

Do we have a test that asserts 100% code coverage on a newly minted project?

ssteinerx:
@luzfcb Could you please merge master into this branch or rebase?  A bunch of my change on this branch are due to my branch being is up to date with master and it makes it hard to see what is different that is not due to that.

I have a couple of small changes (updating .gitignore, for example) that should go into the 1.10 release, but a pull request now would contain all the master branch updates too which is not what we want.

Thanks!

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
```diff
@@ -27,8 +27,8 @@ def test_clean_username_false(self):
         # Instantiate the form with the same username as self.user
         form = MyUserCreationForm({
             'username': self.user.username,
-            'password1': '123456',
-            'password2': '123456',
+            'password1': 'notalamodespassword',
+            'password2': 'notalamodespassword',
```
jayfk: :shipit:

pydanny:
@ssteinerx This isn't standard practice for projects, but we're not worrying about getting to 100% test coverage in a Django version update pull request. Here's why:

Getting to the 100% test coverage number in Cookiecutter Django is tedious work, especially when jumping versions. What we do is accept it, then as soon as it's done I jump to working on the code coverage. By making code coverage a separate effort, it allows for focus on getting the version update to just work. Some notes:

* If tests fail, we don't accept the version update pull request. Ever!
* Usually the incoming coverage score for tests on a version update is 95-100%.

ssteinerx:
@pydanny -- I'm more concerned about testing our claim of "generating a Django project that has 100% test coverage out of the box."  IOW, not cookiecutter-django itself, but the project generated therefrom.

pydanny:
@luzfcb :shipit:

jayfk:
ship it

PR #795, opened 2016-09-15
Make test_docker.sh tests pass by passing new password auth rules

ssteinerx:
Just fixed the broken test under Django 1.10 we discussed on Glitter.  All I did was make the passwords more complicated.

Also, make_user()'s default values didn't pass the new password validation rules either so I  and was explicit about the cal to make_user() so it was more clear what was going to come back (and so it would pass validation).


luzfcb:
@ssteinerx thanks :smiley:

Issue #721, opened 2016-08-07
Jinja2 license issue prevents installation using all default settings.


tkovalsky:
when trying to start a cookie cutter project using all default settings, I receive a ./LICENSE error. 

| => cookiecutter https://github.com/pydanny/cookiecutter-django
You've cloned /Users/dev/.cookiecutters/cookiecutter-django before. Is it okay to delete and re-clone it? [yes]: yes
Cloning into 'cookiecutter-django'...
remote: Counting objects: 7253, done.
remote: Compressing objects: 100% (10/10), done.
remote: Total 7253 (delta 4), reused 0 (delta 0), pack-reused 7243
Receiving objects: 100% (7253/7253), 2.74 MiB | 818.00 KiB/s, done.
Resolving deltas: 100% (4677/4677), done.
Checking connectivity... done.
project_name [project_name]: trec
project_slug [trec]: 
author_name [Your Name]: Todd Kovalsky
email [Your email]: t.e.kovalsky@gmail.com
description [A short description of the project.]: a web page for trec solutions.
domain_name [example.com]: 
version [0.1.0]: 
timezone [UTC]: 
use_whitenoise [y]: 
use_celery [n]: 
use_mailhog [n]: 
use_sentry_for_error_reporting [y]: 
use_opbeat [n]: 
use_pycharm [n]: 
windows [n]: 
use_python2 [n]: 
use_docker [y]: 
use_heroku [n]: 
use_compressor [n]: 
Select js_task_runner:
1 - Gulp
2 - Grunt
3 - Webpack
4 - None
Choose from 1, 2, 3, 4 [1]: 
use_lets_encrypt [n]: 
Select open_source_license:
1 - MIT
2 - BSD
3 - GPLv3
4 - Apache Software License 2.0
5 - Not open source
Choose from 1, 2, 3, 4, 5 [1]: 
Traceback (most recent call last):
  File "/usr/local/bin/cookiecutter", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

Issue #528, opened 2016-04-20
cookiecutter fails on jinja2.exceptions.TemplateSyntaxError


patvdleer:
~~~
timezone [UTC]:     
Traceback (most recent call last):
  File "/usr/local/bin/cookiecutter", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/cookiecutter/cli.py", line 100, in main
    config_file=user_config
  File "/usr/local/lib/python3.4/dist-packages/cookiecutter/main.py", line 131, in cookiecutter
    context['cookiecutter'] = prompt_for_config(context, no_input)
  File "/usr/local/lib/python3.4/dist-packages/cookiecutter/prompt.py", line 126, in prompt_for_config
    val = render_variable(env, raw, cookiecutter_dict)
  File "/usr/local/lib/python3.4/dist-packages/cookiecutter/prompt.py", line 86, in render_variable
    template = env.from_string(raw)
  File "/usr/local/lib/python3.4/dist-packages/jinja2/environment.py", line 862, in from_string
    return cls.from_code(self, self.compile(source), globals, None)
  File "/usr/local/lib/python3.4/dist-packages/jinja2/environment.py", line 565, in compile
    self.handle_exception(exc_info, source_hint=source_hint)
  File "/usr/local/lib/python3.4/dist-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.4/dist-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "<unknown>", line 1, in template
  File "/usr/local/lib/python3.4/dist-packages/jinja2/environment.py", line 470, in _parse
    return Parser(self, source, name, encode_filename(filename)).parse()
  File "/usr/local/lib/python3.4/dist-packages/jinja2/parser.py", line 897, in parse
[...]

PR #820, opened 2016-09-26
Update readme to point to cookiecutter>=1.4

townie:
I struggled with a stale version of cookiecutter on my machine. That being said indicating that it requires a specific version or great would have helped reduce the pain of getting started.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,7 +90,7 @@ and then editing the results to include your name, email, and various configurat
 
 First, get Cookiecutter. Trust me, it's awesome::
 
-    $ pip install cookiecutter
+    $ pip install "cookiecutter>=1.4.0"
 
 Now run it against this repo::
 


ssteinerx:
You shouldn't be installing cookiecutter (or anything else, really) in your global Python and the instructions are pretty clear about starting with a new virtualenv.

From: https://cookiecutter-django.readthedocs.io/en/latest/developing-locally.html

> First make sure to create and activate a virtualenv.

It then  walks you through:

> Then install the requirements for your local development:
> `$ pip install -r requirements/local.txt`

I think that maybe the readme should be modified to point to the online docs and the "getting started" instructions should be removed from the readme itself as it just makes for two places to maintain the same information.

townie:
@ssteinerx Using virtualenvs is always best practice, I don't disagree with you that. This change is not referencing local development of this package/django but the actual use of it to generate an new django project.

That being said I have a general/miscellaneous/scratch virtualenv that I just stick all random packages that I want to try out(probably not the most ideal), having used cookiecutter for other projects in the past and the version of cookiecutter which I had pip install a while back was 1.3 which is incompatible with the template tool Jinja2 used to create the initial project. 

Therefore this doc change it ideally to reduce the chance that someone will make a bone head move and spend 30 minutes trying to get it to work on an outdated version of cookiecutter(like I did 😢)

Reference: https://github.com/pydanny/cookiecutter-django/issues/528#issuecomment-212650373 
https://github.com/pydanny/cookiecutter-django/issues/721#issuecomment-238084609

luzfcb:
thanks @townie

Issue #841, opened 2016-10-08
psycopg2 not installed with Windows 10 + Docker


mimischi:
The current setup does not generate a working project on Windows 10 with Docker for Windows. Basically the requirements/base.txt file checks whether Windows 10 is used or not. If it is used, then it will not include psycopg2 in the requirements (see [here](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/requirements/base.txt#L33-L39)):

```
{% if cookiecutter.windows == 'y' -%}
# On Windows, you must download/install psycopg2 manually
# from http://www.lfd.uci.edu/~gohlke/pythonlibs/#psycopg
{% else %}
# Python-PostgreSQL Database Adapter
psycopg2==2.6.2
{%- endif %}
```

This may work in case a user is not intending to use Docker at all. Nonetheless it should be installed inside the corresponding containers.

To fix this as of now, one has to enter the Django container (and PyCharm, if specified) with `docker exec -it <container-id> bash` and run `pip install psycopg2` manually. Afterwards `docker-compose -f dev.yml up` will boot up.

ssteinerx:
I think that the documentation should be updated to specifically say this if it doesn't, and some sort of error message should come out of the build process if it doesn't.

@mimischi Did you see a place in the docs where this should be mentioned and, if so, could you write what you would have expected to see there and submit a pull request?

Since this `%if` is in the requirements.txt generation, is there a way of getting a message out to the console during the build process to flag the required manual step?

mimischi:
The best place would probably be directly under the Prerequisites [here](https://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html). I'll make a pull-request.

I see one major problem with including another if in the base.txt file:

What if I'm using Windows and say `yes` to Docker, but actually want to develop in an local environment without it? This would yield an error, because we can't install `psycog2` on Windows using pip.
So the question is whether for this case there should be a new requirements file `docker-win.txt` with `psycog2` defined, which would be included inside of `compose/Dockerfile-dev`?

KrunchMuffin:
I don't think this is a docker + win10 issue as I am having the same issue w/o docker option. Tho I am a python/django n00b.

> Traceback (most recent call last):
  File "C:\Users\me\Anaconda3\lib\site-packages\django\db\backends\postgresql\base.py", line 20, in <module>
    import psycopg2 as Database
[...]

PR #845, opened 2016-10-11
Add documentation for Docker users on Windows

mimischi:
`psycopg2` is currently not installed when using Docker on an Windows environment. Before this gets fixed (see #841), we need some documentation around this issue.


jayfk:
Thanks!

PR #853, opened 2016-10-17
remove html5.js as source no longer exists

shireenrao:
I removed html5.js from base.html as the source no longer exists.


shireenrao:
I updated the source to point to https://cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv.min.js instead.

PR #863, opened 2016-10-24
read the env, so the env works

spoutnik16:
Spend a whole afternoon tryin to make it works after reading your book... that was the solution to the first problem...


pydanny:
Solution to what problem?

pydanny:
Looks like django-environ changed the API. Thanks for the catch.

jayfk:
Merging this breaks the docker setup, because the `.env` file is not copied into the image.

The `.env` file was intended to be used only by docker-compose to set up the environment during the build phase.

pydanny:
Oops. I forgot about that detail. I'll look more carefully at the API change and implement the fix.

PR #930, opened 2016-12-08
Updated settings to only include lists and '

schacki:
Substituted " with ' in files
Migrated tuples  to lists

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -32,7 +32,7 @@ SECRET_KEY = env('DJANGO_SECRET_KEY', default='CHANGEME!!!')
 
 EMAIL_PORT = 1025
 {% if cookiecutter.use_mailhog == 'y' and cookiecutter.use_docker == 'y' %}
-EMAIL_HOST = env("EMAIL_HOST", default='mailhog')
+EMAIL_HOST = env('EMAIL_HOST', default='mailhog')
 {% else %}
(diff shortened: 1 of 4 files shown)


{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -50,14 +50,14 @@
 
 # django-debug-toolbar
 # ------------------------------------------------------------------------------
-MIDDLEWARE += ('debug_toolbar.middleware.DebugToolbarMiddleware',)
-INSTALLED_APPS += ('debug_toolbar', )
+MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware']
```
pydanny: No trailing comma?

pydanny:
No trailing comma?

pydanny:
... and thanks for this pull request. 😄

Sgiath:
And what about other files? For example [here](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/wsgi.py#L28)

schacki:
Sorry for the late reply. Trailing commas are not used consistently, I am happy to make it consistent in settings files. What do you think? Is there any place in the docs, where to put this as a requirement for contributors? The same for the "?

schacki:
So what is the preference, with or without trailing comma?

webyneter:
@schacki, trailing comma is always a great, defensive choice, IMO.

PR #936, opened 2016-12-09
Update ADMINS from tuple to list.

Afrowave:
In Django 1.9, the ADMINS email details were changed from a tuple to a list
and the values are wrapped in single quote marks.


{{cookiecutter.project_slug}}/config/settings/common.py
```diff
@@ -89,9 +89,9 @@
 # MANAGER CONFIGURATION
 # ------------------------------------------------------------------------------
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#admins
-ADMINS = (
-    ("""{{cookiecutter.author_name}}""", '{{cookiecutter.email}}'),
```
luzfcb: maintain the triple quote marks, because if the `cookiecutter.author_name` is, for example, `O'Brien`, the code will break

Afrowave: The change  from the tuple to a list is okay?

Issue #944, opened 2016-12-15
env.example and .env


schacki:
Currently, I manually need to copy env.example to .env after each project creation. Would it make sense to add that step to the post_gen_project hook?

Furthermore, django_environ complains about the .env file not existing. This could be resolved by:
a) also copying env.example to config/settings/.env
or 
b) pointing to the .env file in the project root by saying: env.read_env(os.path.join(PROJECT_ROOT, '.env')
I would prefer option b.

TidyData:
I like @schacki 's idea and generated a pull request to implement it assuming maintainers agree.

TidyData:
I can't figure out how to deconflict it though, I don't understand because I literally just forked it from this repo....

webyneter:
No longer an issue since #1295.

PR #950, opened 2016-12-17
env.example and .env #944

TidyData:
saves others from the same mistake when recreating .env

My .env file was embedded in my repo and I cleaned out my repo thinking every file in it was version controlled, only to find a key file missing.
Then recreating the .env file, I accidentally left a python string ' at the end and didn't catch it until now.

issue: #944
@schacki, I immediately appreciated your idea.

Signed-off-by: TidyData <mckweb@outlook.com>


schacki:
what about generating the .env file as copy of the example.env in the hook?

kerstin:
@TidyData I think it conflicts because you are replacing an existing line. Line 19 is not empty in my `common.py` file:

```
env = environ.Env()
env.read_env()
```

PR #971, opened 2016-12-31
Use django user for Docker user

moas:
Use user django instead of root


webyneter:
Closing this in favour of #1144.

PR #983, opened 2017-01-06
 Invalid HTTP_HOST header in Docker Local: fixes #879

### LINKED ISSUE #879: Invalid HTTP_HOST header in Docker Local
I tried deploying cookie cutter app using Docker locally.

At this step of running
```
docker-compose -f dev.yml

```
Everything is fine, but when accessing http://dockerbox:8000, I got following error
```
django.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: 'dockerdev:8000'. You may need to add 'dockerdev' to ALLOWED_HOSTS.
```
I already set dockerdev in local.py
ALLOWED_HOSTS='.dockerdev'
*dockerdev: is my docker machine box set in /etc/hosts

Any help?

rgcr:
This is a permanent solution for https://github.com/pydanny/cookiecutter-django/issues/879

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -30,8 +30,8 @@ SECRET_KEY = env('DJANGO_SECRET_KEY', default='CHANGEME!!!')
 # SITE CONFIGURATION
 # ------------------------------------------------------------------------------
 # Hosts/domain names that are valid for this site
-# See https://docs.djangoproject.com/en/1.6/ref/settings/#allowed-hosts
-ALLOWED_HOSTS=env.list('DJANGO_ALLOWED_HOSTS', default=['*'])
+# See https://docs.djangoproject.com/en/1.10/ref/settings/#allowed-hosts
+ALLOWED_HOSTS = env.list('DJANGO_ALLOWED_HOSTS', default=[])
 
 # Mail settings
 # ------------------------------------------------------------------------------


{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -27,6 +27,12 @@
 # Note: This key only used for development and testing.
 SECRET_KEY = env('DJANGO_SECRET_KEY', default='CHANGEME!!!')
 
+# SITE CONFIGURATION
+# ------------------------------------------------------------------------------
+# Hosts/domain names that are valid for this site
+# See https://docs.djangoproject.com/en/1.6/ref/settings/#allowed-hosts
```
Sgiath: Why 1.6 version? Actually, the link leads to 404. What about https://docs.djangoproject.com/en/1.10/ref/settings/#allowed-hosts ?

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -27,6 +27,12 @@
...
+ALLOWED_HOSTS=env.list('DJANGO_ALLOWED_HOSTS', default=['*'])
```
Sgiath: Why is the default `['*']`? According to the doc:
> When DEBUG is `True` and ALLOWED_HOSTS is empty, the host is validated against `['localhost', '127.0.0.1', '[::1]']`.

Shouldn't be the default value empty list?

{{cookiecutter.project_slug}}/config/settings/local.py
rgcr: I took the comments from the production.py file, it can be changed

{{cookiecutter.project_slug}}/config/settings/local.py
rgcr: good catch

pydanny:
This should occur only for those using Docker in development. Please put it in a `{% use_docker %}` statement

webyneter:
@rgcr, any news on that?

Braintelligence:
I also stumbled on this problem and had to add ALLOWED_HOSTS to the `local.py`.

browniebroke:
The setting has been updated recently elsewhere:

https://github.com/pydanny/cookiecutter-django/blob/baf08b2f5f14961d0c89876db68dd89c99e38b01/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/local.py#L18-L23

I think this can be closed now @webyneter ?

PR #985, opened 2017-01-09
WIP: load .env file from root dir only if DJANGO_READ_DOT_ENV_FILE is True

luzfcb:
In the [current configuration](https://github.com/pydanny/cookiecutter-django/blob/de4c8c4012757aa544b68cae4904c0383b418bd9/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/common.py#L19), the `.env` file is being loaded by django and this has caused some problems (issues #944 #950 #1000 #1001 ). The `.env` file should be loaded only by Docker and in the production environment, the settings must be configured via the environment variable as define https://12factor.net/config . All available environment variables are listed at https://cookiecutter-django.readthedocs.io/en/latest/settings.html#settings.

However, in some cases, practicality beats purity, like in the development environment. 

To do this, optionally, the `.env` file can be read by Django by manually setting the environment variable on current command line shell (not on `.env` file):

```bash
export DJANGO_READ_DOT_ENV_FILE=True
```

> Note: Remember, never, under any circumstance, commit the `.env` file or equivalent in your git repository. This can introduce serious security breaches, because environment variables commonly contain privated information that should certainly never be public.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/common.py
+++ b/{{cookiecutter.project_slug}}/config/settings/common.py
@@ -24,7 +24,7 @@ env = environ.Env()
 READ_DOT_ENV_FILE = env('DJANGO_READ_DOT_ENV_FILE', default=False)
 
 if READ_DOT_ENV_FILE:
-    # Operating System Environment variables have precende over variables defined in the .env file,
+    # Operating System Environment variables have precedence over variables defined in the .env file,
     # that is to say variables from the .env files will only be used if not defined
     # as environment variables.
     env_file = str(ROOT_DIR.path('.env'))
--- a/{{cookiecutter.project_slug}}/config/settings/common.py
+++ b/{{cookiecutter.project_slug}}/config/settings/common.py
@@ -16,7 +16,21 @@ ROOT_DIR = environ.Path(__file__) - 3  # ({{ cookiecutter.project_slug }}/config
 APPS_DIR = ROOT_DIR.path('{{ cookiecutter.project_slug }}')
 
 env = environ.Env()
-env.read_env()
+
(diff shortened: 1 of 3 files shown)


pydanny:
I'm very happy you thought add this to the project. 👍

jayfk:
LGTM 👍

{{cookiecutter.project_slug}}/config/settings/common.py
```diff
@@ -16,7 +16,21 @@
...
-env.read_env()
+
+# Load operating system environment variables and then prepare to use them
+env = environ.Env()
+
+# .env file, should load only in development environment
+READ_DOT_ENV_FILE = env('DJANGO_READ_DOT_ENV_FILE', default=False)
+
+if READ_DOT_ENV_FILE:
+    # Operating System Environment variables have precende over variables defined in the .env file,
```
webyneter: A typo in 'precende'.

luzfcb: @webyneter fixed. Thanks for pointing out this error.

Afrowave:
I was wondering whether there should be a development ```.env``` file. I found that once I started adding ```ALLOWED_HOSTS``` in the *config.settings.local* and these going into the repo, my dev colleagues, on pulling the repo got my local settings and made changes.

What do you guys think of having 2 ```.env``` files, one for development and once for production?

emorozov:
I think Mozilla Pontoon project solves this in a more elegant way:
https://github.com/mozilla/pontoon/blob/master/manage.py
```
warnings.filterwarnings('ignore', module='environ')
```

Issue #490, opened 2016-02-24
How do I use the env.sample?


pcompassion:
There's a env.sample file but I can't find a note on how I can use it in the documentation.

Eraldo:
I guess you are talking about `env.example`.

1. You make a copy of it and call it `.env` (hidden in many OS GUIs)

2. Then you enter the data for your environment.
Example: `DJANGO_SECRET_KEY=asdasd;kfj3@#$asdkfjasdjkfk3G4ASDFASF`
.oO( cat on keyboard *giggle* )

3. In your settings file (e.g. `common.py`) you can read it like so: `env.read_env(ROOT_DIR('.env'))` depending on where you put your copy.

Does that help @pcompassion?

pydanny:
`env.example` is documented in https://github.com/pydanny/cookiecutter-django/blob/8a7ecb0ca74a4e3376eb3bc79dcaedd6f3e94244/docs/deployment-with-docker.rst. 

Thanks for answering @Eraldo! You think maybe we should expand that documentation?

pcompassion:
yes it helps.. (env.example is needed for installation without docker so I think it's proper to include in the other install doc)

iamwithnail:
Thanks for that answer @Eraldo , just helped me out of a long-standing jam.  Having come recently to the cookie-cutter world, the usage of the env.example/.env file isn't obvious, particularly for those of us that haven't worked up to Docker yet - expanding the explanation would be really helpful.

audreyr:
If anyone wants to add this to the documentation, we'd love a pull request 😄

shevitz:
I'm also a cookiecutter newbie and agree how to use the .env file is not obvious. 

Where is the best practices place to put the .env file? Where sample.env is? in config/settings/, outside of the repository? 

Also where to read the .env file. In common.py, local.py, or someplace else?

thanks

IntelBob:
Yeah, this definitely needs better documentation especially for someone who is new to Django. It took some trial and error to figure out that I have to add `env.read_env(ROOT_DIR('.env'))` to `common.py`. For the uninitiated, add it after this line `env = environ.Env()`.

pydanny:
@Intelbob, while you're code works, the .env file is meant to be read by configuration tools as environment variables, not read directly in the python code. It's meant to provide a seperation of concerns for Docker-powered tools, not as a general coding method.

Sigh. This needs to be documented better. The problem is lack of free time. :(

[...]

PR #1001, opened 2017-01-20
Look for .env file in root dir

philippeluickx:
Based on https://github.com/pydanny/cookiecutter-django/issues/490 and some other discussions.

Docker is looking for a .env file in root directory. But with the old code, django settings will expect the file in /config/settings directory. This change fixes this, so that django settings will look for .env file in the root directory.


philippeluickx:
https://github.com/pydanny/cookiecutter-django/pull/985 is a better implementation

PR #1060, opened 2017-02-28
gitignore .cache/

webyneter:
Ignore py.test-generated .cache/


webyneter:
I'd rather separate these commits into separate PRs, yet I realized they would be put together too late...

PR #1067, opened 2017-03-04
Updated aws cache type to t2.micro

MightySCollins:
The cache type which is currently used is now no longer (https://aws.amazon.com/elasticache/previous-generation/). Now the t2 nodes are meant to be both cheaper and better.
When deploying a new project to AWS this the creation failed and took hours of my life to debug and I hope it can help someone else :)


matt7277:
Cool, thanks for the fix! I was stumped by this as well

pydanny:
Thanks!

Issue #850, opened 2016-10-14
Use Argon2 Password Hasher?


jhancia:
Hello, noticed that the Django 1.10 docs recommend to use the Argon2 password hasher rather than the defaults.  Any thoughts on using this?  Apparently it requires an additional 3rd party library.  See [here](https://docs.djangoproject.com/en/1.10/releases/1.10/#django-contrib-auth) and [here](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#argon2-usage)

pydanny:
I'm very excited to include Argon2, but the [installation path is a bit worrisome](https://argon2-cffi.readthedocs.io/en/stable/installation.html#installation). According to the docs (linked above), recent `pip` and `setuptools` should account for any problems. That's wonderful, and I really respect the maintainer's efforts, but sometimes things can go haywire (especially when dealing with closed-source vendors like Microsoft and Apple). Considering the size and dynamism of our community, I don't want to dump a ton of work on the maintainer of argon2-cffi unless they are certain their package is ready.

Therefore, I'm going to ask the maintainer of argon2-cffi, @hynek, the following question: Considering that we have a LOT of Windows users for Cookiecutter Django, are you okay with us using your argon2-cffi library for password hashing?

hynek:
I think that ship has sailed once Django added and endorsed it. :)  As with Python 3, once Django encourage something, things tend to start moving.

FWIW, Windows is least of my concerns since it‘s shipping wheels from day 1 (for exactly those reasons; I knew I could rather work out building wheels than supporting people building a C extension).

So sure, go for it.

pydanny:
Thanks for the response @hynek. 😄 

@jhancia, did you want to put together a pull request?

jhancia:
Sure, I can give it a go.

pydanny:
Closed by #1079

PR #1079, opened 2017-03-15
Use Argon2 password hashing algorithm by default

archinal:
As per the recommendation in the Django docs [here](https://docs.djangoproject.com/en/dev/topics/auth/passwords/#using-argon2-with-django) to use the argon2 password hashing algorithm, this pull request enables that hashing algorithm by default for users of cookiecutter-django.

I noticed that this change was discussed and approved in [this issue](https://github.com/pydanny/cookiecutter-django/issues/850), but the aforementioned pull request to implement it was never produced.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -141,6 +141,7 @@ Listed in alphabetical order.
   Vitaly Babiy
   Vivian Guillen           `@viviangb`_
   Will Farley              `@goldhand`_                 @g01dhand
+  William Archinal         `@archinal`_
   Yaroslav Halchenko
 ========================== ============================ ==============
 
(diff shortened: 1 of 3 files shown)


webyneter:
@archinal, shouldn't we also be using [bcrypt](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#using-bcrypt-with-django) for the complete picture? :)

archinal:
@webyneter As I understand it we need to choose **one** hasher to be used to store our passwords (from [the docs](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#how-django-stores-passwords): "The first entry in this list (that is, settings.PASSWORD_HASHERS[0]) will be used to store passwords, and all the other entries are valid hashers that can be used to check existing passwords")

I've already included bcrypt in the list of hashers, which means that Django can use bcrypt to check **pre-existing** passwords if they themselves were encrypted with bcrypt, but as per [the recommendation of the Password hashing competition](https://password-hashing.net/), my PR intentionally lists Argon2 as the hashing algorithm for **new passwords**.

webyneter:
@archinal, yes, you're right; I've missed on that.

pydanny:
I love security-related PRs. 😺

Issue #1082, opened 2017-03-16
Introduce a single source of truth for project domain name


webyneter:
For now, project domain name is determined at the initial, project scaffolding stage only, for it to then be hard-coded into numerous files, -- it might take quite some time to accurately update all the occurrences once required.

Since the domain name is used across multiple file types, -- 

* python files (including  migrations), 
* templates, 
* shell scripts,

-- introducing an **environment variable representing the domain name** might be the single feasible solution, IMO. As for the variable name, it might be `DOMAIN_NAME`.

To ensure the variable is accessible project-wide, -- and, given the fact that `.env` file comprising other variables is for production use only, -- I suggest that we create a separate, say, `common.env` file with the variable definition, referencing it through `env_file` setting in both `dev.yml` and `docker-compose.yml` like that:
```yaml
env_file:
- common.env
- .env
```
(Elaborating on that even further, my other suggestion would be to contain 'orphan' variables scattered all around  `dev.yml` in a separate `dev.env` file which can be stored in VCS likewise the development-specific variables it defines are.)
Yet, only a single `.env` file must be available at production, so we'd need to concatenate `common.env` and `.env` in an automatic manner (incorporating this functionality in, say, `./compose/django/create-env.sh`, called upon in `django` service's `command` in `docker-compose.yml` before `./compose/django/gunicorn.sh` is called should suffice).

To handle that, special treatment would be required for various kinds of files:

* python files (migrations included):
```python
import environ
env = environ.Env()
# reference the variable via env('DOMAIN_NAME')
```
* templates: pass `DOMAIN_NAME` through the corresponding view's context;
* shell scripts: so far, `./compose/nginx/start.sh` is the only place the domain name is utilized, and happily it's the only place where the problem is solved already (there, the use of `nginx` service-provided `MY_DOMAIN_NAME` variable is made, but referencing `common.env` in the `nginx` service's `env_file` would be one further enhancement).

I'm in huge appreciation of any critics and suggestions to follow; that would also be a pleasure for me to provide a PR ones we reach consensus on the issue.

browniebroke:
I think the situation has improved slightly with changes from #1801

We might be able to simplify this a bit further, but not by much: https://github.com/pydanny/cookiecutter-django/search?q=DOMAIN_NAME&unscoped_q=DOMAIN_NAME

What do you think @webyneter? Is it still an issue?

browniebroke:
Here is where it's used at the moment:

[...]

PR #1087, opened 2017-03-20
Reference dev.yml's 'orphan' environment variables from local.env

webyneter:
**NB: This PR is effectively frozen and is going to be closed automatically once #1295 is merged.**

This is to elaborate on the _side_ proposal made in #1082. 

Contain 'orphan' variables scattered all around dev.yml in a separate dev.env file which can be stored in VCS likewise the development-specific variables it defines are.

N.B. `django` service's `USE_DOCKER=yes` variable left intact for its use to be isolated within the service only.

1. [ ] move all development-time env variables declarations from `./local.yml` to `./.env.local`;
1. [ ] generate placeholder values for the keys in `./.env.local` from those provided during client project generation;
1. [ ] `.gitignore` `./.env*` as by the @jayfk's suggestion;
1. [ ] (optionally) update both the cookiecutter-django's and the generated project's docs;
1. [ ] test everything out manually.


{{cookiecutter.project_slug}}/dev.yml
```diff
@@ -19,7 +19,8 @@ services:
       dockerfile: ./compose/django/Dockerfile-dev
     command: /start-dev.sh
     depends_on:
-      - postgres
+      - postgres{% if cookiecutter.use_mailhog == 'y' %}
```
luzfcb: jinja2 have a sintax to control newlines/whitespace

http://jinja.pocoo.org/docs/2.9/templates/#whitespace-control

webyneter: Does it mean one may not worry about line breaks in jinja2 should `trim_blocks` and `lstrip_blocks` settings be set?

luzfcb: Github email fooled me with wrong diff. So my comment was unnecessary because you did not touch those lines. sorry.

in relation of Jinja2 sintax, you could use `{%-` instead of `{%` to suppress new line before and after `if`


```yml
    depends_on:
      - postgres
{%- if cookiecutter.use_mailhog == 'y' %}
      - mailhog
{%- endif %}
    env_file:
      - dev.env
```

the rendered result is

```yml
    depends_on:
      - postgres
      - mailhog
    env_file:
      - dev.env
```

if you use `{%`, the rendered result is will be

```yml
    depends_on:
      - postgres

      - mailhog

    env_file:
      - dev.env
```

`-%}` It can also be useful in some cases.

To explore syntax options a bit more, this site may be useful http://jinja2test.tk/

webyneter: Thank you for clarifying! I'll explore the options in more detail, and alter the PR accordingly.

jayfk:
I would call the file `.env.local`. This way we can add `.env*` to `.gitignore`.

webyneter:
@jayfk , this is one great suggestion!

webyneter:
@jayfk , on the other hand, gitignoring the `.env.local` would prevent us from ensuring this bit of reproducibility across multiple local hosts utilizing the same code base, which is what I was trying to achieve in the first place (and this is just a DRY way to declare shared docker service envs)

jayfk:
Hehe. That's the whole reason I've hardcoded them in `dev.yml` in the first place.

webyneter:
@jayfk ,then what do you think on extracting and grouping those together in `.env.local`?

webyneter:
My suggestion is to
1. [ ] move all development-time env variables declarations from `./local.yml` to `./.env.local`;
1. [ ] generate placeholder values for the keys in `./.env.local` from those provided during client project generation;
1. [ ] `.gitignore` `./.env*` as by the @jayfk's suggestion;
1. [ ] (optionally) update both the cookiecutter-django's and the generated project's docs.

pydanny:
:+:

jayfk:
While I think about it, maybe we should group all the environment variables by their respective services in a folder.

```
.env.local/
  - .django
  - .postgres
  - .whatever
.env.production/
  - .django
  - .postgres
  - .whatever
```

jayfk:
On top of that, we should move all the `.env` files in a post gen hook so that there's no need to rename them by hand.

PR #1088, opened 2017-03-20
[proposal] Cookiecutter option to introduce `domain` app

webyneter:
The option allows the user to utilize DDD: If confirmed, a `domain` package will be created in apps dir, and referenced in `./config/settings/base.py`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -265,10 +265,26 @@ Support This Project
 
 This project is maintained by volunteers. Support their efforts by spreading the word about:
 
-.. image:: https://s3.amazonaws.com/tsacademy/images/tsa-logo-250x60-transparent-01.png
-   :name: Two Scoops Academy
+Two Scoops Press
+~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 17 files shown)


pydanny:
I'm curious: How does adding a (mostly) empty app called `domain` turn this into a Domain-Driven design project?

webyneter:
@pydanny, you're right, it doesn't, but it still might be considered a first step towards DDD, saving up on a few keystrokes (just a few indeed, but still :).

pydanny:
Upon reflection, I think adding this is a bit out of scope. That said, how about one or more of the following:

* Create a separate script that adds this tooling outside of Cookiecutter Django.
* Add documentation of this technique to a pull request to Cookiecutter Django.

What do you think?

webyneter:
@pydanny, I agree, that would be a handy addition -- I'm gonna elaborate on that as soon as I can.

pydanny:
I'm going to close this, as its a documentaton issue, not a code thing.

PR #1092, opened 2017-03-24
Replace links section with depends_on in docker-compose config files

### LINKED ISSUE #1091: [question] Should we replace links with depends_on in docker-compose configs?
It's stated [here, in Step two — removing links](https://medium.com/@giorgioto/docker-compose-yml-from-v1-to-v2-3c0f8bb7a48e#.uj9y6325k) that since `version: 2` of docker-compose, `links` section can be dropped safely in favour of `depends_on`. 

Yet, not only we've been using both `links` and `depends_on` in our configs so far, but also those sections' contents differ, e.g.
```yaml
depends_on:
    - postgres
links:
    - postgres
    - mailhog
``` 
in `dev.yml` -- was this made for a reason?

webyneter:
Closes #1091

### DIFF
--- a/{{cookiecutter.project_slug}}/dev.yml
+++ b/{{cookiecutter.project_slug}}/dev.yml
@@ -19,7 +19,8 @@ services:
       dockerfile: ./compose/django/Dockerfile-dev
     command: /start-dev.sh
     depends_on:
-      - postgres
+      - postgres{% if cookiecutter.use_mailhog == 'y' %}
+      - mailhog{% endif %}
     environment:
       - POSTGRES_USER={{cookiecutter.project_slug}}
       - USE_DOCKER=yes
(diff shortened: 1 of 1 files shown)


webyneter:
@jayfk, the [build](https://travis-ci.org/pydanny/cookiecutter-django?branch=master) has failed with `RecursionError: maximum recursion depth exceeded`. [It's evident from logs](https://travis-ci.org/pydanny/cookiecutter-django/jobs/214741982#L804) that it was caused by an erroneously generated `SECRET_KEY`.

Surprisingly, I've hit the very same issue recently, dealing with exactly the same exception at one of my projects. This is [a known problem](https://github.com/joke2k/django-environ/issues/60), and **the workaround is to avoid assigning environment variables values starting with `$`** (provided the values are being generated, one either has to manually remove `$`, or restart the process in the hope of value to be generated without the leading `$`). (Apparently, the root cause has nothing to do with the PR.) 

So, I suggest that we try restarting Travis pipeline.

PR #1112, opened 2017-04-06
Switch to current runtime version

### LINKED ISSUE #1094: Switch to current runtime version in runtime.txt
On the one hand, the `django` image is created `FROM python:3.5`, resolving to `3.5.3` [as of now](https://hub.docker.com/_/python/). On the other hand, we use `python-3.5.1` in `./runtime.txt` -- there seems to be a mismatch.

Provided my suggestion is valid, I can to provide a PR upgrading `./runtime.txt` to  `python-3.5.3`.

webyneter:
Closes #1094


luzfcb:
@webyneter  :ship:  thanks.

PR #1142, opened 2017-04-24
Add a pre-hook that checks the python version

### LINKED ISSUE #1141: Add a pre-hook that checks the python version
Since we no longer support Python 2.7 in target environments we should have a pre-hook that if called by Python 2.7, produces a warning stating and a continuation (y/N) that the project won't run on Python 2.7. It won't stop the project generation, but will warn the user that they need to run it in Python 3.

webyneter:
Closes #1141


webyneter:
Didn't come up with a way to exit "more" gracefully except for naive `sys.exit(1)`: Is there some kind of intrinsic  `Cookiecutter` mechanism for preventing further execution once a hook has "failed" of sort and exiting generation process at the same time with zero?

jayfk:
I think `sys.exit` is fine. Thoughts on this?

pydanny:
`sys.exit(1)` is fine with me! Thanks @webyneter!

Issue #1132, opened 2017-04-21
Switch to Caddy on Docker


jayfk:
Caddy is so much easier to set up using let's encrypt than nginx. A perfect fit for Cookiecutter Django.

webyneter:
@jayfk, I've heard "rumours" (yeah, sort of "some people say"-stuff) that Caddy is not yet ready for wide real-world adoptioin, or I am wrong?

jayfk:
I've been using Caddy for a lot of smaller and mid-sized sites and I'm really happy with it. Djangopackages.org runs Caddy for ~6 months, no problems.

I wouldn't use it for very large sites, but the only reason is that nginx Plus offers some great enterprise features.

webyneter:
@jayfk, what do you think about introducing a separate 'nginx-master' branch off of `master` for dedicated nginx support while migrating the mainline to Caddy to further cherrypick Caddy-unrelated `master` commits to `nginx-master` so keeping both options?

webyneter:
@jayfk, even better: We could introduce these two mutually-exclusive cookiecutter-django template options, and stick with one branch.

jayfk:
I'm -1 on that. Supporting multiple servers is a huge pain in the long run. Either Caddy or nginx.

I'm in huge favor of Caddy. It's simple, has a great config file format, supports Lets Encrypt with secure defaults out of the box and it's super easy to add basic load balancing capabilities. A prereq for #1133. 

@pydanny @audreyr @luzfcb @burhan @theskumar @webyneter  and everyone interested. Can I get a vote on that? 👍 for caddy, 👎 against it.

webyneter:
@jayfk , I trust your experience wholeheartedly on this :smile:.

webyneter:
@jayfk, how is this going? Can I be of any help?

bdrhoa:
Just for reference, I don't think caddy works on AWS when using and AWS Cert and load balancer. Here's a related article. https://msaizar.com/blog/cookiecutter-django-nginx-route-53-and-elb/

PR #1154, opened 2017-05-04
Install ps on the nginx container and fix docker-compose.yml when using letsencrypt

jangeador:
The file docker-compose.yml ends up with the port option defined twice when using letsencrypt. The arrangement of the if statements in the PR fixes that issue.

The file `start.sh` in the nginx compose folder has the command `kill $(ps aux | grep 'nginx' | grep -v 'grep' | awk '{print $2}')`. The problem is that the `ps` command is not installed in the nginx image. The command added in the PR installs the `ps` command so that `start.sh` runs without error.

This eliminates a problem that I was having where nginx was reporting that port `80` was already in use. The error in the logs was:

```
nginx_1         | /start.sh: 59: /start.sh: ps: not found
nginx_1         | /start.sh: 59: kill: Usage: kill [-s sigspec | -signum | -sigspec] [pid | job]... or
nginx_1         | kill -l [exitstatus]
nginx_1         | 2017/05/04 16:20:08 [emerg] 26#26: bind() to 0.0.0.0:80 failed (98: Address already in use)
nginx_1         | nginx: [emerg] bind() to 0.0.0.0:80 failed (98: Address already in use)
```


ashishnitinpatil:
The `RUN` command addition works for me.

luzfcb:
LGMT

jangeador:
@luzfcb can we get this merged?

jayfk:
Thanks! This is a needed fix. I don't know how long it will last, though: #1132

jangeador:
@jayfk I have never used Caddy, but so far every decision made by this community has been accurate. Looking forward to adopt it if the community goes that way. Thanks for the merge.

PR #1158, opened 2017-05-05
Optimize Django Dockerfile

### LINKED ISSUE #1157: Improve the docker image.
The docker image could be improved with a few tricks:

* Reduce the number of layers: Each instruction in the Dockerfile creates a new layer.
* Do not cache pip data: There is no point of caching the packages in a container, `--no-cache-dir` should be used
* Remove clutter files
* Adding the code should be one of the last steps. Every step in the Dockerfile is cached so it has not to be executed everytime it changes. The code being the step that will change the most it must be one of the last steps.

Here's an example of the optimized Dockerfile:

```
FROM python:3.6
ENV PYTHONUNBUFFERED 1

# Requirements have to be pulled and installed here, otherwise caching won't work
COPY ./requirements /requirements
COPY ./compose/django/gunicorn.sh ./compose/django/entrypoint.sh /

RUN pip install --no-cache-dir -r /requirements/production.txt \
    && groupadd -r django \
    && useradd -r -g django django \
    && sed -i 's/\r//' /entrypoint.sh \
    && sed -i 's/\r//' /gunicorn.sh \
    && chmod +x /entrypoint.sh \
    && chown django /entrypoint.sh \
    && chmod +x /gunicorn.sh \
    && chown django /gunicorn.sh \
    && rm -rf /requirements

COPY . /app
RUN  chown -R django /app

USER django

WORKDIR /app

ENTRYPOINT ["/entrypoint.sh"]
```

If you would like those improvements I can create a pull request.

Darkheir:
Closes #1157.

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/django/Dockerfile
@@ -1,15 +1,16 @@
 FROM python:3.5
+
 ENV PYTHONUNBUFFERED 1
+
+RUN groupadd -r django \
+    && useradd -r -g django django
 
 # Requirements have to be pulled and installed here, otherwise caching won't work
 COPY ./requirements /requirements
+RUN pip install --no-cache-dir -r /requirements/production.txt \
+    && rm -rf /requirements
 
-RUN pip install -r /requirements/production.txt \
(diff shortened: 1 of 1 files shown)


webyneter:
This is OK with me.

PR #1164, opened 2017-05-14
Fix Deployment with Docker docs ALLOWED_HOSTS reference

webyneter:


webyneter:
@jayfk what do you think?

Issue #1173, opened 2017-05-18
Nginx gets stuck in redirect loop on production


ashishnitinpatil:
When accessing the production (with SSL), I was getting a 301 for every request.

ashishnitinpatil:
Found a possible solution that works for me -

```
# config/nginx/nginx-secure.conf

        ...
        # cookiecutter-django app
        location @proxy_to_app {
+           proxy_set_header X-Forwarded-Proto https;
+           proxy_set_header X-Url-Scheme $scheme;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header Host $http_host;
            proxy_redirect off;
        ...
```

PR #1174, opened 2017-05-18
[Fix] #1173 Nginx SSL redirect loop

ashishnitinpatil:
Headers not being passed was causing nginx to redirect to SSL again & again.


ashishnitinpatil:
Tested on my production setup.

PR #1175, opened 2017-05-19
Fix docs ipdb docker-compose needed command

chrisRubiano:
'run' must be after the -f flag, otherwise it doesn't run


webyneter:
@chrisRubiano, nice, thanks!

PR #1184, opened 2017-06-02
Update runtime.txt

jangeador:
The current setting prevents the app from deploying in heroku as it has a version of python which is not supported by the heroku framework. This update fixes the deployment on heroku.

### DIFF
--- a/{{cookiecutter.project_slug}}/runtime.txt
+++ b/{{cookiecutter.project_slug}}/runtime.txt
@@ -1,1 +1,1 @@
-python-3.5.3
+python-3.6.1


webyneter:
@jangeador, thank you for your efforts, but I don't remember for us to officially support Python 3.6 (even though this workaround might fix Heroku deployment issues), are we, @pydanny?

auvipy:
heroku now supports py3.6

luzfcb:
Closed in favor of https://github.com/pydanny/cookiecutter-django/pull/1239

PR #1185, opened 2017-06-02
Fix typo and deprecated option in certbot command

chrisRubiano:
The --standalone-supported-challenges option has been deprecated since certbot version 0.9.0.


webyneter:
@chrisRubiano, thanks!

PR #1128, opened 2017-04-21
Node.JS + docker-compose = ♥

webyneter:
Introducing a long-awaited (by myself, first and foremost :smile: ) Node.JS docker-compose integration. The scope of the PR is local/debug-only configuration.

Aspects addressed are:
* `gulp`-powered static asset processing and bundling:
    * `package.json` extended;
    * `gulpfile.js` rewritten, just to name a few,
        * tasks overloaded with responsibilities were broken up into a number of small, atomic, single-purpose units called in a sequential/parallel manner; 
        * [`pump`](https://www.npmjs.com/package/pump) used [to handle task failures](https://github.com/terinjokes/gulp-uglify/tree/master/docs/why-use-pump#why-use-pump);
        * `pathsConfig` section replaced the previous solution to allow for extensive asset paths reuse;
        * `browserSync` made to play nicely with `watch` task;
        * ... 
    * all static assets are to be found in `static/build/`, which is
        * `.gitignored`;
        * referenced from `base.py` to facilitate `manage.py collectstatic`;
        * referenced from `base.html`, providing `js` and `css` assets;
        * referenced from `gulpfile.js` for obvious reasons :smile:.
* `docker-compose` `node` integration:
    * `node` service built from `node:7.7-slim`-based `Dockerfile` (empirically found out to be the minimal working `node` distribution) defined, with `BrowserSync` main and UI ports exposed, and `command` set to `gulp`;
    * `django` service now `depends_on` `node`.

Note that due to the way `node_modules` is being used as `node` volume (thus allowing the former to be accessible from the latter), one cannot directly read/modify `node_modules` from within docker host filesystem (which is perfectly OK provided we agree on interacting with `node` via docker only).

Moreover, the [`standard`](https://github.com/feross/standard) js code style is installed and utilized (which IMO is the best one out there to follow, mainly due to its popularity therefore wider acceptance in general), but this might not be an option for a range of projects, and I'm ready to pull this off into a separate PR allowing for js code style selection from a predefined set of most well-known ones at `cookiecutter-django` initialization stage.
 
The proposed configuration has been tested with the latest `docker` and `docker-compose` versions, however I invite everyone to review the PR to present its weak spots, suggest enhancements, etc. 

My strong belief is that the PR **must not** be accepted unless accompanied with
* [x] extensive documentation update suggested some time ago in #604 , and I can deliver on that a bit later;
* [x] remove `cookiecutter` initialization message `You selected to use docker and a JS task runner. This is NOT supported out of the box for now...`

[...]

PR #1205, opened 2017-06-21
Following up on Node.JS + docker-compose = ♥ (#1128)

### LINKED ISSUE #1204: docker-compose up static files cannot be found by wsgi nor nginx
Steps to reproduce:
Using the docker CE in windows, version Version 17.03.1-ce-win12 

1. create a project using this template whitenoise=n, docker=y, windows=y, pycharm=y, postgres=9.6,js_task_runner=none,celery=y, opbeat=n, sentry=n, timezone=America/Panama, mailhog=n, heroku=n
2. rename env.example to .env
3. docker-compose build
4. docker-compose up

Static fields are not served, collectstatic is working and copies files to the destination dir
however only templates are served.

webyneter:
## Problem
Static assets are not available in production.

## Cause
Static assets don't exist in the project's `static/build` given the stack is run with production configuration immediately after project generation.

## Solution
Make sure static assets are built by the time `docker-compose build` gets invoked. 
I see two ways to ensure that:

1. (the easiest, least invasive one) [Document](https://cookiecutter-django.readthedocs.io/en/latest/deployment-with-docker.html) the preliminary step of running the `node` service from development configuration prior to moving to production stage (`build`, `up` etc.):
```shell
docker-compose - f local.yml run -d node
# Wait for build to complete.
# Remove the container afterwards.
```
1. (clean, but presumptive of Docker version) Opt out for [multi-stage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/) with the `node`'s Dockerfile essentially included in the `django`'s Dockerfile as a preliminary build stage; however, for this to work out one of the latest (+17.05) versions of Docker must be running on host machine which might not be the case for the majority of hosts out there for now. 

I personally vote for the first option due to its implementation straightforwardness and ease of maintenance. @jayfk what do you think?

## TODO
* [ ] Disable `node`-powered gulp if grunt was selected (in fact, the whole changeset must be "reverted" i.e. extra files removed and template artifacts not generated).
* [ ] (???) #1228.
* [x] Account for #1227.
* [x] Account for #1257.
* [ ] <s>Upgrade `node` to the latest (pinned) `slim` version.</s>
* Make sure docs are updated:
    * [ ] the project README;
    * [ ] Node.JS-Docker workflow guides.

Closes #1204.


bluesurfer:
@webyneter 

What kind of benefits brings node.js w.r.t. the previous production setup? I'm just curious.

webyneter:
@bluesurfer, none: I've realized the real problem is that static assets are not available to `django` image on its build phase (since at the `django` image build stage we copy the contents of the project dir to the image file system while in dev setup the assets are built by `node` *after* all the images are built and are accessed by `django` through the `/app` volume mapping available both to `django` and `node`).

webyneter:
@bluesurfer, for now the only way to fix this seamlessly is to ensure static assets are *already* compiled before running `docker-compose build`, so they can then be copied to `django` during build.

However, this is an ugly workaround in the long-run, and I/we have yet to come up with minimally invasive solution to this (I guess, the concept of Docker Compose pre-build actions should be further explored.) 

@pydanny, @jayfk what do you think?

webyneter:
@bluesurfer, @chopanpma, I am rolling back my yesterday's #1128 PR which led to this issue.

webyneter:
The upcoming changes are to succeed those introduced in #1128.

badri:
hi @webyneter. What's the current status of this PR? Why are the commits in https://github.com/pydanny/cookiecutter-django/pull/1128 reverted back? I think that's the way to compile static assets before deploying to production. It eliminates the need to have all node + node related dependencies in the dev env. Let me know if anything can be done from my end.

webyneter:
@badri,

> I think that's the way to compile static assets before deploying to production. It eliminates the need to have all node + node related dependencies in the dev env.

what do you mean?

ArcadiaLandTrust:
Hello @webyneter et al, I'm wondering if the gulp/docker-compose project is still in the works? I'd really love to try it on my current project. If it is not being incorporated, do you have any tips or links for setting it up manually? Would it be safe to copy your setup from before the reversion? 

Thanks and I appreciate all your hard work, really inspiring!

webyneter:
@ArcadiaLandTrust, it's been a while since I last checked out on this, but I hope I will have some time to spare in the coming weeks to review things out here.

To put things into perspective, the initial PR [#1128](https://github.com/pydanny/cookiecutter-django/pull/1128) introducing Docker Compose + Node.JS integration caused static assets not being available in production. As of now, this PR has been far behind `master` for quite some time. 

However, you can still try and integrate Node.JS with your setup by means of:
1. scaffolding a new project with cookiecutter-django;
2. adding/editing project files added/edited over the course of [#1128](https://github.com/pydanny/cookiecutter-django/pull/1128).
3. see how it works alltogether, especially when `docker-compose -f production.yml up` (it will fail but why wouldn't it, though? :)
4. fix the problem either by
4.1. resisting the temptation of running `docker-compose -f production.yml up` ever again, lol (well, actually, unless this PR is resolved :), or
4.2. ensure static assets are created prior to having your production up-n-running every time you need to.

That's how I see it. Ones again, it's been a while since I last worked on this one...

ArcadiaLandTrust:
@webyneter Thanks so much for the thoughtful reply!

browniebroke:
I'm going to close this now that #1687 is merged. We might want to reuse some work from here, but I don't think there is much point in keeping it open.

PR #1209, opened 2017-06-26
Change to process *.scss instead of only project.scss.

omin:


webyneter:
@omin, thanks! I should ask, have you tested this?

omin:
Yep. Well, it depends on what the author was trying to achieve. If he/she was trying to concatenate to a single file, we need another solution.

Sames goes for processing javascript files with gulp btw.

PR #1227, opened 2017-07-10
Rename dev.yml to local.yml

### LINKED ISSUE #1226: Rename dev.yml to local.yml
That would match up with `local.py` and `local.txt`. 

@pydanny, @jayfk, @luzfcb what do you think?

webyneter:
Closes #1226.


jayfk:
For consistency, maybe we should rename `docker-compose.yml` too: https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/docker-compose.yml

webyneter:
@jayfk I agree, but that would prevent us from being able to conveniently `docker-compose buld`/`up`...

webyneter:
@jayfk, on the other hand, "explicit is better than implicit", so we'll at the same time be safeguarded from accidentally invoking Docker Compose with production configuration.

jayfk:
> @jayfk, on the other hand, "explicit is better than implicit", so we'll at the same time be safeguarded from accidentally invoking Docker Compose with production configuration.

This. It's better to be safe than sorry. Starting production locally may result in side effects for the *real* production setup or your users (if you are using a subset of real user data locally).

PR #1251, opened 2017-07-25
moved to a python entry point, This will give more control to python programmers 

### LINKED ISSUE #1250: why is entrypoint not a py script?
my issue with the current setup is that i cannot easily debug my entry point script.
in order to debug i need to constantly add and remove print statements, It would be more friendly for 
python users if it was written in python, since part of the stock script already is I don't think this is an unusual request.

 ```
48 #!/bin/bash
 47 set -e
 46 cmd="$@"
 45
 44 # This entrypoint is used to play nicely with the current cookiecutter configuration.
 43 # Since docker-compose relies heavily on environment variables itself for configuration, we'd have to define multiple
 42 # environment variables just to support cookiecutter out of the box. That makes no sense, so this little entrypoint
 41 # does all this for us.
 40 export REDIS_URL=redis://redis:6379
 39
 38 # Read from secrets
 37 post_user=$(cat "$POSTGRES_USER_FILE")
 36 post_pass=$(cat "$POSTGRES_PASSWORD_FILE")
 35
 34 # the official postgres image uses 'postgres' as default user if not set explictly.
 33 if [ -z "$post_user" ]; then
 32     #export POSTGRES_USER=postgres
 31     >&2 echo "Some secret is missing"
 30 fi
 29
 28 export CELERY_BROKER_URL=$REDIS_URL/0
 27
 26 function postgres_ready(){
 25 python << END
 24 import sys
 23 import psycopg2
 22 import urllib.parse as parse
 21 try:
 20     dburl = '$DATABASE_URL'
 19     dburl.format(username='$secret_user', password='$secret_pass')
 18     sys.stderr.write('$secret_pass')
 17     sys.stderr.write(dburl)
 16     conn = psycopg2.connect(dburl, connect_timeout=3)
 15 except psycopg2.OperationalError as error:
 14     print(error)
 13     sys.exit(-1)
 12 sys.exit(0)
 11 END
 10 }
  9
  8 until postgres_ready; do
  7   #post_user=`cat $POSTGRES_USER_FILE`
  6   #post_pass=`cat $POSTGRES_PASSWORD_FILE`
  5   >&2 echo "Postgres is unavailable - sleeping"
  4   sleep 1
  3 done
  2
  1 >&2 echo "Postgres is up - continuing..."
49  exec $cmd
```

japrogramer:
Since half of the sh entry-point is already written in python, why not have all of it written in python?

Closes #1250.


webyneter:
Overall, the intent is positive IMO, however for consistency we would need to migrate our whole infrastructure to python scripts and, therefore, other existing shell scripts also should be reviewed. Although it's possible to switch to python-based scripting, to the best of my knowledge shell scripts are the industry standard (ideally, POSIX-compliant variants of those) thanks to the omnipresence of the latter.

japrogramer:
I agree that the rest of the shell scripts should be ported to python. One very good reason for doing this is for test coverage. As of now there are no tests for any of this scripts and considering that this scripts are basically the entire application for the containers which the main app will depend so heavily, it seems odd that the only way to find out if something might be wrong is to build the image and run it. I know that they are small scripts that aren't often touched or altered but if the only way to debug them is manually.

webyneter:
@japrogramer you can test shell scripts via [bats](https://github.com/sstephenson/bats).

japrogramer:
@webyneter Thats fine, but it looks to me like its, just braking out each line out and testing it. which than limits testing how things interact together if one outcome happens earlier in the script.

jayfk:
As @webyneter already mentioned, using shell scripts as an entry point for docker images is a de facto standard. Having this written in Python has its benefits, but it's also surprising for people that join the project later on.

I know you've put a lot of hard work into this. I'm sorry, but I'm going to close this.

Issue #710, opened 2016-08-03
Set CONN_MAX_AGE in DATABASE_URL


audreyr:
[CONN_MAX_AGE](https://docs.djangoproject.com/en/1.9/ref/databases/#persistent-connections) defaults to 0. I think it would be nice for us to be more opinionated about it, setting it to a non-zero value by default.

[StackOverflow](http://stackoverflow.com/questions/19937257/what-is-a-good-value-for-conn-max-age-in-django) suggests 60, [RevSys](http://www.revsys.com/blog/2015/may/06/django-performance-simple-things/) suggests 600 (10 min). I lean toward 600 since we're using allauth and authenticated users tend to stick around a site longer.

theskumar:
I would go with 1min, as 10mins way to much for default, having a larger value can have hard to debug issue, soaking up db connection limit. 

With 1min: 
1) we are providing a placeholder which people can see and update as needed (still better if configurable with env variable) 
2) 60sec is a decent for moderately crowded site.

Issue #270, opened 2015-07-13
database configuration unclear


musicformellons:
I run into problems when installing. Instruction says 
>Then, create a PostgreSQL database and add the database configuration using the dj-database-url >app pattern: postgres://db_owner:password@dbserver_ip:port/db_name either:
>
>in the config.settings.common.py setting file,
>or in the environment variable DATABASE_URL"

Remarks:
1) I get this error: "FATAL:  password authentication failed for user "cookie" "; cookie is my db_owner. For dbserver_ip:port  I use 127.0.0.1:5432
2) config.settings.common.py should be config/settings/common.py 
3) I used the common.py file; I am curious though how exactly I would set the environment variable DATABASE_URL ; I tried to look it up but do not get clear answers.
4) it seems to implicate one should install dj-database-url ...? More detail please.

Please help.

pydanny:
What is the precise value that you used for `postgres://db_owner:password@dbserver_ip:port/db_name`?

musicformellons:
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql_psycopg2',
        'NAME': 'cookieDB',
        'USER': 'cookie',
        'PASSWORD': 'mypassword',
        'HOST': '127.0.0.1',
        'PORT': '5432',
    }}

DATABASES['default']['ATOMIC_REQUESTS'] = True


2015-07-13 22:27 GMT+02:00 Daniel Greenfeld <notifications@github.com>:

musicformellons:
I also tried (but get same error):
DATABASES = {
    # Raises ImproperlyConfigured exception if DATABASE_URL not in
os.environ
    'default': env.db("DATABASE_URL", default="postgres://
cookie:mypassword@127.0.0.1:5432/cookieDB"),
}


2015-07-14 16:21 GMT+02:00 mailsietse . <mailsietse@gmail.com>:

luzfcb:
please try to set the environment variable: 
`DATABASE_URL` environment variable

into terminal:

```bash
export DATABASE_URL=postgres://cookie:mypassword@127.0.0.1:5432/cookieDB
```

in the same terminal, run:

```
python manage.py migrate
python manage.py runserver
```

about question 4: 
[django-environ](https://github.com/joke2k/django-environ) use the same [api](https://github.com/joke2k/django-environ/blob/master/environ/environ.py#L315) of dj-database-url

musicformellons:
did that, but error remains.

[...]

PR #1252, opened 2017-07-26
make the database and .env cookiecutter connection more clear

NathanQ:
in regards to issues #490, #270 and #710

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -112,8 +112,8 @@ DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql',
         'NAME': '{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.project_slug}}',
-        'USER': env.db('DATABASE_USER', default=''),
-        'PASSWORD': env.db('DATABASE_PASSWORD', default=''),
+        'USER': env.db('POSTGRES_USER', default=''),
+        'PASSWORD': env.db('POSTGRES_PASSWORD', default=''),
         'HOST': 'localhost',
         'PORT': '',
         'CONN_MAX_AGE': 0
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -109,7 +109,15 @@ MANAGERS = ADMINS
 # ------------------------------------------------------------------------------
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
 DATABASES = {
-    'default': env.db('DATABASE_URL', default='postgres://{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.project_slug}}'),
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': '{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.project_slug}}',
+        'USER': env.db('DATABASE_USER', default=''),
+        'PASSWORD': env.db('DATABASE_PASSWORD', default=''),
+        'HOST': '127.0.0.1',
+        'PORT': '',
+        'CONN_MAX_AGE': 0
+    }
 }
 DATABASES['default']['ATOMIC_REQUESTS'] = True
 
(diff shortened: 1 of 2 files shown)


webyneter:
Overall, the intent is positive IMO, however for consistency we would need to migrate our whole infrastructure to python scripts and, therefore, other existing shell scripts also should be reviewed. Although it's possible to switch to python-based scripting, to the best of my knowledge shell scripts are the industry standard (ideally, POSIX-compliant variants of those) thanks to the omnipresence of the latter.

luzfcb:
POSTGRES_PASSWORD and POSTGRES_USER only is used because is required by [docker-entrypoint.sh](https://github.com/docker-library/postgres/blob/972294a377463156c8d61297320c872fc7d370a9/9.6/docker-entrypoint.sh) script of Postgresql docker image

NathanQ:
Thanks for the feedback @webyneter. I feel there is a need to resemble the [Django docs database setting](https://docs.djangoproject.com/en/1.11/ref/settings/#databases) to allow settings to be more easily managed.

Basis for PR:
#270 and #490
```
# base.py
DATABASES = {
    'default': env.db('DATABASE_URL', default='postgres://{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.project_slug}}'),
}
```
and
```
# example.env
POSTGRES_PASSWORD=mysecretpass
POSTGRES_USER=postgresuser
```
If the env file is to be used, it seems like the database setting should include them so the env file can be used to set them. If it's to be rewritten to use the env settings, it's not very cookiecutter.

#710 
Adding `CONN_MAX_AGE` to the cookiecutter allows for this to be an accessible setting.

luzfcb:
By default, cookiecutter-django does not use the .env file, ( https://12factor.net/config ). It is only used if you choose to use docker, and only to automate the creation of environment variables during the build of the docker images.


About CONN_MAX_AGE, this can be included with something like this
```python
DATABASES = {
    'default': env.db('DATABASE_URL', default='postgres://{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.project_slug}}'),
}
DATABASES['default']['ATOMIC_REQUESTS'] = True
DATABASES['default']['CONN_MAX_AGE'] = env.int("DJANGO_CONN_MAX_AGE", default=0)
```

NathanQ:
@luzfcb Fair enough and good point. Fwiw, env files are used by those who don't yet use docker.

The primary point of the pr is to address the mentioned issues. The issues are rooted in the bit of obfuscation between cookiecutter-django's base settings, its corresponding example.env and those of Django docs examples along with available settings. Changing from an url to an object structure to mirror "official" doc examples makes it straightforward to those who rely on documentation.

japrogramer:
I know this thread is about making .env file more clear for database settings .. but I don't think there should be a .env file at all for docker deployments. 
Hear me out.
If a user really needs to set an enviorment variable it can be done from a compose file.
Now for sensitive stuff, I believe that stuff doesn't even belong in environment variables the reason
for this believe: too often are these settings logged out to file and than you have your secrets scattered everywhere.. now you have too worry about sanitizing your logs and even worse securing your logs.
These is my solution, I have gone and removed the .env file, and only rely on docker secrets. And I don't even use environment variables, because they too often get logged by programs, and if you are keeping logs that is a security risk. The file that i read in as a secret is structured like a json object so i can easily parse in python. Each container has a different set of secrets and they only see what they need to see. I do not have the secrets stored in plain txt anywhere instead I archive the swarm dir in docker which stores encrypted secrets.
#on a manager node
```
$docker secret create my_secret -
$ tee -
#mind you, my secrets are mounted into my containers from within their respective compose file
$systemctl stop docker.service 
$tar czpvf swarm-dev.tar.gz /var/lib/docker/swarm/
```
than i rsync that file over an encrypted pipe back home
and when i need to restore my secrets i do 
```
$systemctl stop docker.service 
$rm -fR /var/lib/docker/swarm
$tar xzpvf swarm-dev.tar.gz -C /var/lib/docker/swarm/
$systemctl start docker.service 
$docker swarm init --force-new-cluster
$ tee -
#Now all my secrets are back, and available to my swarm. But only to those containers that have been given permission on a per secret basis.
```

webyneter:
> I feel there is a need to resemble the Django docs database setting to allow settings to be more easily managed.

@NathanQ yes, I agree now. 

@luzfcb `POSTGRES_*` envs are also widely used in backup/restore shell scripts as well as the docs. By the way, @NathanQ, the PR doesn't address all these occurrences -- it would be great if you'd fix this project-wide.

jayfk:
@webyneter already mentioned this: The environment variables are used all over the project in all kinds of different places. A pull request needs to address this properly.

PR #1260, opened 2017-07-30
Gulpfile default task stuck on runServer, not running browsersync and watch tasks

orzarchi:


webyneter:
Works (tested locally).

PR #1282, opened 2017-08-11
switch to caddy

### LINKED ISSUE #471: Lets Encrypt
Lets Encrypt is in public beta since a couple of weeks and it would be nice if we could document a way for people to use it
- [x] on docker
- [ ] on heroku
- [ ] on ubuntu

Personally, I don't like the official client because it does too much. [simp_le](https://github.com/kuba/simp_le) looks promising.

Any thoughts? Is someone already using Lets Encrypt on a side project?

### LINKED ISSUE #596: Adding nginx-proxy
[nginx-proxy](https://github.com/jwilder/nginx-proxy) offers at least the following: 
- simple nginx load-balancing
- simple zero downtime deployments
- a [letsencrypt companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) using simp.le, that handles letsencrypt better than what is currently used in this project.
- multiple sites on one IP address.

The first three could be added out of the box, and I could add docs on how to modify the file structure to support the fourth. 

Would it be best as an option or could nginx-proxy be in docker-compose by default? Is this beyond the scope of the project?

I can put together a PR if it is wanted.

### LINKED ISSUE #622: Document that Cookiecutter Django is secure by default
The security settings are set for HTTPS only by default, sessions and cookies just don't stick unless you have HTTPS. This isn't explained anywhere, which can trip up newer Docker users.

### LINKED ISSUE #670: Fix SuspiciousOperation errors
As @pydanny already predicted such problems here https://github.com/pydanny/cookiecutter-django/pull/319, i think we have to choose to silence these errors by configuring either the [logger](http://stackoverflow.com/questions/15238506/djangos-suspiciousoperation-invalid-http-host-header/17477436#17477436) or the [webserver](https://docs.djangoproject.com/en/1.9/topics/logging/) .

### LINKED ISSUE #941: Add shebang to the project's /compose/nginx/start.sh
Noticed the shebang line is missing in the project's /compose/nginx/start.sh.

### LINKED ISSUE #1097: Why try_files in nginx.conf?
Why is there `try_files` in nginx.conf? Nginx container doesn't have any static files from the project copied or mounted, won't it just make everything slower checking on non-existent files all the time?

https://github.com/pydanny/cookiecutter-django/blob/7da1bf79b807290423bb97262043a4f6e67403cb/%7B%7Bcookiecutter.project_slug%7D%7D/compose/nginx/nginx.conf#L49

### LINKED ISSUE #1132: Switch to Caddy on Docker

### LINKED ISSUE #1273: Don't use the test switch in the certbot command
Why is `--test` included in the certbot command in `production.yml`? This will cause the cert to be issued by a fake issuer.

Or, at least make it clearer that if you use certbot to create certificates, be sure to remove the `--test` flag when you're ready to go into production.

jayfk:
This PR replaces nginx with caddy.

- closes #1273
- closes #1132 
- closes #1097 
- closes #941 
- closes #670
- closes #622 
- closes #596 
- closes #471


jayfk:
Tests are passing, tested on two droplets, I'm merging this. Goodbye `nginx` 👋 !

webyneter:
@jayfk wow, that was fast!

webyneter:
@jayfk I think we should also update the README, adding Caddy support to the list of features.

jayfk:
Yeah, we needed to move fast on this.  `nginx` with `certbot` was complex, confusing and had a lot of issues. 

If there are some problems with `caddy` we'll address them as they come in.

jayfk:
> @jayfk I think we should also update the README, adding Caddy support to the list of features.

Added in abf8f2c

thaxy:
The docs still referencing to the nginx setup, this is confusing. Also it is not clear for me why this should resolve #622.

jayfk:
> Also it is not clear for me why this should resolve #622.

Caddy comes with LetsEncrypt out of the box. It's the default so there's no longer a need to mention it explicitly.

> The docs still referencing to the nginx setup, this is confusing.

Currently working on it.

thaxy:
@jayfk Ah ok, I think I get it now. We enforcing HTTPS (through LetsEncrypt) and since it is much easier to configure you switched out the web server. So that means there will be no option in the future to choose between HTTP and HTTPS (which is a good thing), right?

jayfk:
Exactly. But, If you really want to, you can still use HTTP by setting `tls off`.

ssteinerx:
How does this work in testing as certificates can only be issued on real domains?

jayfk:
> How does this work in testing as certificates can only be issued on real domains?

Multiple options:

- sync your production cert with your staging environment (not recommendend)
- issue a self signed cert
- disable HTTPS support

natecox:
@jayfk Caddy announced yesterday that they now have a commercial licence which prevents the use of distributed binaries in commercial use without a paid licence. Can you tell me what the impact of that is on this PR?

see https://caddyserver.com/blog/accouncing-caddy-commercial-licenses

jayfk:
> Can you tell me what the impact of that is on this PR?

That I'm going to revert it :). That was a horrible move by the whole caddy team. On top of that, they are already threatening forks: https://github.com/WedgeServer/wedge/issues/2

natecox:
Yeah, I was all set to switch my projects over to Caddy until I read the press release.

Apparently you _can_ use it commercially for free if you compile the open source version instead of using the binaries, but I'm not certain that's feasible here.

jayfk:
It probably is. We could just build our own Caddy Docker image from source. But honestly, I don't want to support projects that are doing things like this.

I'm a bit sad to say this because I've enjoyed using Caddy, but it's pretty much dead for me now.

webyneter:
:-1:

Afrowave:
I was afraid of this. Caddy was too good to be left to NOT make money. Fortunately I have only used it on a personal project and was very happy with the experience. So @jayfk, are we going back to NginX and Let's Encrypt?

jayfk:
> Caddy was too good to be left to NOT make money.

Well, this isn't about money. It's about business tactics. Nonfree binaries, server headers with ads, threatening forks are 3 major red flags.

They should have just hired some people to do consulting and released a couple of nonfree plugins like nginx does. I'd be happy to pay a couple hundred for a plugin that allows to do something like rate limiting. It's a shame.

On top of that, the business license is too expensive for the typical Cookiecutter Django usecase.

jayfk:
> So @jayfk, are we going back to NginX and Let's Encrypt?

Probably. But we'll need to find an easier way to make this work.

Afrowave:
@jayfk,

Have you had a look at this? [WedgeServer](https://github.com/WedgeServer/wedge/releases) as a drop in replacement to Caddy?

bluesurfer:
@jayfk 

Don't get me wrong but why would we use Let's Encrypt? 

In my own setup, when I deploy the application on a testing machine I generate my own unsigned certificate using OpenSSL and configure NginX accordingly (who cares if it isn't safe?). 
Whereas, for the production environment I simply went to the boss office and made him buy a signed certificate :D

Isn't this the most common scenario nowadays?

pydanny:
> In my own setup, when I deploy the application on a testing machine I generate my own unsigned certificate using OpenSSL and configure NginX accordingly (who cares if it isn't safe?).

**We core developers care if it isn't safe.** We refuse to have unsafe practices in a project with our name on it used by thousands of projects around the world. The last think we need is to be blamed for a hack because we weren't properly secure by default.

thaxy:
> So @jayfk, are we going back to NginX and Let's Encrypt?
> Probably. But we'll need to find an easier way to make this work.

What do you mean by easier way? Unsatisfied with the [previous solution](https://github.com/pydanny/cookiecutter-django/blob/035dc4d7ab7e5d8bce392896473f11056ef51347/%7B%7Bcookiecutter.project_slug%7D%7D/compose/nginx/start.sh) of the cert generation?
Whats the main problem here with nginx and lets encrypt anyways? Nginx is widely used and with a correct conf file I guess we could solve almost  all the stated issues in this PR. So the only ugly thing left is the retrieval of the certs from lets encrypt via certbot (or another way).

jayfk:
Yep and thats a big ugly thing. Using certbot just feels like a giant hack when compared to caddy.

cardonator:
On the deploy side, have you looked at something like https://github.com/jwilder/nginx-proxy?

I'm using that for cookie cutter deploys quite frequently, using docker. You can connect a letsencrypt container to it that manages your host certs.

jimmyye:
https://github.com/Neilpang/acme.sh How about this one?

- An ACME protocol client written purely in Shell (Unix shell) language.
- Full ACME protocol implementation.
- Simple, powerful and very easy to use. You only need 3 minutes to learn it.
- Bash, dash and sh compatible.
- Simplest shell script for Let's Encrypt free certificate client.
- Purely written in Shell with no dependencies on python or the official Let's Encrypt client.
- Just one script to issue, renew and install your certificates automatically.
- DOES NOT require `root/sudoer` access.
- Docker friendly
- IPv6 support

chidg:
Hi guys, don't mean to hassle but would like to know if there has been any further decision making on this. I am starting a new project, hopefully using cookiecutter django, and I would like some certainty on the likely direction to be taken in terms of server and ssl configuration.

drdaeman:
Just a suggestion - what do you think about Træfik?

It has ACME support built in, and while its primary use case is automatic service discovery, it has a [`file` backend](https://docs.traefik.io/configuration/backends/file/#simple) that accepts static configuration and it can be used as a simple LB/reverse proxy with TLS termination, just like nginx or Caddy.

Unlike nginx, it can't serve static, though, but I don't think current Caddfile has it either.

Alwerdani:
Me and ( a lot ) not from caddy fans and I always prefer Nginx. so can we add Nginx as an optional choice in project initialization to meet all of the developers needs?

achillesrasquinha:
Could we consider NGINX as a choice instead of providing Caddy only?

PR #1284, opened 2017-08-13
Add tota11y to base.html

ad-m:
Hello,

I suggest add tota11y to base.html. It's toolbar like ```django-debug-toolbar```, but for accessibility. tota11y helps visualize how site performs with assistive technologies.
Caring for people with disabilities is important and should be a standard feature in producing high quality websites, so it is advisable to provide the tools that support this process.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
@@ -100,6 +100,10 @@
       <script src="{% static 'js/project.js' %}"></script>
       {% endraw %}{% if cookiecutter.use_compressor == "y" %}{% raw %}{% endcompress %}{% endraw %}{% endif %}{% raw %}
 
+      {% if debug %}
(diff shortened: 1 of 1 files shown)


jayfk:
This is good. But please include it as an asset that can be loaded from the filesystem.

+1

ad-m:
@jayfk: I can't. I do not know enough JS tools (Grunt / gulp).

webyneter:
@jayfk is there a reason that lib cannot be loaded dynamically?

jayfk:
@webyneter I think it's a good practice in general. GitHub can go down, the content can change etc.

ad-m:
@jayfk , the content can't change. This pin to specific commit on GitHub, not master. CDN caches it too.

jayfk:
> @jayfk , the content can't change. This pin to specific commit on GitHub, not master. CDN caches it too.

If the user deletes his account, the repo is free to be taken over. Sure it's just a CSS file that's only served in debug mode, but including it as an asset doesn't hurt.

https://www.theregister.co.uk/2018/02/10/github_account_name_reuse/

I'm going to close this PR but feel free to reopen it and include the file as an asset.

PR #1295, opened 2017-08-24
Group environment variables by the corresponding directories

webyneter:
This one builds upon #1087, and @jayfk's proposal to organize env files hierarchically [here](https://github.com/pydanny/cookiecutter-django/pull/1087#issuecomment-321753113) and [here](https://github.com/pydanny/cookiecutter-django/pull/1087#issuecomment-321755776).

I've implemented a prototypical solution based on the idea that, while being put in VCS throughout cookiecutter-django lifecycle `.envs/**/` get ignored via a post-gen hook ones the client project is generated.

* [x] Move the contents of `.env` into multiple environment-specific dirs as by @jayfk's suggestion.
* [x]  Persist `.envs/**/*` files into cookiecutter-django's VCS.
* [x] Gitignore those locally.
* [x] Reference the newly created `.envs/**/.*` files in `local.yml`.
* [x] Reference the newly created `.envs/**/.*` files in `production.yml`.
* Delete `.env.example` leaving `.env` unignored here and ignored on the client machine:
    * [x] delete `.env.example`;
    * [x] post-gen `make_secret_key` over `.envs/.production/.django`;
    * [x] generate a default `POSTGRES_PASSWORD` conforming to the expected format (has to do with PostgreSQL initialization failing given the random string generated via the same algorithm as `DJANGO_SECRET_KEY` is being);
    * figure out how to dynamically populate `.env` with definitions from multple `.envs/.production/*` files (this one presents quite a challenge since env files are interpreted as-is and there's no directive to import one into another; besides, one should keep in mind that envs are subject to change throughout project development lifecycle leading to ones-persisted `.env` values to get out of sync with the latest project state almost instantly, and definitely so by the time of the first production deployment; seems like will have to come up with an automated, on-demand solution):
        * [x] implement an automation script;
        * [x] create shared PyCharm Run Configuration for the automation script.
    * [x] (?) `.dockerignore` `.envs/`;
    * [x] (?) opt out (for the end user, via the initialization menu, with the `y` default) to commit `.envs/local/*` to VCS to ensure development environment reproducibility;
    * [x] clean up any occurrences of/references to `.env.example`;
    * [x] `keep_local_envs_in_vcs` cookiecutter option;
    * [x] docs.

Closes #1087.


{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
```diff
@@ -0,0 +1,65 @@
+import os
```
jayfk: What is this file for?

{{cookiecutter.project_slug}}/.envs/.production/.postgres
```diff
@@ -0,0 +1,2 @@
+POSTGRES_USER=
```
jayfk: I wonder if it would make sense to pre-populate these with random strings? Just like the secret key?

{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
jayfk: What is this file for?

{{cookiecutter.project_slug}}/.envs/.local/.postgres
```diff
@@ -0,0 +1,2 @@
+POSTGRES_USER={{cookiecutter.project_slug}}
+POSTGRES_PASSWORD=postgrespassword
```
jayfk: I wonder if it would make sense to pre-populate these with random strings? Just like the secret key?

{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
webyneter: To merge multiple `.envs/.production/*` files in a specific order in a single `.env`; effectively, this is supposed to be done manually ones any of the `.envs/.production/*` changes.

webyneter: Since we won't be providing `env.example`/`.env` out-of-box anymore I thought that would the easiest way to let the client have an up-to-date version of `.env` throughout the project lifecycle.

{{cookiecutter.project_slug}}/.envs/.local/.postgres
webyneter: This one's addressed.

{{cookiecutter.project_slug}}/.envs/.production/.postgres
webyneter: This one's now addressed.

{{cookiecutter.project_slug}}/.envs/.local/.postgres
webyneter: Gotta also ensure the generated keys match `POSTGRES_PASSWORD` format (it took Travis a few tries before finishing green, and I guess it might be the case with PostgreSQL failing to create `POSTGRES_USER` with the generated `POSTGRES_PASSWORD`).

webyneter:
@jayfk what do you think?

webyneter:
Removed duplicate comments from @jayfk for clarity.

hooks/pre_gen_project.py
```diff
@@ -6,27 +6,27 @@
...
 """
+from __future__ import print_function
+
+import sys
+
+TERMINATOR = "\x1b[0m"
+WARNING = "\x1b[1;33m [WARNING]: "
+INFO = "\x1b[1;33m [INFO]: "
+HINT = "\x1b[3;33m"
+SUCCESS = "\x1b[1;32m [SUCCESS]: "
```
browniebroke: Could we share them with `post_gen_project.py`?

{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
```diff
@@ -0,0 +1,69 @@
...
+import pytest
+
+ROOT_DIR_PATH = os.path.dirname(os.path.realpath(__file__))
+PRODUCTION_DOTENVS_DIR_PATH = os.path.join(ROOT_DIR_PATH, '.envs', '.production')
+PRODUCTION_DOTENV_FILE_PATHS = [
+    os.path.join(PRODUCTION_DOTENVS_DIR_PATH, '.django'),
+    os.path.join(PRODUCTION_DOTENVS_DIR_PATH, '.postgres'),
+    os.path.join(PRODUCTION_DOTENVS_DIR_PATH, '.caddy'),
+]
+DOTENV_FILE_PATH = os.path.join(ROOT_DIR_PATH, '.env')
```
browniebroke: It looks like this file is Python 3 only? If so I'd recommend using [Pathlib](https://docs.python.org/3/library/pathlib.html), it has a nicer API to work with paths IMO.

hooks/pre_gen_project.py
webyneter: Nope, tried that already: `cookiecutter` refuses to recognize `hooks` as a package, with or without `__init__.py` inside which prevented from refactoring these declarations into a distinct module. Maybe @pydanny could suggest something?

{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
webyneter: Thanks for pointing this out! That's the 'issue' with almost the entire Cookicutter Django. To my mind, we should start a separate PR covering Pathlib migration.

webyneter:
Done. @jayfk, @luzfcb, thoughts?

jayfk:
@webyneter Can you do a rebase?

webyneter:
@jayfk what for? :)

jayfk:
@webyneter the diff is really hard to follow as it contains 31 commits :D

jayfk:
Nevermind, I did a local rebase. The GitHub diff is crap anyway.

webyneter:
@jayfk oh, the diff thing. OK, will keep in mind. :)

PR #1307, opened 2017-08-28
Remove pycharm service from Docker setup

### LINKED ISSUE #1302: Remove pycharm service from Docker setup

webyneter:
* [x] Remove pycharm service from Docker setup.
* [x] Update generated project docs.

Closes  #1302


luzfcb:
@webyneter Will this change anything in the current integration with Pycharm?
Pycharm support was included in https://github.com/pydanny/cookiecutter-django/pull/507 

Much of the discussion took place in https://github.com/pydanny/cookiecutter-django/pull/306

webyneter:
@luzfcb no, PyCharm integration is certainly left intact, except for the `pycharm` Docker service deleted for redundancy reasons: it doesn't prevent `django` service container run via a Run Configuration from stopping ones another container (say, for testing only) is started "in parallel" (should the second one be started against an interpreter from `pycharm` service container, let alone against the same one). To recap, I failed to reproduce the behaviour of the `pycharm` service mentioned in the generated project docs. Nevertheless, this whole initiative could be inherently wrong if you'd help me figure out what I'm missing here.

luzfcb:
Thinking about the current size of the repository, I would like to purge all existing binary files, in order to decrease the size and time of git clone. Unfortunately, I think we can not do this, because it would rewrite git history and maybe it would break all forks (Probably this is something for another issue)

PR #1317, opened 2017-09-04
Re-organize compose/ into environment-specific file groups

### LINKED ISSUE #1316: Re-organize compose/ into environment-specific file groups
```
compose/
-- local/
----- *
-- production/
----- *
```

webyneter:
Closes #1316.


{{cookiecutter.project_slug}}/local.yml
```diff
@@ -19,13 +19,15 @@ services:
...
     ports:
       - "8000:8000"
-    command: /start-dev.sh
+    command: /start.sh
 
   postgres:
-    build: ./compose/postgres
+    build:
+      context: .
+      dockerfile: ./compose/production/postgres/Dockerfile
```
sfdye: @webyneter 

`local.yml` referencing production `postgres` Dockerfile?
I think it might be better keep separate `posrgres/` for local and production

webyneter: @sfdye IMO that's the way to go unless postgres Dockerfile is the same across local and production environments; previously, the same thing happened, btw, when we were referencing one and the same postgres Dockerfile from both the local and production Docker Compose configs. Effectively, that didn't change. Moreover, I suppose local-production parity for postgres service is explicit now.

Issue #1321, opened 2017-09-06
.gitignore ignores ./compose/local/* files


markliederbach:
**Note: for support questions, please use the `cookiecutter-django` tag on stackoverflow**. This repository's issues are reserved for feature requests and bug reports. If you need quick professional paid support for your project, contact support@cookiecutter.io.

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?** Bug

* **What is the current behavior?** When committing a newly generated project to Github, the .gitignore filters out the `local` directory in the `compose` folder. This means that the generated `local.yml` file returns an error if you tried to run it.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** 

1. Generate a project:
```
root@b6296e448afe:/# cookiecutter https://github.com/pydanny/cookiecutter-django
project_name [Project Name]: Blog Backend
project_slug [blog_backend]:
author_name [Daniel Roy Greenfeld]: Mark Liederbach
email [you@example.com]: contact@example.com
description [A short description of the project.]:
domain_name [example.com]: example.com
version [0.1.0]: 0.0.1
timezone [UTC]:
use_whitenoise [y]:
use_celery [n]:
use_mailhog [n]: y
use_sentry_for_error_reporting [y]:
use_opbeat [n]:
use_pycharm [n]: y
windows [n]:
use_docker [n]: y
use_heroku [n]:
use_elasticbeanstalk_experimental [n]:
use_compressor [n]:
Select postgresql_version:
1 - 9.6
2 - 9.5
3 - 9.4
4 - 9.3
5 - 9.2
Choose from 1, 2, 3, 4, 5 [1]: 1
Select js_task_runner:
1 - Gulp
2 - Grunt
3 - None
Choose from 1, 2, 3 [1]: 3
custom_bootstrap_compilation [n]:
Select open_source_license:
1 - MIT
2 - BSD
3 - GPLv3
4 - Apache Software License 2.0
5 - Not open source
Choose from 1, 2, 3, 4, 5 [1]: 1
```
Note that this **does** create a `local` directory at `./compose/local`

2. Push to Github. The folder is ignored.

3. Attempt to run docker-compose with the local.yml file. It won't work.

* **What is the expected behavior?** The local directory should not be ignored.

[...]

PR #1322, opened 2017-09-06
Added Whitelist for /compose/local/ based on issue #1321

markliederbach:
Something in the file above ignores the /compose/local directory, which is not an expected pattern. Barring a reasonable explanation or modification of the above rules, this whitelists that folder. See issue #1321.


webyneter:
@markliederbach thank you very much! When I was working on the PR recently, I forgot to make those changes to `.gitignore` and wondered, why in the world the `./compose/local/` gets ignored :) Yep, sometimes silly mistakes manage to find their way out.

markliederbach:
@webyneter Happy to help!

PR #1332, opened 2017-09-15
Change pep 8 for pycodestyle in docs and project requirements

rlaverde:
- Some little changes to the documentation
- Update the requirement in the project  requirements (I think that pyup didn't catch the rename of the package.)


rlaverde:
The built is failing, but I think It's unrelated to my changes, could somebody restart it?

PR #1335, opened 2017-09-18
Fix anymail v1.0 change to EmailBackend

sfdye:
from https://github.com/anymail/django-anymail/releases
> There are no new breaking changes in the 1.0 release, but a breaking change introduced several months ago in v0.8 is now strictly enforced. If you still have an EMAIL_BACKEND setting that looks like "anymail.backends.espname.EspNameBackend", you'll need to change it to just "anymail.backends.espname.EmailBackend". (Earlier versions had issued a DeprecationWarning. See the v0.8 release notes.)


webyneter:
@sfdye, thank you :)

sfdye:
No problem. By the way I have tested it with my application. So should be good to merge.

PR #1350, opened 2017-09-30
More inline docs for DATABASES variable

chuckus:
Use of database url is not part of Django, but comes from django-environ. Was initially confusing where feature came from, as link points to offical django docs.


webyneter:
@chuckus thanks! Please, add yourself to `CONTRIBUTORS.rst`.

PR #1367, opened 2017-10-18
Fix MediaRootS3BotoStorage: do not override existing files

pchiquet:
see https://github.com/jschneier/django-storages/issues/50

The default settings is AWS_S3_FILE_OVERWRITE = True. This makes sense for static files but not for media files.
see https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings


luzfcb:
@pchiquet thanks. Please, add yourself in https://github.com/pydanny/cookiecutter-django/blob/master/CONTRIBUTORS.rst

Issue #1111, opened 2017-04-05
Update to Django 1.11


jayfk:
I've told the bot to stay on `1.10.x` for now (https://github.com/pydanny/cookiecutter-django/commit/f62e05b95e60c043cb463c66b40ca69c1f2d9c26) and merged the latest security release.

I believe @luzfcb was working on a release checklist. Let's put this here.

luzfcb:
Checklist:

- [ ] Check if all dependencies officialy work with Django 1.11( if not, wait).

  Dependencies that are Django version-dependent ( Check off each of the below when it officially works on 1.11. ):
  
  > If you want cookiecutter-django to be updated faster, please help the below projects in implementing compatibility with Django 1.11
  - [ ] [django-environ](https://github.com/joke2k/django-environ)
  - [ ] [whitenoise](https://github.com/evansd/whitenoise)
  - [ ] [django-braces](https://github.com/brack3t/django-braces)
  - [ ] [django-crispy-forms](https://github.com/django-crispy-forms/django-crispy-forms)
  - [ ] [django-model-utils](https://github.com/carljm/django-model-utils)
  - [ ] [django-allauth](https://github.com/pennersr/django-allauth)
  - [ ] [django-autoslug](https://github.com/neithere/django-autoslug)
  - [ ] [raven-python](https://github.com/getsentry/raven-python)
  - [ ] [django-redis](https://github.com/niwinz/django-redis)
  - [ ] [django-extensions](https://github.com/django-extensions/django-extensions)
  - [ ] [django-test-plus](https://github.com/revsys/django-test-plus)
  - [ ] [factory_boy](https://github.com/FactoryBoy/django-factory_boy)
  - [ ] [django-debug-toolbar](https://github.com/jazzband/django-debug-toolbar)
  - [ ] [pytest-django](https://github.com/pytest-dev/pytest-django)
  - [ ] django-storages-redux (deprecated) - Should be replaced by https://github.com/jschneier/django-storages
  - [ ] [django-anymail](https://github.com/anymail/django-anymail)
  - [ ] [opbeat](https://github.com/opbeat/opbeat_python)

- [ ] Check if dependencies require changes on django settings files.
- [ ] Add Python 3.6 to test suite [tox.ini](https://github.com/pydanny/cookiecutter-django/blob/master/tox.ini#L3), [.travis.yml](https://github.com/pydanny/cookiecutter-django/blob/master/.travis.yml#L15)
[...]

PR #1368, opened 2017-10-21
Upgrade Django and PostgreSQL

axeliodiaz:
Upgrade Django version to 1.11.6
Adding PostgreSQL version: 10.0

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -6,7 +6,7 @@ wheel==0.30.0
 
 
 # Bleeding edge Django
-django==1.10.8 # pyup: >=1.10,<1.11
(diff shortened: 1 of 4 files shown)


{{cookiecutter.project_slug}}/requirements/base.txt
```diff
@@ -6,7 +6,7 @@ wheel==0.30.0
 
 
 # Bleeding edge Django
-django==1.10.8 # pyup: >=1.10,<1.11
+django==1.11.6 # pyup: <=1.11.6
```
rlaverde: I think Its better `# pyup: <2.0`, So pyup will update it if another bugfix is released

jayfk:
We are tracking the upgrade process here: https://github.com/pydanny/cookiecutter-django/issues/1111

PR #1374, opened 2017-10-25
Remove boto related settings that aren't used

svleeuwen:
Since upgrade to boto3 in commit 12db5176d6c4c we don't need AWS_HEADERS anymore

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -100,17 +100,6 @@ AWS_STORAGE_BUCKET_NAME = env('DJANGO_AWS_STORAGE_BUCKET_NAME')
 AWS_AUTO_CREATE_BUCKET = True
 AWS_QUERYSTRING_AUTH = False
 
-# AWS cache settings, don't change unless you know what you're doing:
-AWS_EXPIRY = 60 * 60 * 24 * 7
-
-# TODO See: https://github.com/jschneier/django-storages/issues/47
-# Revert the following and use str after the above-mentioned bug is fixed in
-# either django-storage-redux or boto
-control = 'max-age=%d, s-maxage=%d, must-revalidate' % (AWS_EXPIRY, AWS_EXPIRY)
-AWS_HEADERS = {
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -100,17 +100,6 @@
...
-# AWS cache settings, don't change unless you know what you're doing:
-AWS_EXPIRY = 60 * 60 * 24 * 7
-
-# TODO See: https://github.com/jschneier/django-storages/issues/47
-# Revert the following and use str after the above-mentioned bug is fixed in
-# either django-storage-redux or boto
-control = 'max-age=%d, s-maxage=%d, must-revalidate' % (AWS_EXPIRY, AWS_EXPIRY)
-AWS_HEADERS = {
-    'Cache-Control': bytes(control, encoding='latin-1')
-}
```
browniebroke: Looks like this should be replaced by [`AWS_S3_OBJECT_PARAMETERS`](http://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings) instead? It is [empty by default](https://github.com/jschneier/django-storages/blob/06cf8209bf87524c51988ab8e72c8cf97eaa5b89/storages/backends/s3boto3.py#L185).

svleeuwen:
@browniebroke can you check my commit? I followed the previous style, but I don't know if the bytes workaround is needed in boto3.

browniebroke:
I tested this on a Django deployment I own. I've set the headers using a normal string, with spaces and I uploaded a file.

```python
AWS_MAX_AGE = 60 * 60 * 24 * 7  # 1 Week in seconds
AWS_S3_OBJECT_PARAMETERS = {
    'CacheControl': f'max-age={AWS_MAX_AGE}, s-maxage={AWS_MAX_AGE}, must-revalidate',
}
```

```bash
> curl -I $url
HTTP/2 200
...
cache-control: max-age=604800, s-maxage=604800, must-revalidate
...
```

Therefore, the bytes workaround isn't needed anymore. Thank you for flagging this.

browniebroke:
Thanks for your contribution!

PR #1382, opened 2017-11-01
Set CONN_MAX_AGE to 60 seconds -- fixes #710

### LINKED ISSUE #710: Set CONN_MAX_AGE in DATABASE_URL

browniebroke:
Setting the value in the production config rather than in the environment variable as it seems clearer like this.

### DIFF
--- a/tests/test_cookiecutter_generation.py
+++ b/tests/test_cookiecutter_generation.py
@@ -84,3 +84,14 @@ def test_flake8_compliance(cookies):
         sh.flake8(str(result.project))
     except sh.ErrorReturnCode as e:
         pytest.fail(e)
+
+
+def test_connection_max_age(cookies):
+    result = cookies.bake()
+
+    prod_config_file = result.project.join('config', 'settings', 'production.py')
+    prod_config_lines = prod_config_file.read()
+
+    expected = "DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)"
+
+    assert expected in prod_config_lines, prod_config_lines
--- a/tests/test_cookiecutter_generation.py
+++ b/tests/test_cookiecutter_generation.py
@@ -84,14 +84,3 @@ def test_flake8_compliance(cookies):
         sh.flake8(str(result.project))
     except sh.ErrorReturnCode as e:
         pytest.fail(e)
-
-
-def test_connection_max_age(cookies):
-    result = cookies.bake()
-
-    prod_config_file = result.project.join('config', 'settings', 'production.py')
-    prod_config_lines = prod_config_file.read()
-
-    expected = "DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)"
-
-    assert expected in prod_config_lines
--- a/tests/test_cookiecutter_generation.py
+++ b/tests/test_cookiecutter_generation.py
@@ -94,4 +94,4 @@ def test_connection_max_age(cookies):
 
     expected = "DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)"
 
-    assert expected in prod_config_lines, prod_config_lines
+    assert expected in prod_config_lines
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -182,14 +182,14 @@ DATABASES = {
         'PASSWORD': env('RDS_PASSWORD'),
         'HOST': env('RDS_HOSTNAME'),
         'PORT': env('RDS_PORT'),
-        'CONN_MAX_AGE': 60,
+        'CONN_MAX_AGE': env.int('CONN_MAX_AGE', default=60),
     }
 }
 {% else %}
 # Use the Heroku-style specification
 # Raises ImproperlyConfigured exception if DATABASE_URL not in os.environ
 DATABASES['default'] = env.db('DATABASE_URL')
-DATABASES['default']['CONN_MAX_AGE'] = 60
+DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)
 {%- endif %}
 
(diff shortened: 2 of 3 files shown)


sfdye:
Can you set it as an environment variable in .env? Which if the user didn't set, will default to 0. IMO, `CONN_MAX_AGE` does not really apply to local environment.

browniebroke:
Not sure I understand this:

> IMO, CONN_MAX_AGE does not really apply to local environment.

Is it a comment on something I've done wrong? I've updated `production.py` config, which isn't used by the local environment, as far as I know, and `.env` file has `DJANGO_SETTINGS_MODULE=config.settings.production` in it.

sfdye:
What I mean is have something like:

```python
# base.py

DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=0)
```

Users can then override this value on the `.env` file to whatever they want, you shouldn't just hard-code the value.

```bash
# .env file

CONN_MAX_AGE=60
```

browniebroke:
I see, thanks! I've updated the PR.

sfdye:
@webyneter @luzfcb what do you think?

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -182,12 +182,14 @@
...
         'HOST': env('RDS_HOSTNAME'),
         'PORT': env('RDS_PORT'),
+        'CONN_MAX_AGE': env.int('CONN_MAX_AGE', default=60),
     }
 }
 {% else %}
 # Use the Heroku-style specification
 # Raises ImproperlyConfigured exception if DATABASE_URL not in os.environ
 DATABASES['default'] = env.db('DATABASE_URL')
+DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)
```
webyneter: The `default` argument should be refactored to avoid duplication, IMO.

webyneter: Make it smth like `_DEFAULT_CONN_MAX_AGE` so the name indicates its intricate nature.

browniebroke: Is there a way to set this globally without adding to the list of questions? I know of Jinja's `{% set ... %}` but that would still be duplicated between the 2 files `production.py` and `.env`.

webyneter: @browniebroke I don't think we need to make this a cookiecutter option -- merely introducing `_DEFAULT_CONN_MAX_AGE = 60` to `production.py` should suffice; as for the `env.example`, I suggest that you leave its `CONN_MAX_AGE` blank (`CONN_MAX_AGE=`).

tests/test_cookiecutter_generation.py
```diff
@@ -84,3 +84,14 @@ def test_flake8_compliance(cookies):
...
+
+def test_connection_max_age(cookies):
+    result = cookies.bake()
+
+    prod_config_file = result.project.join('config', 'settings', 'production.py')
+    prod_config_lines = prod_config_file.read()
+
+    expected = "DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)"
+
+    assert expected in prod_config_lines, prod_config_lines
```
webyneter: Why the two of `prod_config_lines` in `assert expected in prod_config_lines, prod_config_lines`?

browniebroke: Forgot to clean that up, the second one was to print the content of the file when the assertion was failing...

{{cookiecutter.project_slug}}/config/settings/production.py
browniebroke: Ok agreed, I will do that. Thanks for your help.

tests/test_cookiecutter_generation.py
```diff
@@ -84,3 +84,14 @@ def test_flake8_compliance(cookies):
...
+
+def test_connection_max_age(cookies):
+    result = cookies.bake()
+
+    prod_config_file = result.project.join('config', 'settings', 'production.py')
+    prod_config_lines = prod_config_file.read()
+
+    expected = "DATABASES['default']['CONN_MAX_AGE'] = env.int('CONN_MAX_AGE', default=60)"
+
+    assert expected in prod_config_lines
```
webyneter: @browniebroke I'm not sure what do we need this test for?..

browniebroke: I've used it to verify that my template variable was rendered properly.

webyneter: I have mixed feelings about that test: Should the default value change the test will also fail albeit altering the default value is a _non-functional_ change.

browniebroke:
I don't have any strong opinions. Any thoughts from the other maintainers?
Who has the final word on this? I'll remove it if no one else object.

webyneter:
@browniebroke please, leave a comment in the conversation, so the context of discussion is preserved.

tests/test_cookiecutter_generation.py
browniebroke: I don't have any strong opinions. Any thoughts from the other maintainers?
Who has the final word on this? I'll remove it if no one else object.

webyneter: @pydanny, @jayfk, @luzfcb what do you think?

pydanny: LGTM

PR #1386, opened 2017-11-13
dairy --> dessert

alexgleason:
https://hooktube.com/watch?v=UcN7SGGoCNI

ice cream is often made from cashews, almonds, coconuts, or other plant-based sources.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,7 +110,7 @@ Two Scoops of Django 1.11
    :alt: Two Scoops of Django
    :target: http://twoscoopspress.org/products/two-scoops-of-django-1-11
 
-Two Scoops of Django is the best dairy-themed Django reference in the universe
+Two Scoops of Django is the best dessert-themed Django reference in the universe
 
 pyup
 ~~~~~~~~~~~~~~~~~~


webyneter:
@alexgleason there's something wrong with this video yet I can't explain what... Feeling contradictory emotions, actually :)

alexgleason:
If you'd like to share your thoughts privately please feel free to send me an email.

At any rate, I do think the word "dessert" is just a more accurate and broadly acceptable term. It encompasses ice cream in every form without the presumption of animal-based ingredients. This would make the readme more generalized for a wide audience of people, including ethical vegans, people who are lactose intolerant, and people who happen to prefer non-dairy ice cream because of the flavor. Of course, it neither excludes people who consume dairy; but for someone like me who does not consider dairy to be "food," it equally does not fall under my definition of dessert anyway. If you don't know this, it won't affect you. Win-win?

webyneter:
@alexgleason it's definitely a win-win, and I'm truly amazed you've pointed this out: never thought of this from that perspective.

pydanny:
On semantics alone you had me.

PR #1429, opened 2018-01-03
Improve custom_bootstrap_compilation option

### LINKED ISSUE #1428: Improvements to custom_bootstrap_compilation
### I'm submitting a ...

  - [x] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

### Do you want to request a *feature* or report a *bug*?

I've recently generated a project with this and selected `custom_bootstrap_compilation=y` with `js_task_runner=Gulp` and it didn't produce the result I expected.

Looking at the template for Gruntfile.js, I can see it has some logic for that, but relies on bower, and as far as I cann see, there is no `bower.json` (was it removed recently as [they recommend using something else](https://bower.io/)?)

### What is the current behavior?

Problems I see:
- The generated base HTML template uses Bootstrap CSS and JS from the CDN
- The generated package.json doesn't reference Bootstrap
- The generated gulpfile doesn't do anything for custom compilation
- The generated project.scss is including individual Bootstrap sass files instead of the whole thing by default.
- The generated Gruntfile.js uses Bootstrap v3 instead of Boostrap v4

### What is the expected behavior?

I think selecting this option should:
- add Bootstrap v4 as dev dependency to package.json, as well as Bootstrap's dependencies (jquery and popper.js)
- Make sure Gulp and Grunt pick up the sources from the Bootstrap NPM package and compile them properly
- HTML templates should use the compiled files rather than link to the CDN ones.

Am I missing anything? Did I understood the expected behaviour of this option properly? All being OK, I'm happy to put together a PR. Might need to correct or extend the documentation on this topic.

Here is the change I've done to my generated project:
https://github.com/browniebroke/carriere_alla/commit/4f786a993f4b195744d7c3fe1869489470f4ab69

browniebroke:
Fixes #1428

Summary of actions:
- [x] Add Bootstrap to package.json in case of custom_bootstrap_compilation
- [x] Update Gulpfile.js and Gruntfile.js to use Bootstrap SASS from the node module
- [x] Update the base HTML to use the CSS from Boostrap's CDN only when appropriate
- [x] Update Gulpfile.js to compile a vendors.js from jQuery, popper.js and Boostrap node modules
- [ ] Update Gruntfile.js to compile a vendors.js from jQuery, popper.js and Boostrap node modules
- [x] Update the base HTML to use the vendor.js when appropriate

### DIFF
--- a/{{cookiecutter.project_slug}}/package.json
+++ b/{{cookiecutter.project_slug}}/package.json
@@ -32,6 +32,9 @@
     "del": "^2.2.2",
     "gulp": "^3.9.1",
     "gulp-autoprefixer": "^3.1.1",
+    {% if cookiecutter.custom_bootstrap_compilation == 'y' %}
+    "gulp-concat": "^2.6.1",
+    {% endif %}
     "gulp-cssnano": "^2.1.2",
     "gulp-imagemin": "^3.0.3",
     "gulp-pixrem": "^1.0.0",
--- a/{{cookiecutter.project_slug}}/package.json
+++ b/{{cookiecutter.project_slug}}/package.json
@@ -6,7 +6,7 @@
     {% if cookiecutter.js_task_runner == 'Grunt' %}
     "autoprefixer-core": "~5.2.1",
     {% if cookiecutter.custom_bootstrap_compilation == 'y' %}
-    "bootstrap": "^4.0.0-beta.2",
+    "bootstrap": "^4.0.0",
     {% endif %}
     "connect-livereload": "~0.3.2",
     "cssnano": "~2.1.0",
@@ -26,7 +26,7 @@
     "time-grunt": "~1.2.1"
     {% elif cookiecutter.js_task_runner == 'Gulp' %}
     {% if cookiecutter.custom_bootstrap_compilation == 'y' %}
-    "bootstrap": "^4.0.0-beta.2",
+    "bootstrap": "^4.0.0",
     {% endif %}
     "browser-sync": "^2.14.0",
     "del": "^2.2.2",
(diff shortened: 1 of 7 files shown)


browniebroke:
For the last piece I wanted to do, I'm afraid I've never used Grunt. Therefore, I'm not sure I'm in a good position to do the change. 
As far as I can see, the Gruntfile isn't doing anything in terms of Javascript currently and seems to rely more on Django Compressor to do the work... If someone more familiar with that bit could shed some lights on this, it would be appreciated.

browniebroke:
Testing for this PR in this repo:
https://github.com/browniebroke/cookiecutter-bs-compilation-tests

{{cookiecutter.project_slug}}/package.json
```diff
@@ -5,21 +5,36 @@
   "devDependencies": {
     {% if cookiecutter.js_task_runner == 'Grunt' %}
     "autoprefixer-core": "~5.2.1",
+    {% if cookiecutter.custom_bootstrap_compilation == 'y' %}
+    "bootstrap": "^4.0.0-beta.2",
+    {% endif %}
```
browniebroke: Bootstrap v4 final has been released:
http://blog.getbootstrap.com/2018/01/18/bootstrap-4/

pydanny:
@browniebroke now that bootstrap v4 is out, should this PR be merged in?

browniebroke:
Updated to Bootstrap v4. Looks ready on my end, thanks.

pydanny:
Thanks so much @browniebroke!

PR #1445, opened 2018-01-12
Remove caddy, use acme.sh as Let's Encrypt client

sfdye:
This PR addresses some of the concerns in #1282 to drop caddy.

Thanks to the comment by @jimmyye , I discovered https://github.com/Neilpang/acme.sh with which I had great success.  

To re-cap, some of the features of **acme.sh**:

- An ACME protocol client written purely in Shell (Unix shell) language.
- Full ACME protocol implementation.
- Simple, powerful and very easy to use. You only need 3 minutes to learn it.
- Bash, dash and sh compatible.
- Simplest shell script for Let's Encrypt free certificate client.
- Purely written in Shell with no dependencies on python or the official Let's Encrypt client.
- Just one script to issue, renew and install your certificates automatically.
- DOES NOT require root/sudoer access.
- Docker friendly
- IPv6 support

I've tested the PR on a fresh AWS EC2 instance, and HTTPS worked with zero additional setup.

Please have a review @jayfk @pydanny @webyneter @luzfcb 

This approach may not be perfect but it's definitely promising and cleaner than certbot.


sfdye:
Re-based the commits, so it's easier to merge.

jayfk:
Thanks for putting this together!

I talked about this a couple time before, I really don't like mounting the Docker socket into nginx (this is what nginx-proxy does). Even in `ro`, there's so much valuable data flying over that socket.

Nginx (like every other software) has a long track record of security vulnerabilities. Now, if someone manages to break into your webserver and on top of that gets access to your raw Docker socket, it's game over.

I'm afraid I'll have to give a 👎 to this :(.

sfdye:
@jayfk Thanks for the review, points well taken. If other maintainers don't have any other opinion, then I will close this. @pydanny @webyneter

badri:
What do you folks think of Traefik? I've used it in couple of other projects which use Docker in production and I've been happy so far. I'd be glad to send a PR.

jayfk:
Thanks for your effort @sfdye but for the reasons pointed out above, I'm going to close this PR.

We are going to accept a PR that comes without the usage of nginx-proxy. If you find the time to work on a solution that uses acme.sh only, feel free to open this PR again or create a new one.

webyneter:
@jayfk don't you think we should switch back to Nginx by no means?

jayfk:
I’d love to switch back to nginx, but with a sane implementation only (default https + no privileged docker container).

If someone wants to work on this, please do! I’m currently fighting for every free minute I can get...

luzfcb:
@jayfk the problem is how to update the SSL certificates inside a no privileged docker container ?

PR #1448, opened 2018-01-13
Add Create React App as a possible front-end choice

genomics-geek:
## Reason
In many of my projects, I use ReactJS for front-end development.  This adds the choice of using react for the frontend in a Django project.  

## Features
1. Integration of React (via Facebook's create-react-app)
1. React Hot Reloading integration - during development, one can make changes it will automatically render on the browser.
1. REST API integration with Django Rest Framework
1. Authentication with Django REST Auth
1. Documentation of API with CoreAPI (now standard in DRF 3.7.7).

## How to use it

After creating the project layout using the cookiecutter, follow the steps below.

Open 2 terminals (one will be for django, the other will be for node).

In the first terminal:
1. `cd frontend`
1. `yarn install` or `npm install`
1. `yarn start`

In the second terminal:
1. create a virtualenv and activate it - `pyenv virtualenv project_name; pyenv activate project_name`
1. `pip install -r requirements/local.txt`
1. Populate `.env` file
1. `python manage.py migrate`
1. `python manage.py createsuperuser`
1. `python manage.py runserver`

Visit http://127.0.0.1:8000/app/ - you should see a welcome to React page.

To authenticate:
Visit http://127.0.0.1:8000/api-auth/login/ and login using the credentials for the superuser

Once authenticated you should be able to see the API docs page at http://127.0.0.1:8000/docs/


jayfk:
Would you be able to package this functionality into a hook that can be called after Cookiecutter Django has finished?

I see no way we can keep up with current react releases/best practices here. Others may have a different view on this, but most of the features here are kept current by people that use that kind of setup in their own projects.

When it comes to react, we'd probably need 2-3 people managing this behemoth. Thoughts?

pydanny:
First off, thesis for this pull request.

However, I agree with @jayfk, this is a huge addition of maintenance burden for us. We're already stretched too far. Even if it's a separate package, that will generate a lot of support requests.

What I suggest instead is that this be turned by @genomics-geek into a maintained fork, just like we have with Foundation. We'll link to it in the README. What do you think @genomics-geek?

genomics-geek:
@pydanny and @jayfk  - I figured as much and what you suggest makes a lot of sense.  I agree, we can have a different maintained fork.  That works for me!

Afrowave:
@genomics-geek Thank you very much for this. I was about to embark on this road. I have come into some time to do open source work and Cookiecutter Django was at the top of the list.

I would like to contribute to this React fork. I found a deprecated version and this "diallowed" version. Which is it?

genomics-geek:
@Afrowave this is the fork to use.  This fork uses create-react-app

jayfk:
Awesome. Please let us know once the fork is ready so we can link to it in the docs.

genomics-geek:
Actually @Afrowave - do you want to take the lead on this? Seems like you have more time than I would to actively maintain this.  I kind of do this on the side/for fun :).  Let me know

Afrowave:
@genomics-geek Whoa! That was fast.  I must admit I am a bit behind the curve. But this is it.
Let me set up the fork, build the tests and send @jayfk the link. Hmm, we will need to do our own docs. 

Let us do this. 😁 😄

genomics-geek:
@Afrowave - awesome! Send me the fork once you set it up. Thanks a lot - excited for the collaboration!

PR #1457, opened 2018-01-17
Fixed old docker-compose filenames

reggieriser:
The post-generation hook references the old docker-compose filenames of `dev.yml` and `docker-compose.yml` when cleaning up files in a non-Docker project.  Those files were renamed a while back to `local.yml` and `production.yml`.  See related commit e69ceebfc584373bfec446033dd1de2fa48560fe.


pydanny:
Thanks @reggieriser!

PR #1467, opened 2018-01-26
Fixes wsgi.py app_path to avoid deleting all /config occurrences.

### LINKED ISSUE #1466: wsgi.py: Bug in app_path when project path includes /config not just at the end
This bug occurs because the resulting path has replaced every ocurrence of /config in the project path:

``` python
# config/wsgi.py:

app_path = os.path.dirname(os.path.abspath(__file__)).replace('/config', '')
```

If project path is something like: ```/var/www/config/whatever/src/config/wsgi.py``` the above line is going to convert it into ```/var/www/whatever/src``` when it sould be ```/var/www/config/whatever/src```.

**This fixes the problem:**

``` python
# config/wsgi.py:

app_path = os.path.abspath(os.path.join(
    os.path.dirname(os.path.abspath(__file__)), os.pardir))
```

ddiazpinto:
Fixes #1466 using parent directory instead of replacing /config in app_path.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/wsgi.py
+++ b/{{cookiecutter.project_slug}}/config/wsgi.py
@@ -20,7 +20,8 @@ from django.core.wsgi import get_wsgi_application
 
 # This allows easy placement of apps within the interior
 # {{ cookiecutter.project_slug }} directory.
-app_path = os.path.dirname(os.path.abspath(__file__)).replace('/config', '')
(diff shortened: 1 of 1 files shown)


joseGabriel06:
it not solve problem at heroku.

pydanny:
@joseGabriel06 please explain in detail what's wrong.

PR #1418, opened 2017-12-28
Fixing leftovers from migration to Django 1.11 and Python 3.6.

reggieriser:
I've been using Django 1.11 and Python 3.6 in a cookiecutter-django-based project for the last month or so.  This PR contains some potential related cleanups given that @pydanny has now moved the baseline to 1.11/3.6.

PR #1482, opened 2018-02-05
 Move to the python:alpine docker image 

### LINKED ISSUE #498: Move to the python:alpine docker image

browniebroke:
Fixes #498 

- Switch the base images for local and production to alpine
- Install extra dependencies for psycopg2, Pillow and cffi
- Change shebang for shell scripts to use sh instead of bash, as bash isn't installed
- Remove `function` keyword from shell scripts, it is unsupported in sh
- Migrate `django` group and user creation to the Alpine syntax

I've started from a generated project, which I've migrated by running `docker-compose -f local.yml build` and `docker-compose -f production.yml build`. The tests are passing with the local config and I can start with the production config.


{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
```diff
@@ -1,7 +1,16 @@
-FROM python:3.5
+FROM python:3.5-alpine
```
webyneter: Make it `FROM python:3.6-alpine`.

{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
```diff
@@ -1,9 +1,18 @@
-FROM python:3.5
+FROM python:3.5-alpine
```
webyneter: Make it `FROM python:3.6-alpine`.

jayfk:
Another thing to keep in mind is support for translations.

browniebroke:
Thanks for pointing this out, I'll further testing around this.

browniebroke:
I think it's still missing something to work out of the box, a `locale` folder. Suggest to add it with a README in it?

browniebroke:
I think it's merge-ready from my end. Anything to add?

tests/test_docker.sh
```diff
@@ -19,3 +19,6 @@ docker-compose -f local.yml run django python manage.py test
 
 # return non-zero status code if there are migrations that have not been created
 docker-compose -f local.yml run django python manage.py makemigrations --dry-run --check || { echo "ERROR: there were changes in the models, but migration listed above have not been created and are not saved in version control"; exit 1; }
+
+# Test support for translations
+docker-compose -f local.yml run django python manage.py makemessages
```
webyneter: I'd make it `docker-compose -f local.yml run --rm django python manage.py makemessages`.

pydanny:
is this good to go? I want to make #1484 happen!

jayfk:
This one should be good. 

It isn‘t blocking anything, though. We are already using Python 3.6 in Docker.

browniebroke:
@jayfk is correct, #1418 unblocked it. Going to update this branch from master now.

PR #1485, opened 2018-02-06
Use AWS IAM roles

### LINKED ISSUE #1453: Use AWS roles instead of AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY
If the server is compromised, the attacker can easily get `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` from the environment,  with which they can do anything. 

In stead, the AWS best practice is to create an IAM [role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html) and attach the IAM to the instance that requires the specific permissions, like for example in the default `cookiecutter-django` setup, it's the `S3FullPermission` for writing the static files.

Then we can safely remove the following:

env.example: 
https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/env.example?#L18:L19

production.py:
https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/env.example?#L18:L19

sfdye:
Fixes #1453 

@pydanny @jayfk please have a review.


docs/deployment-with-docker.rst
```diff
@@ -42,6 +42,13 @@ You will probably also need to setup the Mail backend, for example by adding a `
 .. _sentry.io: https://sentry.io/welcome
 .. _Mailgun: https://mailgun.com
 
+Create AWS IAM Role for EC2 instance
+------------------------------------
+As a security best practice, we don't store `AWS_ACCESS_KEY_ID` AND `AWS_SECRET_ACCESS_KEY` on the server. In stead, in order to authorize Django to access your S3 bucket, you need to create an `IAM role`_ and `attach`_ it to the existing EC2 instance or create a new EC2 instance with that role. This role should assume a minimum permission of `AmazonS3FullAccess`.
```
browniebroke: "Instead"

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -94,8 +94,6 @@
 # See: http://django-storages.readthedocs.io/en/latest/index.html
 INSTALLED_APPS += ['storages', ]
 
-AWS_ACCESS_KEY_ID = env('DJANGO_AWS_ACCESS_KEY_ID')
-AWS_SECRET_ACCESS_KEY = env('DJANGO_AWS_SECRET_ACCESS_KEY')
```
browniebroke: This will break non-AWS deployment that use S3 to store files, e.g. on Heroku with Media files going to a S3 Bucket.

sfdye: Thanks for pointing out this. I am gonna go ahead and revert this.

sfdye:
@browniebroke Done. How about now?

wasabigeek:
Stumbled upon this while figuring out how to setup S3 - how would this work if I weren't deploying direct to AWS e.g. Heroku?

sfdye:
@wasabigeek You can set them as environment variables on Heroku.

wasabigeek:
@sfdye ah sorry, I thought the change removed the variables from `settings/production.py`, not `.env`.

I'd suggest to leave a comment in `.env` instead of removing the variables entirely, as it's a nice, centralised list of all environmental variables used in the project.

sfdye:
@wasabigeek If you deploy to heroku-like PaaS, then you can follow similar steps to set the environment variables:
https://cookiecutter-django.readthedocs.io/en/latest/deployment-on-heroku.html

To me it feels weird to leave variables in the `env.example` if I am not going to set them.

sfdye:
@browniebroke please review again.

sfdye:
@browniebroke It's good enough :smiley:

browniebroke:
Thank you @sfdye

PR #1496, opened 2018-02-07
Add Debian 9 (Stretch) os dependencies

ddiazpinto:
They are properly checked and all of them are the same than version 8, but due to it is necessary to have an specific file, here it is.


webyneter:
@ddiazpinto thanks! :)

PR #1502, opened 2018-02-09
Use the default docker from Travis

### LINKED ISSUE #1199: Upgrade docker and docker-compose deployed by Travis
What do you think of upgrading 

* `docker` deployed during CI to the latest [`v17.05.0-ce`](https://github.com/moby/moby/releases/tag/v17.05.0-ce) release?
* `docker-compose` deployed during CI to the latest [`1.14.0`](https://github.com/docker/compose/releases/tag/1.14.0) release?

By the way, I have already implemented it (and rolled back when realized that wasn't the problem back then) while working on #1128.

browniebroke:
Rather than installing a specific version and having to update it ourselves, we can now use the [default from Travis](https://docs.travis-ci.com/user/docker/). Current versions are:
- Docker 17.09-ce
- Docker compose 1.17.1

Fixes #1199

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,15 +13,6 @@ env:
     - TOX_ENV=py36
 
 before_install:
-  - sudo sh -c 'echo "deb https://apt.dockerproject.org/repo ubuntu-precise main" > /etc/apt/sources.list.d/docker.list'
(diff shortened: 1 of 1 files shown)


webyneter:
@browniebroke this one LGTM, however, I'd like for us to have a way of informing our users about the Docker and Docker Compose versions which are guaranteed to be supported (effectively, the ones pulled by Travis). Sure thing, we can't just copy-paste those from Travis environment, so, I guess, automating the process is the mission impossible for now.

PR #1513, opened 2018-02-16
Add atomic requests setting to production.py

hashimmm:
Since config/production.py sets a new value for `DATABASES['default']`, the `DATABASES['default']['ATOMIC_REQUESTS'] = True` setting from base gets overridden. So it's probably a good idea to add it back.


webyneter:
@hashimmm nice catch!

PR #1522, opened 2018-02-20
Remove AWS S3 header bytes workaround

sfdye:
#1374 introduced the new `AWS_S3_OBJECT_PARAMETERS ` header but did not remove the bytes workaround.

This PR does the same.


sfdye:
@browniebroke

PR #1544, opened 2018-03-04
update grunt-sass to latest version

ericgroom:
I was having issues installing grunt-sass running `npm install` which was failing on installing a dependency: 'node-sass'. I noticed that it was trying to install a really old version of node-sass so I tried installing the latest version of `grunt-sass` in an empty project which fixed the issue. I haven't done any extensive testing but the latest version of `grunt-sass` seems to work fine and all of the tox tests pass (although I'm not totally sure if they test grunt). I noticed some of the other dependencies in package.json are outdated, so hopefully this pull request can be the start of updating those if needed.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -89,6 +89,7 @@ Listed in alphabetical order.
   Dong Huynh               `@trungdong`_
   Emanuel Calso            `@bloodpet`_                  @bloodpet
   Eraldo Energy            `@eraldo`_
+  Eric Groom               `@ericgroom`_
   Eyad Al Sibai            `@eyadsibai`_
   Felipe Arruda            `@arruda`_
   Garry Cairns             `@garry-cairns`_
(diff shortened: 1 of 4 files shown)


webyneter:
@ericgroom thank you for the PR. As this is the first time you contribute would you mind adding yourself to `CONTRIBUTORS`?

ericgroom:
Done. Right now I'm looking into updating the other dependencies as `gulp-sass` is also outdated and some of them have even been deprecated (`gulp-util`)

ericgroom:
I think that about does it. For testing, I created a project for gulp and grunt from the new template and ran `py.test` on each with no issues as well as changed some bootstrap sass variables to make sure that worked. I mentioned in an earlier comment that `gulp-util` was deprecated but too many of the other deps depend on `gulp-util` and it seems like a recent change (end of 2017).

webyneter:
@ericgroom FWIW, none of our tests cover frontend.

PR #1557, opened 2018-03-07
Prettify generated project requirements

webyneter:


{{cookiecutter.project_slug}}/requirements/test.txt
```diff
@@ -1,18 +0,0 @@
-# Test dependencies go here.
```
jayfk: why remove them?

webyneter: What we call test environment is essentially a superset of a dev env with an extra few infrastructural dependencies (e.g. `pytest-sugar`).

jayfk: Okay. It would be better to do deletions like that in a single PR.

The current diff makes it super hard to see what's going on since every line has changed :D

webyneter: @jayfk yes, indeed.

PR #1615, opened 2018-04-19
Use Heroku's Release Phase for Migrations

mcastle:
Automatically run migrations on deployments to Heroku. Advantages include deployments are rolled-back if a migration fails, preventing broken applications due to failed migrations, no time between when application is released and database is migrated, and removes risk of forgetting to manually run migrations.

### DIFF
--- a/{{cookiecutter.project_slug}}/Procfile
+++ b/{{cookiecutter.project_slug}}/Procfile
@@ -1,3 +1,4 @@
+release: ./manage.py migrate
 web: gunicorn config.wsgi:application
 {% if cookiecutter.use_celery == "y" -%}
 worker: celery worker --app={{cookiecutter.project_slug}}.taskapp --loglevel=info


mcastle:
I don't think the Heroku deployment documentation needs to account for this, since the new code is only in the Procfile and doesn't require any extra configuration or Heroku commands to be run (it runs automatically after a typical git push to Heroku).

browniebroke:
Sorry, I wasn't clear, what I meant is that the documentation can be simplified, we don't need to run `heroku python manage.py migrate` anymore, therefore, we can remove that line from the documentation.

mcastle:
No prob, I should've caught that. Done! https://github.com/pydanny/cookiecutter-django/pull/1630

{{cookiecutter.project_slug}}/Procfile
```diff
@@ -1,3 +1,4 @@
+release: ./manage.py migrate
```
webyneter: Let's make it `release: manage.py migrate`, shall we? It's pretty much obvious what `manage.py` we refer to here anyway.

browniebroke:
FYI, we had to revert this, see #1646 for more details.

PR #1637, opened 2018-05-06
Migrate to Django2 path urls

### LINKED ISSUE #1519: Migrate from old url regex urls to new path urls
**Note: for support questions, please use the `cookiecutter-django` tag on stackoverflow**. This repository's issues are reserved for feature requests and bug reports. If you need quick professional paid support for your project, contact support@cookiecutter.io.

* **I'm submitting a ... **
  - [ ] bug report
  - [x] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?**

Django versions pre 2 used django.conf.urls.url() instances within a projects urlpatterns. The "preferred" way of doing this is to use django.urls.path(), which does not require the use of regex. For any paths which require the use of regex, we can use django.urls.re_path(). 

* **What is the current behavior?**

The old django.conf.urls.url() still works, but is mapped to django.urls.re_path(), without any warning as it is not yet been marked for depreciation

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** 

* **What is the expected behavior?**

The new system should use django.urls.path(), or django.urls.re_path() when regex is required

* **What is the motivation / use case for changing the behavior?**

The new path syntax is cleaner, simpler and more readable than the previous regex version. 

* **Please tell us about your environment:**

cookiecutter-django with django2

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax
https://docs.djangoproject.com/en/2.0/ref/urls/

saschalalala:
Closes #1519.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -216,8 +216,8 @@ EMAIL_BACKEND = env('DJANGO_EMAIL_BACKEND', default='django.core.mail.backends.s
 
 # ADMIN
 # ------------------------------------------------------------------------------
-# Django Admin URL regex.
-ADMIN_URL = r'^admin/'
+# Django Admin URL.
+ADMIN_URL = 'admin/'
 # https://docs.djangoproject.com/en/dev/ref/settings/#admins
 ADMINS = [
     ("""{{cookiecutter.author_name}}""", '{{cookiecutter.email}}'),
--- a/{{cookiecutter.project_slug}}/config/urls.py
+++ b/{{cookiecutter.project_slug}}/config/urls.py
@@ -1,25 +1,25 @@
 from django.conf import settings
-from django.conf.urls import include, url
+from django.urls import include, path
 from django.conf.urls.static import static
 from django.contrib import admin
 from django.views.generic import TemplateView
 from django.views import defaults as default_views
 
 urlpatterns = [
-    url(r"^$", TemplateView.as_view(template_name="pages/home.html"), name="home"),
-    url(
-        r"^about/$",
+    path("", TemplateView.as_view(template_name="pages/home.html"), name="home"),
+    path(
+        "about/",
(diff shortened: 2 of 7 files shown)


{{cookiecutter.project_slug}}/config/urls.py
```diff
@@ -29,24 +29,24 @@
...
-            r"^400/$",
+        path(
+            "400/",
             default_views.bad_request,
             kwargs={"exception": Exception("Bad Request!")},
         ),
-        url(
-            r"^403/$",
+        path(
+            r"403/",
```
browniebroke: This `r"..."` string should be changed to a regular one, shouldn't it?

{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -216,8 +216,8 @@
 
 # ADMIN
 # ------------------------------------------------------------------------------
-# Django Admin URL regex.
-ADMIN_URL = r'^admin/'
+# Django Admin URL.
+ADMIN_URL = 'admin/'
```
browniebroke: This pattern should be updated in -at least- 2 other places:
- In post gen hooks, for Docker: 
  https://github.com/pydanny/cookiecutter-django/blob/661fc3e449c779ad3fa2bad8964b6b7082f88ca3/hooks/post_gen_project.py#L159-L168
- In the documentation, for Heroku: http://cookiecutter-django.readthedocs.io/en/latest/deployment-on-heroku.html

webyneter:
@saschalalala any news of this one?

saschalalala:
@webyneter Yeah, just did the things @browniebroke mentioned and looked around in basically all other files if I missed something more but just found the development default url in `settings.rst` which I also adjusted to stay consistent.

Issue #1016, opened 2017-02-03
django_extensions app is imported in local settings only


webyneter:
I noticed the `django_extensions` app is being imported in local configuration only; provided I want to use its [`TimeStampedModel`](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/models.py#L14) to build my models upon, shouldn't the app be present in production configuration as well i.e installed in `common.py`?

luzfcb:
@webyneter django-extensions is only used in development for security reasons. See: https://github.com/pydanny/cookiecutter-django/issues/417#issuecomment-158452987

said that, django-model-utils provide a very similar `TimeStampedModel` model class https://github.com/carljm/django-model-utils/blob/master/model_utils/models.py#L31-L41

and other usefull features

webyneter:
@luzfcb, thanks for the exhaustive reply!

PR #1644, opened 2018-05-12
making django_extensions as a base requirement and base installed apps.

scripterkaran:
Stuff breaks when you are giving flexibility of using "django_extensions" in local environment but not in production env. 

For example using docker system (uses base+local requirements ): 

`from django_extensions.db.models import TimeStampedModel`  would work.

but using docker in production mode will use (base + production requirements)
But production requirements file does not have "django_extensions".

Also typically I don't see why django_extensions should not be allowed in production environment. 
Am I missing anything there?

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -73,6 +73,7 @@ THIRD_PARTY_APPS = [
     'allauth.account',
     'allauth.socialaccount',
     'rest_framework',
+    'django_extensions'
 ]
 LOCAL_APPS = [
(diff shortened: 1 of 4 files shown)


webyneter:
@scripterkaran here's why [we discourage the use of `django_extensions` in production](https://github.com/pydanny/cookiecutter-django/issues/1016). Use the `django-model-utils` `TimeStampedModel` instead.

luzfcb:
in addition, i have used `TimeStampedModel` in conjunction with `CurrentUserField` of [django-cuser](https://github.com/Alir3z4/django-cuser), without problems so far.

Issue #1646, opened 2018-05-13
Don't migrate during Heroku release phase


webyneter:
[Release Phase, Design considerations](https://devcenter.heroku.com/articles/release-phase#design-considerations):
> If you perform a database migration, you should always use transactions. A transaction ensures that all migration operations are successful before committing changes to the database. This minimizes the possibility of a failed partial migration during release phase. If a database migration fails during release phase (i.e., the migration command exits with a non-zero status), your new release will not be deployed. If transactions were not used, this can leave your database in a partially migrated state. **We suggest using heroku run, rather than release phase, to correct your schema/data**.

browniebroke:
Yea, also I think it's broken when deploying the app the first time.

mcastle:
Good point, esp for projects using neither PostgreSQL nor SQLite (django migrations for these db's run inside a transaction by default... https://docs.djangoproject.com/en/2.0/howto/writing-migrations/#non-atomic-migrations).

PR #1651, opened 2018-05-19
use python command

hadeelsharaf:
use python to fix heroku deployment issue


webyneter:
Firstly, [#1646](https://github.com/pydanny/cookiecutter-django/issues/1646). Secondly, we won't be switching to [mdillon/postgis](https://hub.docker.com/r/mdillon/postgis/) by no means as having PostGIS around is an unwanted maintenance burden. Thirdly, I don't see any reason for us to revert our `django` image back to full `python` base.

hadeelsharaf:
Thanks for your response.

I was adding postgis to my project I thought it might be helpful to add it.

PR #1655, opened 2018-05-21
Integrate Flower with Docker Compose setup

webyneter:
Thought it'd be nice to have it.

* [x] Integrate Flower Celery monitoring tool with our Docker Compose setup locally
* [x] Cover production Compose configuration as well
* [Auth](http://flower.readthedocs.io/en/latest/auth.html)
    * [x] local
    * [x] production
* [x] Update the docs

@jayfk, @pydanny, @browniebroke, @luzfcb, thoughts?


{{cookiecutter.project_slug}}/requirements/local.txt
```diff
@@ -8,6 +8,9 @@ psycopg2==2.7.4 --no-binary psycopg2  # https://github.com/psycopg/psycopg2
 {%- else %}
 psycopg2-binary==2.7.4  # https://github.com/psycopg/psycopg2
 {%- endif %}
+{%- if cookiecutter.use_docker == 'y' and cookiecutter.use_celery == 'y' %}
+flower==0.9.2  # https://github.com/mher/flower
```
browniebroke: Have you tested this? Incidentally, I tried to upgrade our Flower instance at work today and came across an incompatibility with Tornado v5+. It was fixed [recently in a PR](https://github.com/mher/flower/pull/793), which is merged, but not yet released.

webyneter: Just tested this out locally with a fresh project, however, I've only examined basic use cases like opening up some of the panel pages etc.

webyneter: @browniebroke as I understand we haven't got many alternatives out there anyway, unfortunatelly (at least, according to [the Celery monitoring user guide](http://docs.celeryproject.org/en/latest/userguide/monitoring.html))...

webyneter: Until the release the only option for our users would be to install the latest dev version then.

browniebroke: > Just tested this out locally with a fresh project, however, I've only examined basic use cases like opening up some of the panel pages etc.

I think it was failing when trying to open Flower's home page, so a pretty basic test.

> as I understand we haven't got many alternatives out there anyway, unfortunatelly

Apologies, I realise now that my comment wasn't very obvious. I use Flower and I agree that there are not many alternatives. My comment was more directed at the version of Flower. 

> Until the release the only option for our users would be to install the latest dev version then.

I don't think we depend on Tornado, so alternatively we could pin its version to the latest v4 with a `# pyup >=4,<5`? Or wait for a newer version of Flower...

webyneter: > I think it was failing when trying to open Flower's home page, so a pretty basic test.

All the pages I've seen has been working OK, including the one for Redis broker, queues etc. I also tried it out on one of my previous projects, and hasn't spotted any problems either, at least, within the trivial scope of tasks I had at hand.

webyneter: > I don't think we depend on Tornado, so alternatively we could pin its version to the latest v4 with a # pyup >=4,<5? Or wait for a newer version of Flower...

No, we don't, and I see no reason for us to include that as a dependency only to pin its version, or, am I missing something about your statement?

browniebroke: Correct, it's not needed if everything works, only to prevent getting the latest _if_ it were to cause any trouble.

Not sure why I had the problem then, maybe related to something else...

webyneter: > Correct, it's not needed if everything works, only to prevent getting the latest if it were to cause any trouble.

Does `pyup` support such semantics? :) If yes, I'm eager to take a look.

browniebroke:
Looks good, and to answer your question (?) I am in favour of getting it into the production config too.

{{cookiecutter.project_slug}}/requirements/local.txt
browniebroke: I don't think so, I meant to explicitly declare the indirect dependency, and adding a pyup filter to it.

https://github.com/founders4schools/celery-flower-heroku/blob/master/requirements.txt

webyneter: Oh, I see. Nah, I'm feeling bad about adding any dependencies we don't actually depend on...

webyneter: On the other hand, it'd be easier to leave a TODO/NOTE whatever in the comment following the declaration, encompassing an explanatory link like the one from Flower tracker you mentioned above.

luzfcb:
perhaps it would be good, enable the authentication by default in flower .
depending on how the tasks were created, they can leak sensitive information.
http://flower.readthedocs.io/en/latest/auth.html

{{cookiecutter.project_slug}}/compose/local/django/celery/flower/start
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
+
+
+celery flower -A {{cookiecutter.project_slug}}.taskapp worker --broker="${CELERY_BROKER_URL}"
```
sfdye: The `worker` is not needed here?

Agree adding a basic auth for production.

webyneter: Sure, just copy-pasted existing `celeryd` commands :)

webyneter:
@jayfk, @pydanny, @browniebroke, @luzfcb, thoughts?

webyneter:
@browniebroke since local and production setups are essentially the same I thought  referencing the local Flower docs section would be enough. As for Flower UI usage, I think it's best if our users merely follow the official links and see the original writing for themselves.

luzfcb:
LGTM

docs/deployment-with-docker.rst
```diff
@@ -24,7 +24,10 @@ Before you begin, check out the ``production.yml`` file in the root of this proj
 Provided you have opted for Celery (via setting ``use_celery`` to ``y``) there are two more services:
```
sfdye: three services? :)

webyneter: sure :)

PR #1664, opened 2018-05-25
Remove project doc files likely to remain unused

webyneter:
* ? Also, remove `./docs/pycharm` as well: those are outdated -- I'd rather have PyCharm setup described in Cookicutter Django docs, under `IDE Setup/PyCharm` or smth like that.


{{cookiecutter.project_slug}}/docs/docker_ec2.rst
```diff
@@ -1,186 +0,0 @@
...
-    }
-    server {
-        location / {
-            proxy_pass http://localhost;
-        }
-    }
-
-Running `sudo docker-compose -f production.yml build webserver` will build your server container. Running `sudo docker-compose -f production.yml up` will now expose your application directly on `localhost` (no need to specify the port number).
-
-Building and running your app on EC2
```
browniebroke: I think this part is not described well enough in our current template docs, we should probably add an equivalent & updated section over there.

browniebroke:
@webyneter this hasn't been touched in months, any reason to not merge it? Yes, we could (re)move or move the PyCharm docs, but as it stands, I'd say it's an improvement anyway, isn't it?

webyneter:
@browniebroke, yeah, pretty much. Feel free to merge at will.

PR #1676, opened 2018-06-06
Add celery prefix for configuration

apirobot:
My tasks haven't been worked until I added celery namespace. Spent about 30 minutes on this :(


webyneter:
@apirobot did you also prepend all your Celery settings with `CELERY_`? Last time I checked on one of my production projects that was enough, no need for explicit `namespace='CELERY'`...

PR #1446, opened 2018-01-12
Upgrade celery to 4.2

sfdye:
Closes #1249.

According to the official [upgrading guide](http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#step-4-upgrade-to-celery-4-0
) from 3.X to 4.X, the only thing that could break the existing code for `cookicutter-django` seems to be the new lower case setting names. However this change is backward-compatible, so until celery officially depreciates upper case settings name, we are good.

Therefore this update would be very simple, just one incompatible settings name change. I've already bumped celery from 3.1.25 to 4.1 for one of my company's projects in production and so far I don't see any issues arise from it.

@luzfcb I haven't encounter the timezone issues you mentioned, so if you can reproduce it please let me know.

PR #1679, opened 2018-06-09
Extend & enhance Celery configuration

webyneter:
[Celery **4.2**](https://github.com/pydanny/cookiecutter-django/pull/1446) is a [necessary](https://github.com/celery/celery/issues/4341) prerequisite for this PR.

* [x] `CELERYD_TASK_TIME_LIMIT`
* [x] `CELERYD_TASK_SOFT_TIME_LIMIT`
* [x] Rename `CELERY_ALWAYS_EAGER` to `CELERY_TASK_ALWAYS_EAGER`
* [x] Set `CELERY_TASK_EAGER_PROPAGATES = True` as by @browniebroke's suggestion
* [x] `CELERY_RESULT_BACKEND = CELERY_BROKER_URL` always
* [x] `CELERY_TIMEZONE = TIME_ZONE` given `USE_TZ` as by @browniebroke [suggestion](https://github.com/pydanny/cookiecutter-django/pull/1446#pullrequestreview-127518828)
* Flower's [soon to be integrated](https://github.com/pydanny/cookiecutter-django/pull/1655) as well
* ? (IMO, nice to have it by default but we also can leave a # TODO: ... in the `celeryworker` `start` script) [-Ofair](http://docs.celeryproject.org/en/latest/reference/celery.bin.worker.html#cmdoption-celery-worker-o)
* ? (shouldn't be enabled by default, IMO, yet we could leave a "best practice" comment somewhere in the code base) [custom retry delay](http://docs.celeryproject.org/en/latest/userguide/tasks.html#using-a-custom-retry-delay) + [Automatic retry for known exceptions](http://docs.celeryproject.org/en/latest/userguide/tasks.html#automatic-retry-for-known-exceptions) using [exponential backoffs](http://docs.celeryproject.org/en/latest/userguide/tasks.html#Task.retry_backoff)

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -230,12 +230,9 @@ MANAGERS = ADMINS
 # ------------------------------------------------------------------------------
 INSTALLED_APPS += ['{{cookiecutter.project_slug}}.taskapp.celery.CeleryAppConfig']
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-broker_url
-CELERY_BROKER_URL = env('CELERY_BROKER_URL', default='django://')
+CELERY_BROKER_URL = env('CELERY_BROKER_URL')
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-result_backend
-if CELERY_BROKER_URL == 'django://':
-    CELERY_RESULT_BACKEND = 'redis://'
-else:
-    CELERY_RESULT_BACKEND = CELERY_BROKER_URL
+CELERY_RESULT_BACKEND = CELERY_BROKER_URL
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-accept_content
 CELERY_ACCEPT_CONTENT = ['json']
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_serializer
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -244,10 +244,10 @@ CELERY_TASK_SERIALIZER = 'json'
 CELERY_RESULT_SERIALIZER = 'json'
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-time-limit
 # TODO: set to whatever value is adequate in your circumstances
-CELERY_TASK_TIME_LIMIT = 5 * 60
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -229,19 +229,25 @@
...
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_serializer
 CELERY_TASK_SERIALIZER = 'json'
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-result_serializer
 CELERY_RESULT_SERIALIZER = 'json'
+# http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-time-limit
+# TODO: set to whatever value is adequate in your circumstances
+CELERYD_TASK_TIME_LIMIT = 5 * 60
+# http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-soft-time-limit
+# TODO: set to whatever value is adequate in your circumstances
+CELERYD_TASK_SOFT_TIME_LIMIT = 60
```
browniebroke: Why are you prefixing these 2 with `CELERYD_` instead of `CELERY_`? Is it to "disable" the setting by default? I'm not sure we should decide for these numbers, they seem a bit specific to each project and might be harmful in some cases. I'm a bit

webyneter: Beacause these are [the names Celery expects...](https://ibb.co/ggUHAJ)

webyneter: Was quite a surprise for me as well.

browniebroke: Good spot on the name 👍

webyneter:
@browniebroke what do you think about `-0fair` and exponential backoffs: should we explicitly mention those in the docs or something?

browniebroke:
IMO it's important to avoid duplicating too much of Celery's documentation in cookiecutter-django, if possible, but we could highlight a few advanced interesting options, yes, and point to the Celery docs. The [optimizing](http://docs.celeryproject.org/en/latest/userguide/optimizing.html) section has some good options, for instance.

For now, this is good enough to upgrade Celery to v4.2, and I'd like to close the few open Celery-related PRs we have currently opened. Maybe follow up with another pull request for the documentation?

webyneter:
@browniebroke I agree, pointing out a few crucial aspects like typical optimization techniques should be more than enough. I also share your eagerness to finish up with Celery 4-related edits: it has been quite a while, indeed.

browniebroke:
I think we should merge this one too now that #1446 is merged

PR #1682, opened 2018-06-15
Remove entry from .gitignore causing compose/local to match

browniebroke:
## Description

Not sure why this line was part of the `.gitignore`, but it causes the folder `{{cookiecutter.project_slug}}/compose/local` to match. This may cause partial commits when its content gets added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -215,7 +215,6 @@ tags
 [Ii]nclude
 [Ll]ib
 [Ll]ib64
-[Ll]ocal
 [Ss]cripts
 pyvenv.cfg
 pip-selfcheck.json


webyneter:
Because it's part of one of the "standard" `.gitignore` template, one of those you can find in the corresponding GtiHub repo. Some time ago I thought it'd be a great idea to unify `.gitignore` so as to offer industry standard ignore patterns depending on user's circumstances -- as it turns out a few of them tend to interfere with our own layout inadvertently re-raising the concern of an appropriate `.gitignore` structure fitting with any setup generically (not sure if that'd be possible though).
Feel free to remove this and other lines from `.gitignore` whenever you spot such undesired behaviors; I do this myself occasionally too, and I believe, removing `[Ll]ocal` from the template has been on my Cookiecutter Django todo list for a while now :)

webyneter:
N.B. In practice, generic `.gitignore` snippets appear to be evil to a greater extent than bless.

browniebroke:
Ok thanks for clearing my concerns @webyneter

PR #1687, opened 2018-06-16
Add a node image to run Gulp when selecting it with Docker

### LINKED ISSUE #1762: Selected Gulp and CSS compilation. `npm run dev` does not run automatically when docker-compose build+up is invoked
I have selected CSS compilation to be enabled and have chosen Gulp to be used as a task runner.  
When `docker-compose up` is  invoked auto compilation of css does not take place. I have to manually install nodejs, npm, gulp, compass to get the whole thing working. 

I also had to run npm run dev manually.

It kind of feels like a leaky abstraction, when some setup and commands work automatically, while others don't.

The expectation is that,  because of the choices made while setting up the project, `npm run dev` command will be automatically started, when docker-compose is invoked.

It also appears that the DockerFile does not install nodejs, npm or gulp in spite of choosing Gulp as task runner.

It should be trivial to incorporate.

The gulp task runner already has `python manage.py runserver` as one of it's options. So all the start script has to do is run `npm run dev` instead of `python manage.py runserver_plus`. (The gulpfile.js should be changed to use runserver_plus instead of runserver).

One way to incorporate this would be to have two Dockerfiles, one where gulp and bootstrap compilation is enabled and one where gulp and bootstrap compilation is disabled.

I am new to the cookiecutter-django community. So maybe I am missing something obvious. If so, I would be very interested in knowing the reasons behind not incorporating the above steps.

If there is interest in the community, I would be happy to implement the solution after discussing and deciding on the right way to do the same.

browniebroke:

[//]: # (Before you proceed:)

## Description

Following up on @webyneter attempt in #1205, which is now getting outdated, I've tried to make Gulp task runner work with Docker. There is no documentation yet, but this seems to work locally with the custom bootstrap compilation...

- [x] Add a node image for local developement
- [x] Proxy the django image rather than localhost in Browser sync, pass header to keep hostname
- [x] Don't call the runserver command from Gulp, let docker-compose handle
- [x] Update package.json & gulpfile.js templates to reduce the number of unwanted empty lines
- [x] Use [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) in production to make sure the static assets are produced
- [x] Update documentation
- [x] Verify that the previous issue with static assets missing from production isn't there 

## Rationale

Currently the static build isn't working nicely with Docker, extra manual setup is required

Fixes #1762


webyneter:
@browniebroke in a week or so I'm about to ship a PR simplifying our Docker infrastructure, introducing parametrized service `Dockerfile`s and Docker Compose configs so we should expect a few merge conflicts here afterwards. What's more, I recall my `gulpfile.js` was a complete re-write of the existing one instead, but that's about everything I can remember now. Thank you for taking time to look into it, it's been more than a year now, I think :)

nitishd:
I just tested this out by merging on master and it still works. Thanks @browniebroke!

drewmccal:
> I just tested this out by merging on master and it still works. Thanks @browniebroke!

Hey, were you getting any compile errors with npm.  Although the image builds, and I'm able to get sass and autoreloads to work correctly, I'm concerned about these errors:
```
⚠ spawn /app/node_modules/gifsicle/vendor/gifsicle ENOENT
  ⚠ gifsicle pre-build test failed
  ℹ compiling from source
  ✖ Error: autoreconf -ivf && ./configure --disable-gifview --disable-gifdiff --prefix="/app/node_modules/gifsicle/vendor" --bindir="/app/node_modules/gifsicle/vendor" && make install
Command failed: ./configure --disable-gifview --disable-gifdiff --prefix="/app/node_modules/gifsicle/vendor" --bindir="/app/node_modules/gifsicle/vendor"
configure: error: in `/app/node_modules/gifsicle/673732cb-87f9-4235-8eae-18cf780e2c8e':
configure: error: C compiler cannot create executables
See `config.log' for more details

    at ChildProcess.exithandler (child_process.js:281:12)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at maybeClose (internal/child_process.js:915:16)
    at Socket.stream.socket.on (internal/child_process.js:336:11)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at Pipe._handle.close [as _onclose] (net.js:561:12)

> jpegtran-bin@3.2.0 postinstall /app/node_modules/jpegtran-bin
> node lib/install.js

  ⚠ spawn /app/node_modules/jpegtran-bin/vendor/jpegtran ENOENT
  ⚠ jpegtran pre-build test failed
  ℹ compiling from source
  ✖ Error: touch configure.ac aclocal.m4 configure Makefile.am Makefile.in && ./configure --disable-shared --prefix="/app/node_modules/jpegtran-bin/vendor" --bindir="/app/node_modules/jpegtran-bin/vendor" && make install
Command failed: ./configure --disable-shared --prefix="/app/node_modules/jpegtran-bin/vendor" --bindir="/app/node_modules/jpegtran-bin/vendor"
configure: error: in `/app/node_modules/jpegtran-bin/989a47a5-1225-4eea-a986-718275b88954':
configure: error: C compiler cannot create executables
See `config.log' for more details

    at ChildProcess.exithandler (child_process.js:281:12)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at maybeClose (internal/child_process.js:915:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)

> optipng-bin@3.1.4 postinstall /app/node_modules/optipng-bin
> node lib/install.js

  ⚠ spawn /app/node_modules/optipng-bin/vendor/optipng ENOENT
  ⚠ optipng pre-build test failed
  ℹ compiling from source
  ✖ Error: ./configure --with-system-zlib --prefix="/app/node_modules/optipng-bin/vendor" --bindir="/app/node_modules/optipng-bin/vendor" && make install
Command failed: ./configure --with-system-zlib --prefix="/app/node_modules/optipng-bin/vendor" --bindir="/app/node_modules/optipng-bin/vendor"

    at ChildProcess.exithandler (child_process.js:281:12)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at maybeClose (internal/child_process.js:915:16)
    at Socket.stream.socket.on (internal/child_process.js:336:11)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at Pipe._handle.close [as _onclose] (net.js:561:12)
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN medusa@0.1.0 No description
npm WARN medusa@0.1.0 No repository field.
npm WARN medusa@0.1.0 No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.7 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.7: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

added 1256 packages from 656 contributors and audited 13971 packages in 37.226s
found 11 vulnerabilities (1 low, 6 moderate, 4 high)
  run `npm audit fix` to fix them, or `npm audit` for details
npm WARN using --force I sure hope you know what you are doing.

```

nitishd:
> Hey, were you getting any compile errors with npm? Although the image builds, and I'm able to get sass and autoreloads to work correctly, I'm concerned about these errors:

I don't remember. The logs show that for some reason it is unable to install the binaries and instead trying to build them. But the alpine image does not have the build tools, so we can add it like this:
https://github.com/nitishd/cookiecutter-django/commit/4fe333c3e48cbf6300b5012e0a56f4ebf7dd6151

It should then be able to build those plugins without errors.

drewmccal:
@nitishd Thanks for your help.  I also had to add a couple other packages to get jpegtran to compile and behave, according to this [post](https://gist.github.com/welcoMattic/9f0991fa81a80096a3877ee42562c504):
```
RUN apk --no-cache add \
    pkgconfig \
    autoconf \
    automake \
    libtool \
    build-base \
    nasm \
    zlib-dev \
    file \
    build-base \
    nasm \
    musl \
    libpng-dev
```
Also, is there any easy fix to migrate away from gulp-util?
```
npm WARN deprecated gulp-util@3.0.8: gulp-util is deprecated - replace it, following the guidelines at https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5
npm WARN deprecated browserslist@1.7.7: Browserslist 2 could fail on reading Browserslist >3.0 config used in other tools.
npm WARN deprecated graceful-fs@3.0.11: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated graceful-fs@1.2.3: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
```

browniebroke:
@drewmccal @nitishd thank you for giving this a try. I think there are a couple of things to add to the image. I will try to find some time to add these. There is not much to be done about the NPM warning apart maybe from upgrading our actual libs, but Gulp isn't as active as in the past... 

Given that it's used only as an interim step to build some assets, it might make sense to use a non-alpine image? One notable potential disadvantage I'm aware of is that it would share fewer layers with the other alpine based images. Thoughts?

nitishd:
@browniebroke Yes, I tested out using `node:8-jessie-slim` ~180 MB image vs Alpine's ~60 MB image. The Jessie image shares only 2 of the 6 layers. But considering the excessive build times and future edge cases with alpine I would avoid it for node. With Jessie's image we do not have to add anything to it.

While I was at it, I also updated Gulp to version 4 https://github.com/nitishd/cookiecutter-django/commit/6a4f4e17eaa9b292994f208c9ad915125d0746f9, this removes all the deprecated warnings, vulnerabilities of old packages, move off of `gulp-util` (@drewmccal ).

drewmccal:
@nitishd Fantastic, thank you

jaddison:
What's the status on getting gulp actually working out of the box with `cookiecutter-django`? I think most people would simply be happy with documentation updates that instruct on how to do it manually within docker-compose/dockerfile files.

thelittlebug:
it seems to work if you clone the work from @nitishd (i think its based on the work of @browniebroke ) like so:
```
cookiecutter https://github.com/nitishd/cookiecutter-django
cd <yourproject>
docker-compose -f local.yml build 
docker-compose -f local.yml up
```

the npm container gets started in the background, together with browsersync. so you have to visit browsersync (port 3000) and not django (port 8000): `http://127.0.0.1:3000`

to test the browsersync open, for example, this file: `<your project>/<your project>/static/css/project.css` and try to change the font size of the html tag from
```
html {
  font-family: sans-serif;
  line-height: 1.15;
  -webkit-text-size-adjust: 100%;
  -webkit-tap-highlight-color: rgba(0, 0, 0, 0); }
```
to
```
html {
  font-size: 6px;
  font-family: sans-serif;
  line-height: 1.15;
  -webkit-text-size-adjust: 100%;
  -webkit-tap-highlight-color: rgba(0, 0, 0, 0); }
```

and save the file -> tada, a short moment later your site gets reloaded :)

browniebroke:
Generated this empty project with the current version of this PR: https://github.com/browniebroke/djcc-docker-gulp

{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
```diff
@@ -1,5 +1,5 @@
 {% if cookiecutter.js_task_runner == 'Gulp' -%}
-FROM node:8-alpine as client-builder
+FROM node:8-jessie-slim as client-builder
```
jaddison: Shouldn't this be `node:10-stretch-slim`?

- [`stretch` is latest stable Debian release](https://www.debian.org/releases/)
- [`10` is the current Node LTS](https://nodejs.org/en/download/releases/)

{{cookiecutter.project_slug}}/gulpfile.js
```diff
@@ -127,7 +127,13 @@ function initBrowserSync() {
         `${paths.js}/*.js`,
         `${paths.templates}/*.html`
       ], {
+        {%- if cookiecutter.use_docker == 'n' %}
         proxy: "localhost:8000"
+        {% else %}
+        proxy:  "django:8000",
+        // Doesn't work from inside the Docker image
```
jaddison: What does this mean, out of curiosity? What about this doesn't work, exactly?

{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
browniebroke: Yes, good point. I haven't tested yet but it should be easy to do on the generated project.

{{cookiecutter.project_slug}}/gulpfile.js
browniebroke: By default, browser sync tries to open the default browser at the location it runs. It works great when it's running directly on the same machine, but doesn't work from inside a container.

Will elaborate this comment.

jaddison:
@browniebroke it appears that this cannot currently be merged cleanly

zapotocnylubos:
Hi @browniebroke, I have tried to implement your solution on clean project.

**Environment:**
docker, Grunt, MailHog, web accessed via browsersync (localhost:3000)

**Issue:**
After sign up, while being on localhost:3000 (browsersync), I have received in MailHog (localhost:8025) confirmation email, which contained url for account verification. But url address is 
`http://django:8000/accounts/confirm-email/MQ:1h6XzF:ounuN25PEDrsSRFlQ9YYFocaTpQ/`
which is not valid address, but it is product of proxying.

**Question:**
Can it be somehow solved (configured)? 
Be able to view web via browsersync and be able to send emails correctly?

Thanks

browniebroke:
Thanks both for your comments. I'm not sure where the hostname is coming from, to be honest. 

For the wrong host in the email, does it look like [this template](https://github.com/pennersr/django-allauth/blob/master/allauth/templates/account/email/email_confirmation_message.txt) was used? If yes, then it looks like it's using [Django's `get_current_site`](https://github.com/django/django/blob/0b70985f42da7a8eb2e206e6780681b7849564ef/django/contrib/sites/shortcuts.py#L4)...

browniebroke:
My bad, the URL is actually built here:

https://github.com/pennersr/django-allauth/blob/9614b862932a63eb1242a31e601cd6075eff45b7/allauth/account/adapter.py#L445

And that calls Django's [Request.build_absolute_uri() method](https://github.com/django/django/blob/77d25dbd0f20d6a907c805ffae8aaadd87edbacf/django/http/request.py#L168).

I don't see any options there, unfortunately.

zapotocnylubos:
I have found provisional working solution. 
I don't know much about this, but for me, when this is only for development, I don't mind too much.

**My setup:**
- Mac
- running browsersync on binded port 3000 - so my access is on `http://localhost:3000`
```
ports:
    - "3000:3000"
```
(local.yml > services > node)
- localhost is in allowed host in Django local settings (**removed** "django")
`ALLOWED_HOSTS = ["localhost", "0.0.0.0", "127.0.0.1"]`
- modified proxy settings in **gulpfile.js**
```
// proxy:  "django:8000",

// https://www.browsersync.io/docs/options/#option-proxy
proxy: {
  target: "django:8000",
  proxyReq: [
      function(proxyReq) {
          proxyReq.setHeader('Host', 'localhost:3000');
      }
  ]
},
```
this basically rewrites "Host" header value to "localhost:3000".

Now confirmation email generates correctly:
`To confirm this is correct, go to http://localhost:3000/accounts/confirm-email/Mg:1h6j8e:aBDXFUlXhTRQSxXd7hJa64pPUnQ/`

Hope my solution helps.

Could this bring any other difficulties?

browniebroke:
Amazing! I'll update tomorrow 👍

browniebroke:
I noticed that Django debug toolbar wasn't working when proxying through Browsersync. It looks like this new change might fix it:

https://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html#django-debug-toolbar

browniebroke:
@zapotocnylubos I've updated with the changes you suggested, let me know if I missed anything. If no other issues are found and this work, I suggest we merge it in the coming few days.

zapotocnylubos:
@browniebroke I have tried to setup some code to see if debug toolbar is working when proxied through Browsersync in Docker. 
**Current solution don't work for me.**
I have found an open issue regarding this problem - https://github.com/jazzband/django-debug-toolbar/issues/937

But to me, it is a lot easier to enable debug toolbar on every request in development environment.
So instead implementing arpMiddleware mentioned in that issue, I have changed default config for debug toolbar from this
```
DEBUG_TOOLBAR_CONFIG = {
    "DISABLE_PANELS": ["debug_toolbar.panels.redirects.RedirectsPanel"],
    "SHOW_TEMPLATE_CONTEXT": True
}
```
to this
```
DEBUG_TOOLBAR_CONFIG = {
    "DISABLE_PANELS": ["debug_toolbar.panels.redirects.RedirectsPanel"],
    "SHOW_TEMPLATE_CONTEXT": True,
    "SHOW_TOOLBAR_CALLBACK": lambda request: True
}
```

zapotocnylubos:
@browniebroke I was also thinking about if this solution provides support for local domain.
For example my friends on Windows have in hosts file mapping like this
docker.local -> {DOCKER_MACHINE_IP}

I am also creating multitenant system, in which I have multiple local domains, but cannot use any of them, because generated email will redirect to "localhost:3000", not "domain01.local:3000"

I haven't find solution for dynamic host assignment in Browsersync yet.

browniebroke:
Thanks for the feedback @zapotocnylubos. The Django Debug toolbar is a minor issue IMO, I can get it again by going to `localhost:8000`. I suggest trying to tackle this problem after the current PR is merged.

I'm not sure how common is the use case you describe in the latest comment. Is it related to this pull request or is it a problem you also have without Browsersync/Gulp? If the latter, it's best to open a separate issue on Github.

Thoughts?

zapotocnylubos:
Yeah, I would merge this PR, because it serves its purpouse.
For most people this will be functional.


But this brings up new issue because that proxy. But leave that for another time.
If you are using local domain (like docker.local or example-project.local,..) to access localhost, you cannot use Browsersync server (port 3000) but you need to go directly to Django server (port 8000).

zapotocnylubos:
@browniebroke 
I have solved my problem. Have a look at this code
```
proxy: {
  target: 'django:8000',
  proxyReq: [
    function (proxyReq, req) {
      // Assign proxy "host" header same as current request at Browsersync server
      proxyReq.setHeader('host', req.headers.host)
    }
  ]
},
```
I have added dynamic host assignment by accessed host.
So if you access Browsewrsync from **"localhost:3000"**, generated email contains **"localhost:3000"**.
If you acces from **"sub.domain.local:3000"** for example, generated email contains **"sub.domain.local:3000"**.

Issue #1691, opened 2018-06-21
Redis configuration in local django not needed (using Docker)


demestav:
In local environment, Redis is installed [only if you choose to use Celery](https://github.com/pydanny/cookiecutter-django/blob/5309fbf5e7135087f61f58bb2a940d92596e371b/%7B%7Bcookiecutter.project_slug%7D%7D/local.yml#L48). However, the [`.envs/.local/.django` file defines the Redis URL](https://github.com/pydanny/cookiecutter-django/blob/5309fbf5e7135087f61f58bb2a940d92596e371b/%7B%7Bcookiecutter.project_slug%7D%7D/.envs/.local/.django#L7). Is that referring to something else I am missing, or could it be removed?

luzfcb:
@demestav Redis is also used for django caching
https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L25

webyneter:
@demestav Celery.

webyneter:
Oh, @luzfcb was ahead on this :)

demestav:
@luzfcb @webyneter I see. However, you are referring to production environment, while I was referring to the [local environment](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/local.py) where (maybe I am wrong?) the cache is memory based.

webyneter:
@demestav oh, sure, `.django`'s `REDIS_URL` is supposed to be a conditional setting like the rest of Celery stack -- feel free to send a PR fixing this :)

demestav:
@webyneter Sure!

demestav:
Is [this line](https://github.com/pydanny/cookiecutter-django/blob/ee2bb3f2efef6add9a598597354ef5b3d3541f0b/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/django/entrypoint#L9) required in entrypoint?

- What if Celery is not used in production either?
- Can this association between Celery and Redis variable be moved in the .env file of the production?

global2alex:
Can you post what configuration you used to create your cookie? @demestav

global2alex:
@demestav Yes. Accuracy, reproducibility, future reference... or make a pull request!

demestav:
[...]

PR #1693, opened 2018-06-21
Redis config in local is now conditional on Celery.

demestav:
Redis configuration in .envs/.local/.django should be conditional to Celery. Similarly to `local.yml`, if the user selects not to use Celery, Redis should not be configured since it is not used at all in the local environment. Relevant issue https://github.com/pydanny/cookiecutter-django/issues/1691 .


demestav:
The problem occurs because the local environment Dockerfile uses the enrtypoint of the production environment. Since the production uses Redis irrespective of Celery (also used for caching), it requires the Redis URL definition. So the local .django environment file needs to define it, even if it is actually not used by Django, only because it is required in the production entrypoint file.

{{cookiecutter.project_slug}}/compose/production/django/entrypoint
```diff
@@ -5,9 +5,6 @@ set -o pipefail
 set -o nounset
 
 
-# N.B. If only .env files supported variable expansion...
-export CELERY_BROKER_URL="${REDIS_URL}"
```
webyneter: I'm strongly -1 here: this line is here for exactly the reason mentioned in the comment, to avoid configuration duplication.

demestav: @webyneter Thank you for the feedback. I understand your point; I was reluctant to change it initially for this reason. However, here are some comments:
- The `CELERY_BROKER_URL` is always defined, even though there are cases (like the one that this PR is based on) which Celery is not used. Of course, this doesn't break anything, but why define it if it is not used anywhere?
- Since it is a configuration variable, it would be better to be defined where the rest of the configuration variables are defined, i.e. in the `.django` file
- Even though very unlikely, there could be a case when Celery uses a different Redis server than the one used by the Django cache

For the above reasons, in my opinion, it is better to have the `CELERY_BROKER_URL` defined in the `.django` file. Of course I could be completely wrong :smile: :flushed: so I would like to hear your thoughts.

webyneter: > The CELERY_BROKER_URL is always defined, even though there are cases (like the one that this PR is based on) which Celery is not used. Of course, this doesn't break anything, but why define it if it is not used anywhere?

Sure, go ahead and surround `CELERY_BROKER_URL` with Jinja2 if-clause.

> Since it is a configuration variable, it would be better to be defined where the rest of the configuration variables are defined, i.e. in the .django file

Not much we can do here as`.env` files don't support variable expansion as both of us know.

> Even though very unlikely, there could be a case when Celery uses a different Redis server than the one used by the Django cache

Yes, that's exactly what I was about to address in one of the upcoming PRs -- good for your to have spotted this! Indeed, having `DJANGO_CACHE_URL=redis://redis:6379/0` and `CELERY_BROKER_URL=redis://redis:6379/1` is the best we could do. Tell you what, @demestav, how about you send us a PR? :)

demestav: > Sure, go ahead and surround CELERY_BROKER_URL with Jinja2 if-clause.

@webyneter please check the new commit. I hope that resolves it.

> Tell you what, ..., how about you send us a PR? :)

I will do that!

A general question about the PR process. New changes are constantly made to the upstream repository and as a result conflicts occur on the forked repository. What is the process to follow in such occasions? Do I constantly rebase the fork? Thanks.

webyneter: > Do I constantly rebase the fork?

I'm afraid, yes, rebasing or merging is the only way I could think of and I've given these thoughts quite some time. Unless, of course, I'm missing something.

webyneter: As long as you keep up with upstream changes regularly there shouldn't be a lot of merge conflicts to resolve.

webyneter:
@demestav LGTM. One remark though: unless pushed to a remote you're perfectly OK with rebases, but ones it's out in the wild favor merge commits instead for branch history gets comprehended eaiser then.

browniebroke:
Thanks for you contribution! 

The diff without the rebase noise can be seen on #1956

PR #1696, opened 2018-06-27
Refine generated project Docker / Docker Compose infrastructure

webyneter:

[//]: # (Before you proceed:)

## Description

I've restructured our Docker / Docker Compose project template infrastructure.

### Features

*N.B. `.` denotes a generated project root directory*.

* `./compose` has been replaced with `./docker`:

	```bash
	$ tree ./docker 
	./docker
	├── caddy
	│   ├── Caddyfile
	│   └── Dockerfile
	├── django
	│   ├── Dockerfile
	│   └── scripts
	│       ├── celerybeat
	│       ├── celeryworker
	│       ├── django
	│       ├── django.local
	│       ├── entrypoint
	│       └── flower
	├── mailhog
	│   └── Dockerfile
	├── postgres
	│   ├── Dockerfile
	│   └── scripts
	│       ├── backup
	│       ├── backups
	│       ├── restore
	│       └── _sourced
	│           ├── constants.sh
	│           ├── countdown.sh
	│           ├── messages.sh
	│           └── yes_no.sh
	└── redis
	    └── Dockerfile

	8 directories, 19 files
	``` 

	There is now a single `Dockerfile` for every service we've got, where build instructions vary depending on an `ARG environment` (its de-facto values are `local` and `production` for now), for instance, this is how we're gonna be installing `django` requirements from now on:

	```Dockerfile
	RUN [ "${environment}" = 'production' ] && pip --no-cache-dir install -r "./requirements/${environment}.txt" || pip install -r "./requirements/${environment}.txt"
	```

	Also, I've extracted `mailhog` and `redis` service images to respective `Dockerfile` allowing for image version parity between `local` and `production` Docker Compose environments.

* As for `./envs/*`,

	```bash
	$ tree -a  ./.envs
	./.envs
	├── .caddy
	├── .django
	├── .django.local
	├── .postgres
	└── .postgres.local

	0 directories, 5 files
	```

	Those previously residing in `./.envs/local/` were moved upwards and renamed to `*.local` to match `./docker/**/scripts/*.local` `CMD` scripts; `./.envs/production/*` env files were moved upwards as well.

* `local.yml` and `production.yml` services' `build` sections were extended with build `args`. Self-contained services -- those requiring no extra build artifacts but the ones from the corresponding `Dockerfile` parent directory -- are now built in the context of the containing directory only.

### Trade-offs to be Tackled

* [ ] I had to run `local.yml::services:celerybeat` and `production.yml::services:celerybeat` with `user: root` overriding the `django` `USER` set in the base `./docker/django/Dockerfile`, otherwise `celerybeat.pid`, `celerybeat-schedule` and any other files created by `celery beat` binary cannot be modified by a beat process; this is a known problem, and a number of solutions exist but none of them seem portable enough to me. Any suggestions are welcome. **UPD** filesystem access is now restricted to all users other than root, effectively rendering this workaround regressive -- the only way to fix that is removing `USER django` and `chown ... django:django` from `django` `Dockerfile`...

## Rationale

Until now, generated project environments have been lacking parity, and therefore code execution consistency guarantees could not had been provided. Hope this PR will address that concern once and for all.

`docker-compose -f local.yml build`
`docker-compose -f local.yml up`

`docker-compose -f production.yml build`
`docker-compose -f production.yml up`

@pydanny, @jayfk, @luzfcb, @browniebroke, @sfdye thoughts?


{{cookiecutter.project_slug}}/docker/django/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+COPY ./docker/django/scripts/ /scripts/
+RUN [ "${environment}" = 'production' ] && find /scripts/ -type f -name '*.local' -exec rm -f {} \; || true
+RUN chown --recursive django:django /scripts/
+RUN find /scripts/ -type f -exec sed -i 's/\r//' {} \;
+RUN chmod +x /scripts/*
+RUN mv /scripts/* / \
+    && rmdir /scripts/
+
+COPY . /app/
```
browniebroke: I might be missing something, but won't this slow down local development? My understanding was that it's not required locally as we mount the project root under `/app`:

https://github.com/pydanny/cookiecutter-django/blob/cf8287e5763d3057540ac24fe85c422b7d59413c/%7B%7Bcookiecutter.project_slug%7D%7D/local.yml#L19

{{cookiecutter.project_slug}}/production.yml
```diff
@@ -3,67 +3,88 @@ version: '3'
...
+    <<: *django
     build:
       context: .
-      dockerfile: ./compose/production/postgres/Dockerfile
+      dockerfile: ./docker/django/Dockerfile
+      args:
+        environment: production
+        cmd: /celerybeat
+    image: {{ cookiecutter.project_slug }}_production_celerybeat
+    user: root  # TODO: a temporary workaround -- do it right instead
```
browniebroke: I'm sorry if that's a stupid question, but I didn't understand why it's required now and not before. I thought we were already running the celery beat with the Django image, which is using a non-root user in prod... Is it a pre-existing issue?

webyneter:
@browniebroke since I last pushed here my go-to production Dockerfiles changed a bit: more optimizations and no more if/elses -- you'll like it ones I myself find time to update this PR :)

{{cookiecutter.project_slug}}/docker/django/Dockerfile
webyneter: This is one of the two ways I could think about to facilitate Dockerfile reuse across environments: COPYing the code while further mounting the same dir over the container file system won't slow anything down from my experience with a number of production-grade projects :)

webyneter: Speaking of Dockerfile reuse, unforrtunately, it's impossible now unless you use some clever layer caching-reuse which is bashizm at its finest at least, and producing unmaintainable pile of confiuration at worst... 
N.B. HashiCorp's [Packer](https://www.packer.io/) is a real cure for that, I suggest you find time to take a look at it, if you can -- definitely a noteworthy thing.

PR #1706, opened 2018-07-02
Add mypy to the local dependencies

apirobot:
I see you added types here 8ad7adb11ae276bc8452ba32e5fc7cb53d880b81. It's great, they're useful. We also should add mypy to the dependencies then.


browniebroke:
Thanks!

PR #1711, opened 2018-07-05
changed STATIC_URL and MEDIA_URL to fix errors with Compress

narnikgamarnikus:

[//]: # (Before you proceed:)

## Description

## Rationale

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -87,14 +87,14 @@ AWS_S3_OBJECT_PARAMETERS = {
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
 {%- else %}
 STATICFILES_STORAGE = 'config.settings.production.StaticRootS3BotoStorage'
-STATIC_URL = f'https://s3.amazonaws.com/{AWS_STORAGE_BUCKET_NAME}/static/'
+STATIC_URL = f'https://{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com/static/'
 {%- endif %}
 
 # MEDIA
 # ------------------------------------------------------------------------------
 {% if cookiecutter.use_whitenoise == 'y' -%}
 DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
-MEDIA_URL = f'https://s3.amazonaws.com/{AWS_STORAGE_BUCKET_NAME}/'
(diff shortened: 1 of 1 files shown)


webyneter:
Any explanation why this works and why the former wasn't working?

narnikgamarnikus:
Django compress searches for files at STATIC_URL = f'https: // {AWS_STORAGE_BUCKET_NAME} .s3.amazonaws.com / static / ', but the default path was specified f'https: //s3.amazonaws.com/ {AWS_STORAGE_BUCKET_NAME} / static /

I deployed the project with these settings to the staging server and I managed to solve the problem this way.

{{cookiecutter.project_slug}}/.dockerignore
```diff
@@ -2,3 +2,4 @@
 !.coveragerc
 !.env
 !.pylintrc
+node_modules/
```
webyneter: This must be made conditional depending on presense or absense of `node_modules/`, for instance, this is the case when `cookiecutter.js_task_runner == "y"`. Other cases need to be thought over as well.

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -87,22 +87,22 @@
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
 {%- else %}
 STATICFILES_STORAGE = 'config.settings.production.StaticRootS3BotoStorage'
-STATIC_URL = f'https://s3.amazonaws.com/{AWS_STORAGE_BUCKET_NAME}/static/'
+STATIC_URL = f'https://{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com/static/'
```
webyneter: S3 buckets are reachable via subdomain (like you've got here) and `/` path notation alike. Therefore, the variant you propose is more generic which is a definitive plus.

hooks/post_gen_project.py
```diff
@@ -310,21 +310,6 @@ def main():
...
-            WARNING
-            + "Docker and {} JS task runner ".format(
-                "{{ cookiecutter.js_task_runner }}".lower().capitalize()
-            )
-            + "working together not supported yet. "
-              "You can continue using the generated project like you "
-              "normally would, however you would need to add a JS "
-              "task runner service to your Docker Compose configuration "
-              "manually." + TERMINATOR
-        )
```
webyneter: This is a change of different nature -- make sure to open another PR and address it there.

narnikgamarnikus:
Take this PR.
I still have a few committees to improve.

browniebroke:
Someone submitted another pull request so I want ahead and merged it. Thanks for your help.

PR #1714, opened 2018-07-09
Replace Caddy with Traefik

demestav:

[//]: # (Before you proceed:)

## Description

Replace Caddy with Traefik

## Rationale

There is some trouble with the Caddy license (https://github.com/pydanny/cookiecutter-django/pull/1282#issuecomment-329617536) 

@drdaeman suggested using Traefik (https://github.com/pydanny/cookiecutter-django/pull/1282#issuecomment-353655273) which supports ACME and also plays very nice with Docker.

## Comments
I am currently using the proposed setup on a live site and it working great so far. If this PR is of interest to the maintainers, then I could commit more changes and take care of the documentation. Of course, any suggestions by the more experienced people around here, are welcome!


jayfk:
Is traefik now considered to be a stable webserver? I‘ve used it in the past as internal load balancer, wondering if that has changed.

sfdye:
I'd like to see this.

demestav:
@jayfk Not sure how stable Traefik is considered in the industry. Note that their business model seems to be close to what you mentioned here https://github.com/pydanny/cookiecutter-django/pull/1282#issuecomment-329732412

@sfdye The current state of the branch is fully operational if you want to test it out. I will try to spot any corrections or improvements and commit them.

demestav:
Here are some comments:
- The traefik Dockerfile is not as clean as it should. The reason for that is because the `acme.json` file needs to be in a separate volume so that when you recreate the containers, let's encrypt wont have to generate a new one.
- Traefik can serve a dashboard, where one can observe the status (https://docs.traefik.io/configuration/api/#web-ui). This needs to have an open port (currently 8080). You can block unauthorized access to this using authorization. Do we need that? Because this port should be exposed publicly anyway. I think it is the same case with Celery Flower.

Your comments are welcome.

webyneter:
Why not nginx + [nginx-proxy](https://github.com/jwilder/nginx-proxy) (with the docker socket mounted as `:ro`) + [docker-gen](https://github.com/jwilder/docker-gen) + [docker-letsencrypt-nginx-proxy-companion
](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) as [separate containers](https://github.com/jwilder/nginx-proxy#separate-containers)? That would give us renewable HTTPS certificate infrastructure + rich `nginx-proxy` multitenancy capabilities, as wll as dynamic nginx config updates out of the box. Seems like a decent production setup to me.

webyneter:
OK, I've been looking through Traefik docs for a few minutes now, and I feel like I'm sold -- gotta look into their promises more closely to see if they really stand by them, but what I've read already looks great wihtout doubt.

webyneter:
Anyway, the sound fact of Traefik focusing on microservice architecture deployments should be clearly stated somewhere in our docs as opposed to classic nginx-based setups (like the one I mentioned above) for conventional monoliths.

sfdye:
@webyneter see #1445 for my previous attempt

demestav:
@sfdye @webyneter According to https://github.com/pydanny/cookiecutter-django/pull/1445#issuecomment-357717197 mounting the Docker socket seems to be risky. Maybe I am wrong but Traefik also needs the Docker socker mounted https://docs.traefik.io/#1-launch-trfik-tell-it-to-listen-to-docker

jayfk:
Why is it necessary to mount the docker socket for traefik? This shouldn‘t be the case here because we don‘t need to support multiple independent Cookiecutter Django deployments per server.

demestav:
@jayfk I am not sure whether the mounting the Docker socket is optional or required. I just followed their documentation. Indeed, if mounting the socket is required **only if** you want to have auto discovery, we could remove it.

demestav:
Just an update. It seems that the Docker socket is indeed required for the auto discovery and configuration for Docker. We can remove it and manually setup the configuration using the [file] provider (https://docs.traefik.io/configuration/backends/file/) and set up the appropriate frontend and backend to connect it to the Django container. My initial tests worked. I will do more tests, add the ACME and commit once it is tidy.

yunti:
From the caddy docs, with the transparent preset:
https://caddyserver.com/docs/proxy

transparent
Passes thru host information from the original request as most backend apps would expect. Shorthand for:
header_upstream Host {host}
header_upstream X-Real-IP {remote}
header_upstream X-Forwarded-For {remote}
header_upstream X-Forwarded-Proto {scheme}

demestav:
Sure.

ohld:
LGTM!

yunti:
All set to merge?

browniebroke:
Looks like we let this one slip again, we have new conflicts... 😞 

As a side note, can we check that this won't cause any regression of 2 latest changes that were added for Caddy #1801 and #1852?

Marpop:
@demestav I include your PR in my local project (earlier I used nginx). Do I have to do something more (of course, exclude domain and email changing)? Because it seems doesn't work with my project

demestav:
@browniebroke Once I have some free time, I will take a look.

@Marpop What errors are you getting? From what I can remember, it should work as is.

Marpop:
> @browniebroke Once I have some free time, I will take a look.
> 
> @Marpop What errors are you getting? From what I can remember, it should work as is. I have a production site that runs on this PR setup since July.

Sorry, my bad. Everything is working on Safari, but Chrome report:  NET::ERR_CERT_AUTHORITY_INVALID

Marpop:
@demestav did you have problems with serving static/media files with teafik? Now I have 
It's testing server so I'm serving all media by Django instead something like AWS S3

demestav:
@Marpop Well, as far as I know, static files are served through Django with Whitenoise, so that works with Traefik. However, if you want to serve media files through Django, you cannot do that with Traefik only. I read that this was a design decision from the Traefik team, since there is other software that does this job well. I think they suggest to use Nginx behind Traefik to serve the media files.

Marpop:
@demestav Roger, thank you

demestav:
@browniebroke Ragarding #1801 it seems that Traefik configuration file (.toml) does not support environment variables, and that is the reason we use the cookiecutter parameters directly.

{{cookiecutter.project_slug}}/compose/production/traefik/traefik.toml
```diff
@@ -0,0 +1,44 @@
...
+[file]
+[backends]
+  [backends.django]
+    [backends.django.servers.server1]
+      url = "http://django:5000"
+
+[frontends]
+  [frontends.django]
+    backend = "django"
+    passHostHeader = true
```
browniebroke: To keep the change from #1852 we need to whitelist the `X-CSRFToken` to proxy it.

I dug a bit in the documentation, and it seems it's done like this:

```suggestion
    passHostHeader = true
    [frontends.django.headers]
      HostsProxyHeaders = ['X-CSRFToken']
```

Source: [security headers documentation](https://docs.traefik.io/basics/#security-headers) which points at [unrolled/secure for the possible secure headers](https://github.com/unrolled/secure#default-options).

{{cookiecutter.project_slug}}/compose/production/traefik/traefik.toml
```diff
@@ -0,0 +1,44 @@
+logLevel = "ERROR"
```
browniebroke: I was a bit surprised to not see any logs from the Traefik container, until I saw this. From a simple test, it looks like `INFO` shows some activity without being too noisy. Should we use that instead?

```suggestion
logLevel = "INFO"
```

demestav:
@browniebroke Thank you very much for the suggested changes. I have committed both. What do you think about the other two, `X-Real-IP` and `X-Forwarded-Proto`, which are currently included in the Caddyfile?

browniebroke:
From what I checked, Traefik forwards them already with the header setting you added before, so there is nothing to do for these, I think.

browniebroke:
Do you want to update any documentation before we merge? I can't remember if we had much around Caddy...

demestav:
@browniebroke Sure, I will update the documentation and commit.

docs/developing-locally-docker.rst
```diff
@@ -105,7 +105,7 @@ The most important thing for us here now is ``env_file`` section enlisting ``./.
     │   ├── .django
     │   └── .postgres
     └── .production
-        ├── .caddy
+        ├── .traefik
```
browniebroke: This file does not exist, should remove this line.

demestav: Indeed, I missed that somehow.

demestav:
We still have to decide how to handle the Dashboard feature of Traefik. 
https://github.com/pydanny/cookiecutter-django/blob/72f041253bf61dc9ff7467613c0a960577e77bb2/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/traefik.toml#L4-L7

@browniebroke please check the second bullet point of an older comment of mine https://github.com/pydanny/cookiecutter-django/pull/1714#issuecomment-403756583.

Personally, I would rather not have on a public facing port, even if it is password protected (https://docs.traefik.io/configuration/api/#authentication). I vote to remove the Dashboard entirely. Let me know of your thoughts.

browniebroke:
Thanks for raising this point again, I've missed it before. I agree we should disable it by default.

demestav:
@browniebroke Please note that I am currently not able to do testing at the moment, so I cannot confirm that all work as they should.

{{cookiecutter.project_slug}}/production.yml
```diff
@@ -30,20 +30,19 @@ services:
...
       - django
     volumes:
-      - production_caddy:/root/.caddy
-    env_file:
-      - ./.envs/.production/.caddy
+      - production_traefik:/etc/traefik/acme
     ports:
       - "0.0.0.0:80:80"
       - "0.0.0.0:443:443"
+      - "0.0.0.0:8080:8080"
```
browniebroke: I think this is no longer needed as the dashboard is disabled, correct?

```suggestion
```

luzfcb:
Almost 4 years late, I found out today that Caddy is now 100% open source

https://github.com/caddyserver/caddy/issues/2786
https://github.com/wemake-services/wemake-django-template/issues/36#issuecomment-623161073

PR #1717, opened 2018-07-10
removed default action from project.js #1716

### LINKED ISSUE #1716: Crispy Forms horizontal layout will be deactivated by project.js
I tried some days to use horizontal forms with crispy forms but it didn't work until I found that the default project.js contains the command to remove the row attribute from the form group:

```js
$('.form-group').removeClass('row');
```

Doesn't make it more sense to remove it?

hendrikschneider:

[//]: # (Before you proceed:)

## Description
Closes #1716.

Default should be a project.js that does not modify the html in some way

## Rationale

A cookiecutter project should generate a good starting point to start with and all modules should work without problems. This undocumented feature breaks crispy horizontal forms in a default setup and it is not obvious that there there is line in the project.js that removes css classes.


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/static/js/project.js
```diff
@@ -18,4 +18,3 @@ Issues with the above approach:
 3. Really Unforgiving: No way to override this behavior
 4. Undocumented: No mention in the documentation, or it's too hard for me to find
 */
-$('.form-group').removeClass('row');
```
browniebroke: The comments above this line related to crispy forms should be removed if this line is removed.

hendrikschneider: done. Should I stash the changes or is it fine like this?

browniebroke: No, don't worry about it, we can squash from the Github UI.

webyneter:
LGTM.

browniebroke:
Thanks for the fix!

PR #1735, opened 2018-08-02
Add note about using keep_local_envs_in_vcs

purplediane:

[//]: # (Before you proceed:)

## Description

Add a note that the `keep_local_envs_in_vcs` only makes sense when using Docker or Heroku.

## Rationale

As a newbie, I wasn't sure about `keep_local_envs_in_vcs`, so I said yes, and when CC was building, it gave me the message: 
"[INFO]: .env(s) are only utilized when Docker Compose and/or Heroku support is enabled so keeping them does not make sense given your current setup." 
Seems like it could go in this documentation, and make things easier for newbies.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -89,6 +89,7 @@ Listed in alphabetical order.
   Delio Castillo           `@jangeador`_                 @jangeador
   Denis Orehovsky          `@apirobot`_
   Dónal Adams              `@epileptic-fish`_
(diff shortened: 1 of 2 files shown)


webyneter:
@purplediane thank you! As this is the first time you contribute to us would you mind adding yourself to `CONTRIBUTORS`?

purplediane:
Thank you! I added myself to the CONTRIBUTORS.rst file.

PR #1749, opened 2018-08-15
Update bug issue template to ask to include the replay file

browniebroke:
Often, users report issues due to a specific combination of options, which required to get them. The current way is inefficient for users as well as maintainers:

- The user needs to copy/paste a terminal output, which might be gone if they closed the window
- Maintainer needs to regenerate with the same options
- Maintainer needs to be careful to bypass their potential profile defaults
- In case of error to one question, needs to re-generate

Cookiecutter has got a replay feature for about 3 years now, which seems like a more automated way to get the options from the user.

This updates the issue template for bugs report using this.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -17,7 +17,5 @@ about: Report a bug
 
 [//]: # (Any or all of the following:)
 [//]: # (* Host system configuration: OS, Docker & friends' versions etc.)
-[//]: # (* Project generation options)
+[//]: # (* Replay file https://cookiecutter.readthedocs.io/en/latest/advanced/replay.html)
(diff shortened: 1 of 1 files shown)


Afrowave:
Very good. I had no idea of this "replay feature". Sounds like it can be used to generate a project with a specific set of preconfigured options, but still give choices. Am I going in the right direction?

browniebroke:
Possibly, but I don't know how it handles missing options. I'm looking at getting [an enhancement merged](https://github.com/audreyr/cookiecutter/pull/906) for the consuming side into the cookiecutter CLI, to add the `--reply-file` option.

sfdye:
I like this idea.

webyneter:
I like that.

browniebroke:
Thanks, I'm going to merge this 👍

PR #1784, opened 2018-09-04
Rename custom S3Boto3Storage subclasses

sfdye:
Closes #1788 

This PR preps for the upcoming `django-storages` 2.X:

1. ~~Explicitly set default ACL to None which was [incorrectly set](https://github.com/jschneier/django-storages/blob/3f6ec7494253dd715598a41dd821fc09d6f45e40/storages/backends/s3boto3.py#L195) to `public-read` by `django-storages`~~

2. Renamed `StaticRootS3BotoStorage` to `StaticRootS3Boto3Storage` to avoid confusion. This is to prep for the upcoming `S3Boto` class deprecation.

See also:
https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
https://github.com/jschneier/django-storages/blob/master/CHANGELOG.rst


sfdye:
Ahhh, you are right. Static files should be public, but user uploaded content (media) should not be right?

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -73,6 +73,10 @@
 # https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
 AWS_STORAGE_BUCKET_NAME = env('DJANGO_AWS_STORAGE_BUCKET_NAME')
 # https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
+AWS_DEFAULT_ACL = env('DJANGO_AWS_DEFAULT_ACL', default=None)
+# https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
+AWS_BUCKET_ACL = env('DJANGO_AWS_BUCKET_ACL', default=None)
```
browniebroke: It says in their doc:

> Only used if `AWS_AUTO_CREATE_BUCKET=True`. The ACL of the created bucket.

Since that other setting is False by default and we don't override it, should we set this value? If someone changes the auto-create bucket setting after generating a project, isn't it up to them to make sure this is changed accordingly?

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -101,17 +105,17 @@
 from storages.backends.s3boto3 import S3Boto3Storage  # noqa E402
 
 
-class StaticRootS3BotoStorage(S3Boto3Storage):
+class StaticRootS3Boto3Storage(S3Boto3Storage):
```
browniebroke: 👍

{{cookiecutter.project_slug}}/config/settings/production.py
webyneter: I agree with @browniebroke: I specifically set `AWS_AUTO_CREATE_BUCKET` to `False` as otherwise that would yield unexpected result (maybe the end user wants to create a bucket himself, attaching a few policies to it prior to uploading anything there etc).

browniebroke:
> Static files should be public, but user uploaded content (media) should not be right?

Sounds like a sensible default, depending on the app you're building, you might not want user-uploaded content to be public, e.g. financial or medical documents. Also when using Whitenoise, the bucket is only used for media storage.

webyneter:
@browniebroke that reallly comes down to the definition of "publicly-accessible" files: if we're talking about non-listable bucket, then yes, but eventually you gotta get each individual file somehow -- does allowing GetObject operations make the bucket public (provided no S3 bucket auth mechanism is used)?.. That'swhat I always resented about AWS definitions.

{{cookiecutter.project_slug}}/config/settings/production.py
sfdye: > I specifically set AWS_AUTO_CREATE_BUCKET to False
@webyneter Where did you set this? I couldn't find it..

sfdye:
I guess with `AWS_AUTO_CREATE_BUCKET` set to `False`, there is no point in discussing bucket ACL or default ACL. In this case, I will make this PR to rename from boto to boto3 only.

webyneter:
https://github.com/pydanny/cookiecutter-django/pull/1788

sfdye:
Done

PR #1796, opened 2018-09-12
support docker-sync for mac users

hoffmannkrzysztof:

[//]: # (Before you proceed:)

## Description

Every mac users suffer using docker on their macs. Sharing files through containers is very slow.

## Rationale

Only mac users need this. Other users don't care :)

If you are using many static files in containers loading speed of each request decrease.


webyneter:
@hoffmannkrzysztof could the `delegated` flag facilitate in healing that wound somehow?

hoffmannkrzysztof:
@webyneter I not sure if I understand you clearly. But docker-sync helps a lot of users who using mac while developing. Take a look at the official documentation: http://docker-sync.io/. I am using this solution and see only pros.

sfdye:
> Every mac users suffer using docker on their macs

1. I am a Mac & Docker user. I didn't experience any slowness whatsoever with sharing files from docker containers. How could you reproduce exactly?

2. `docker-sync` seems to [support](https://github.com/EugenMayer/docker-sync/wiki/1.-Installation) windows/linux as well, why not support them?

3. Shouldn't it be an optional feature, for added complexity?

hoffmannkrzysztof:
@sfdye 
1. This issue on mac is related to a large number of I/O operations. The easy way to reproduce this issue is by using a large number of assets. Take a look for more info: https://github.com/docker/for-mac/issues/77 
2. Of course, docker-sync support Windows and Linux and can be used. 
3. This addon is optional. If you don't want sync when you will use regular docker-compose

elyak123:
I used to have that same issue until I started using `docker-machine`, try it, is worth the shot.

browniebroke:
Docker for mac has improved the file sharing implementation with VirtuoFS:

https://www.docker.com/blog/speed-boost-achievement-unlocked-on-docker-desktop-4-6-for-mac/

In my experience it's much faster, so I'm closing this, I consider it no longer needed.

PR #1801, opened 2018-09-14
fixed production bugs in caddyfile and middleware

ChrisPappalardo:
## Description

Fix bugs in the production environment configuration.

## Rationale

Caddyfile change - the redirect should be consistent with the main site definition, which uses the env variable $DOMAIN_NAME

Settings change - Whitenoise docs say to place middleware *DIRECTLY AFTER* Django security middleware per http://whitenoise.evans.io/en/stable/django.html#enable-whitenoise

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/caddy/Caddyfile
+++ b/{{cookiecutter.project_slug}}/compose/production/caddy/Caddyfile
@@ -1,5 +1,5 @@
 www.{% raw %}{$DOMAIN_NAME}{% endraw %} {
-    redir https://{{cookiecutter.domain_name}}
+    redir https://{% raw %}{$DOMAIN_NAME}{% endraw %}
 }
 
 {% raw %}{$DOMAIN_NAME}{% endraw %} {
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/compose/production/caddy/Caddyfile
```diff
@@ -1,5 +1,5 @@
 www.{% raw %}{$DOMAIN_NAME}{% endraw %} {
-    redir https://{{cookiecutter.domain_name}}
+    redir https://{% raw %}{$DOMAIN_NAME}{% endraw %} {
```
browniebroke: It looks like there is an extra `{` at the end of the line. Assume that was a wrong copy/paste? 

Not sure why we didn't use the environment variable here, the new version seems dryer to me 👍

ChrisPappalardo: yep, sorry about that.  i fixed it and merged in the latest changes to master.

PR #1804, opened 2018-09-15
Added React JS integration using create-react-app without ejecting

genomics-geek:

[//]: # (Before you proceed:)

## Description

Adding the option to use ReactJS as an option for UI development.  Leverage Facebook's [create-react-app](https://github.com/facebook/create-react-app) without ejecting, so reduces the need to maintain a lot of webpack, nodejs, etc configurations. 

## Rationale

This allows a developer to quickly setup a Django/ReactJS project with minimal setup.  This is not always so clear - hopefully this helps!

Developing projects with Django REST APIs and using ReactJS


genomics-geek:
I submitted a different PR before, but required create-react-app ejection which would create a lot of overhead to maintain and update.  This is a solution where it would not require create-react-app ejection. 

I wonder if everyone agrees that this may make it easer to maintain since [create-react-app](https://github.com/facebook/create-react-app) abstracts all the webpack config, etc?

Everything in `frontend` is basically what `create-react-app` generates. The only changes were:

1. Adding proxy to `package.json` to allow API calls against Django REST API. [See this](https://github.com/chopdgd/cookiecutter-django-reactjs/blob/feature-create-react-app/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/package.json#L5)
2. Adjusting react-scripts to [build css and allow hot-reloading](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-a-css-preprocessor-sass-less-etc). [here](https://github.com/chopdgd/cookiecutter-django-reactjs/blob/feature-create-react-app/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/package.json#L7-L14).
3. Adding an example to show to call REST API from React. [See this](https://github.com/chopdgd/cookiecutter-django-reactjs/blob/feature-create-react-app/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/src/App.js#L9-L17)

{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -3,9 +3,13 @@
 """
 
 import environ
+import os
 
 ROOT_DIR = environ.Path(__file__) - 3  # ({{ cookiecutter.project_slug }}/config/settings/base.py - 3 = {{ cookiecutter.project_slug }}/)
 APPS_DIR = ROOT_DIR.path('{{ cookiecutter.project_slug }}')
+{% if cookiecutter.js_task_runner == 'CreateReactApp' -%}
```
webyneter: Consistency is key -- make sure to rename both `CreateReactApp` and `createreactapp` to `create_react_app` everywhere, and use that spelling universally.

webyneter:
React is beyond my expertise -- does anyone else possessing the right skill set want to review that? @pydanny, @jayfk, @browniebroke, @sfdye, @luzfcb ?

Afrowave:
@genomics-geek we will need to maintain the fork we have on our own so that those who want to use React can work with it or the one by @anchpop - https://github.com/anchpop/cookiecutter-django or @hersonls - https://github.com/hersonls/frango.

Issue #1781, opened 2018-09-02
Clarify how to setup Postgres for first time with createdb command


yemarnevets:
## Description

[//]: # (What's it you're proposing? How should it be implemented?)

In the section on `createdb` the explanation for default settings used in the PostgresDB could be clarified. 

Specifically, if someone can explain how to do this, given the Django Cookie-Cutter setup, that would be helpful. 

```note: if this is the first time a database is created on your machine you might need to alter a localhost-related entry in your pg_hba.conf so as to utilize trust policy```

## Rationale

[//]: # (Why should this feature be implemented?)

Easier setup for first time users of Django Cookie Cutter. 

## Use case(s) / visualization(s)
`createdb my_project` asks for a password, which I do not know since Postgres has not yet been setup. I'm told to alter a pg_hba.conf, which I then have to figure out how to use and where it should go and then verify that it is setup properly.

browniebroke:
Thanks for reporting this, so given the links you provided, I assume you're trying to run your project on Mac OS (without Docker). 

The easier way to install Postgres on a Mac is using [Homebrew](https://brew.sh/). This is documented [in the Postgres wiki](https://wiki.postgresql.org/wiki/Homebrew) and after installation, homebrew should show instructions on how to set up your server.

If the above resolve your problem and you think [the documentation](https://github.com/pydanny/cookiecutter-django/blob/master/docs/developing-locally-docker.rst) is outdated or can be improved, you are welcome to send us a pull request to update it.

sfdye:
If not using homebrew, I found [this](https://postgresapp.com/) to be the easiest way to install postgres fo mac.

diek:
You should create a database user for your app, that user should only have access to your project database imo

Issue #1775, opened 2018-08-30
Environment variables error in base.py and local.py


Chhaya02:
![screen shot 2018-08-30 at 12 49 17 pm](https://user-images.githubusercontent.com/20675052/44835946-3f5b2100-ac53-11e8-8cc6-aa24dbb2d4dd.png)
When I try to run my project using cookiecutter setup, I got this exception : 
     django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

To solve this, Followed this steps and include the code in manage.py : 
i)import django 
ii)django.setup()
inside the If block.

After that, the new exception raised i.e
   django.core.exceptions.ImproperlyConfigured: Set the DATABASE_URL environment variable

So, it's saying that values are not getting set from the .env files.
To solve this I made changes and set the default value for the variables.

Before setting the default values for my environment variable in my base.py: 
DATABASES = {
    'default': env.db('DATABASE_URL'),
 }
CELERY_BROKER_URL = env('CELERY_BROKER_URL')

After setting the default values for my environment variable in my base.py: 
DATABASES = {
   'default': env.db('DATABASE_URL',default="postgres://user_name:password@127.0.0.1:5432/db_name"),
 }
CELERY_BROKER_URL = env('CELERY_BROKER_URL',default='redis://redis:6379/0')

I can't hard code values for every service. it should be taken from the .env files i.e  .django and .postgres from .local folder of .envs directory

Os: macOS 
version : 10.13.3

sfdye:
Are you using docker?

Chhaya02:
No, Right now we are not using Docker but we have opted it for local and production environment for the future.

connectgautam:
How we will run migration if we are using Docker

browniebroke:
@connectgautam your question seems off-topic, please could you find a more appropriate channel?

> First, make sure to examine [the docs](https://cookiecutter-django.readthedocs.io/en/latest/). If that doesn't help post a question on [StackOverflow](https://stackoverflow.com/questions/tagged/cookiecutter-django) tagged with `cookiecutter-django`. Finally, feel free to join [Gitter](https://gitter.im/pydanny/cookiecutter-django) and ask around.

browniebroke:
@Chhaya02 if you're not using Docker, I would recommend answering "no" when being asked the question by cookiecutter. 

> I can't hard code values for every service. it should be taken from the .env files i.e .django and .postgres from .local folder of .envs directory

[...]

Issue #1773, opened 2018-08-29
env.example missing


shayneoneill:
I've been grepping up and down the directory , even rolling back a few versions of the git repo and for the life of me I cant seem to find the env.example file. I cant currently determine what the .env variables are or what should go in them, and it doesn't seem to be documented anywhere?

Is there somewhere I can go to find an example or some sort of documentation? (The only documentation I found told me to copy env.example!)

sfdye:
See #1295

sfdye:
https://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html#configuring-the-environment

In short, there has been a major refactoring of this repo. There is no more env.example.

shayneoneill:
Ah ok. It would be good if one could be restored, for some reason the docs where not showing up on my google search.

webyneter:
@shayneoneill 

>  It would be good if one could be restored

PR #1805, opened 2018-09-15
Update documentation for loacal developement on bare metal

### LINKED ISSUE #564: Revise and improve docs/developing-locally.rst
The "Getting Up and Running Locally" document needs revision from someone with great attention to detail and writing skills. Basically it needs to be checked for usability and modified as appropriate.

Perhaps this could be a way for a Cookiecutter Django newbie to get comfortable with the project while contributing.

browniebroke:
I'm hoping that this update will clarify the bare metal setup which is causing recurring issues. This is essentially adding some details to the local development on bare metal, but also fixes some outdated settings and a couple of minor styling issues.

Refs #1781 
Refs #1773 
Refs #1775 
Fixes #564


docs/developing-locally.rst
```diff
@@ -9,24 +9,53 @@ Setting Up Development Environment
...
-#. Create a new PostgreSQL database (note: if this is the first time a database is created on your machine you might need to alter a localhost-related entry in your ``pg_hba.conf`` so as to utilize ``trust`` policy): ::
+#. Create a new PostgreSQL database using createdb_: ::
 
     $ createdb <what you've entered as the project_slug at setup stage>
 
+   .. note::
+
+       if this is the first time a database is created on your machine you might need to alter
+       a localhost-related entry in your ``pg_hba.conf`` so as to utilize ``trust`` policy, see the
+       `postgres documentation`_ for more details.
```
browniebroke: I'm not 100% sure what this means tbh, I never had to set this up on my machine. If anyone has more details, I'd be more than happy to elaborate.

docs/developing-locally.rst
```diff
@@ -9,24 +9,53 @@ Setting Up Development Environment
 
 Make sure to have the following on your host:
 
-* virtualenv_;
-* pip;
```
browniebroke: I dropped `pip` & `virtualenv` as they both come pre-installed with Python 3.6.

docs/developing-locally.rst
```diff
@@ -9,24 +9,53 @@ Setting Up Development Environment
...
+* PostgreSQL_.
+* Redis_, if using Celery
 
 First things first.
 
-#. `Create a virtualenv`_.
+#. Create a virtualenv: ::
 
-#. Activate the virtualenv you have just created.
+    $ python3 -m venv <virtual env path>
```
browniebroke: I just realised that `python3` might point to Python 3.7, which isn't supported yet (due to Celery, at least).

sfdye: why not use `virtualenv <path/to/new/virtualenv/> --python=python3.6`

docs/developing-locally.rst
webyneter: I believe this was my input: in past, I did face the issue with `pg_hba.conf` localhost auth records.

docs/developing-locally.rst
browniebroke: This one is included in [python since version 3.3](https://docs.python.org/3.6/library/venv.html), so that's one fewer thing to install, no need to have a requirement on the standalone `virtualenv` library.

Correct me if I'm wrong, but I think they do the same thing.

docs/developing-locally.rst
webyneter: I suggest that this note stays intact: it proved itself helpful to me in past, no reason to assume the opposite for everyone else. It might require clarifications, though.

docs/developing-locally.rst
sfdye: I think you are right. We can get rid of virtualenv. They basically do the same thing but is `venv` is a built-in module.

https://www.reddit.com/r/learnpython/comments/4hsudz/pyvenv_vs_virtualenv/

So we can use `python3.6 -m <virtual env path>` or `/path/to/python3.6 -m <virtual env path>`

docs/developing-locally.rst
luzfcb: I usually set up password for the `postgres` database user and change the authentication mode to md5:

in Ubuntu would be something like this:

```bash
# runs the "psql" command as a "postgres" linux user, and changes the password of "postgres" database user
# yes, when you install postgres on linux (at least on Ubuntu), a linux user named "postgres" is created

sudo -u postgres psql postgres -c "ALTER USER postgres WITH PASSWORD 'new_password';"
```
change `pg_hba.conf` configuration file and enable md5 authentication

>         in ubuntu, this file is located in:
>        /etc/postgresql/POSTGRESH_VERSION_NUMBER/main/pg_hba.conf
>        
>        if the version of postgres is 9.6
>
>       
>        /etc/postgresql/9.6/main/pg_hba.conf
>        
>        if is 10
>        
>        /etc/postgresql/10/main/pg_hba.conf


open and change the pg_hba.conf
from:
```
local   all             all                                     peer
```
to
```
local   all             all                                     md5
```
restart postgresql

```
sudo service postgresql restart

```

docs/developing-locally.rst
```diff
@@ -9,24 +9,53 @@ Setting Up Development Environment
...
 
+   .. note::
+
+       if this is the first time a database is created on your machine you might need to alter
+       a localhost-related entry in your ``pg_hba.conf`` so as to utilize ``trust`` policy, see the
+       `postgres documentation`_ for more details.
+
+#. Set the environment variables for your database(s): ::
+
+    $ export DATABASE_URL=postgres://<project_slug>
```
luzfcb: if use the way I configure the postgres, then must also pass the user and password as a parameter, I'm also explicit about the host and port

```
export DATABASE_URL=postgres://postgres:password@127.0.0.1:5432
```

docs/developing-locally.rst
luzfcb: here is a simple, and good documentation on how to configure postgresql for first use

http://suite.opengeo.org/docs/latest/dataadmin/pgGettingStarted/firstconnect.html

browniebroke: I'd really like to add more details on this, but it seems to be a bit overwhelming to include in this page.

I've looked at this to see how different it is on macOS. I've re-installed `postgresql` with Homebrew and it looks like the setup is a bit different. The `pg_hba.conf` file is located under `/usr/local/var/postgres/pg_hba.conf` and it uses the `trust` policy. The DB server is running as the current user (not postgres), but has no password either.

Finally, I have no way to test how this would work under Windows...

I've deleted `docs/installing_postgres.rst` for now, I could either add it back and elaborate on this topic, or link to that external documentation.

docs/developing-locally.rst
browniebroke: Agreed. I think it would be better to include the entire thing here, actually:

```
export DATABASE_URL=postgres://<user>:<password>@127.0.0.1:5432/<db_name>
```

browniebroke:
I think I have addressed all the comments. This is still perfectible, but it's better than the current version IMO, so I'd say it's worth merging. We can improve further from there. Thoughts?

PR #1828, opened 2018-10-11
Rename local.yml production.yml

minhuyen:

[//]: # (Before you proceed:)

## Description

I want to rename local.yml to docker-compose.yml (the default file load when run docker-compose)
and rename production.yml to docker-compose.prod.yml. 

They help distinguish them from other configuration files and are easy to find for configuration

## Rationale


webyneter:
I'm -1 on this one. The rationale behind current naming style is to distinguish between runtime modes explicitly rather than implicitly falling for Docker Compose defaults. My perception is that having to run `docker-compose -f local.yml ...` vs `docker-compose -f production.yml ...` instead of `docker-compose ...` vs `docker-compose -f docker-compose.yml -f docker-compose.prod.yml ...`, respectively, makes it clear for the end user what stack is desired at the moment. Besides, `docker-compose(.prod).yml` notation is way more verbose.

Cookiecutter Django is a powerful yet generic project template, and in the end of the day it's up to the users to adapt it to their specific requirements.

PR #1830, opened 2018-10-11
Remove requirements.txt file from delete list.

### LINKED ISSUE #1829: Creating a new project with TravisCI support fails first build.
## What happened?
Creating a new project with travisci enabled will fail on its first run

It should pass

## Steps to reproduce

1. Create new project using this cookie cutter template. 
2. Enable travisci and run build.
3. Build fails due to `requirements.txt` not existing in root directory

cmackenzie1:
refs #1829

[//]: # (Before you proceed:)

## Description

Removing the `requirements.txt` from the list of files to delete if heroku is not enabled.

## Rationale

Using TravisCI but not enabling Heroku causes the `requirements.txt` file in the root dir to be deleted, thus needing an overridden `.travis.yml` install key.

As a user of this template, I want to create an app without Heroku support, but support for TravisCI, using the output of `cookiecutter`

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -78,6 +78,7 @@ Listed in alphabetical order.
   Chris Franklin           `@hairychris`_
   Chris Pappalardo         `@ChrisPappalardo`_
   Christopher Clarke       `@chrisdev`_
+  Cole Mackenzie           `@cmackenzie1`_
   Collederas               `@Collederas`_
   Cristian Vargas          `@cdvv7788`_
   Cullen Rhodes            `@c-rhodes`_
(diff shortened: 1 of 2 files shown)


cmackenzie1:
@webyneter thanks for the suggestion. In my opinion, the `requirements.txt` file should exist in the root directory at all times, not just if we are using Heroku or TravisCI - it is a very common practice in python projects. Performing an attempted delete, but then not deleting it because we are using TracisCI seems a little backward in my opinion.

LMK what you think.

webyneter:
@cmackenzie1 the thing is, we've already got `./requirements/` with a number of requirements files in it: some of them serve as base templates -- like `base.txt` -- while the others -- `test.txt`, `local.txt`, and `production.txt` -- are used directly. As a result, there's no need to have yet another requirements file serving merely as a proxy; it's only required in Heroku bare-bones deployment setup anyway... My point is, if not for Heroku we wouldn't have `./requirements.txt` in the first place.

cmackenzie1:
@webyneter ahh ok. I made the change, lmk what you think.

webyneter:
Thanks! I'll check it out later this week.

PR #1836, opened 2018-10-23
Keep ipython history

### LINKED ISSUE #1794: Ipython history: keep it around for the django shell
Every time you run this command to use the django shell, the history is gone:
```
docker-compose -f local.yml run --rm django python manage.py shell_plus
```

Is there a way to keep the ipython history around?

I tried the following:
```
docker-compose -f local.yml run --rm -v /d/my_project/.ipython/:/root/.ipython django python manage.py shell_plus
```

but I get a warning from Ipython:

```
The history saving thread hit an unexpected error (OperationalError('unable to open database file')).History will not be written to the database.
```

davitovmasyan:
## Description

Fixes #1794 

That's it. ¯\\\_(ツ)_/¯

## Rationale

> [IPYTHONDIR](https://ipython.org/ipython-doc/3/config/intro.html#envvar-IPYTHONDIR)
If set, this environment variable should be the path to a directory, which IPython will use for user data. 
IPython will create it if it does not exist.

> Simply install ipython into your virtualenv. [manage.py shell](https://docs.djangoproject.com/en/2.1/ref/django-admin/#shell) should use it by default.

    docker-compose -f local.yml run --rm django python manage.py shell_plus

Type and run something in shell then exit and open it again. You'll be able to browse the history.

@YAmikep @webyneter  Let me know if you have any problems with this.


CONTRIBUTORS.rst
```diff
@@ -85,6 +85,7 @@ Listed in alphabetical order.
   Daniel Hepper            `@dhepper`_                   @danielhepper
   Daniele Tricoli          `@eriol`_
   David Díaz               `@ddiazpinto`_                @DavidDiazPinto
+  Davit Tovmasyan          `@davitovmasyan`_
```
browniebroke: The `@davitovmasyan`_ is a link in Sphinx, but it's currently a broken one. You also need a line at the bottom of this file:

```
.. _@davitovmasyan: https://github.com/davitovmasyan
```

davitovmasyan: @browniebroke Fixed.

webyneter:
LGTM.

PR #1849, opened 2018-11-05
Config celery app immediately after the app being created, otherwise …

### LINKED ISSUE #1847: Celery: app.config_from_object() should be called immediately and not wait for CeleryAppConfig being ready
## What happened?

The current celery implementation configs the celery app from django settings after the CeleryAppConfig being ready.
```
app = Celery('{{cookiecutter.project_slug}}')


class CeleryAppConfig(AppConfig):
    name = '{{cookiecutter.project_slug}}.taskapp'
    verbose_name = 'Celery Config'

    def ready(self):
        # Using a string here means the worker will not have to
        # pickle the object when using Windows.
        # - namespace='CELERY' means all celery-related configuration keys
        #   should have a `CELERY_` prefix.
        app.config_from_object('django.conf:settings', namespace='CELERY')
        installed_apps = [app_config.name for app_config in apps.get_app_configs()]
        app.autodiscover_tasks(lambda: installed_apps, force=True)

```
In one project, I set the `CELERYD_CONCURRENCY` in the django settings file, however it it does not work. The reason is that when celery worker retrieves configs, the celery app is not configured yet.(https://github.com/celery/celery/blob/master/celery/bin/worker.py#L287)

The current behavior is different from the celery django [tutorial](http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html). People usually assume configs in django settings and in a separate conf fiel are equivalent, but they are not now.

The current behavior was introduced in 9bfca28f6f34affa74b950d46e7c821ace6034d4.  It did not explain why `app.config_from_object` should be put inside the ready block. Based on my own experience, I believe it is OK to config the app after the app being created immediately.

glasslion:
Fix #1847.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/taskapp/celery.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/taskapp/celery.py
@@ -12,6 +12,9 @@ if not settings.configured:
 
 
 app = Celery('{{cookiecutter.project_slug}}')
+# Using a string here means the worker will not have to
+# pickle the object when using Windows.
+app.config_from_object('django.conf:settings')
 
 
 class CeleryConfig(AppConfig):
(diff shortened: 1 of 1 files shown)


browniebroke:
Thanks for your contribution.

Could you please make sure your branch is up to date with master and resolve the conflict? 

Also, don't forget to add yourself to `CONTRIBUTORS.rst`.

Thanks again!

glasslion:
It seems that I used an old polluted fork, that cause the conflcts. I made a new PR #1860.

PR #1850, opened 2018-11-06
Modified command for Windows users

canonnervio:
Added comment to use double quotes for heroku pg:backups setup under Windows.
Current single quotes lead to (not very helpful) error message:
 »   Error: Unexpected argument: DATABASE_URL


webyneter:
Thanks. Please, add yourself to `CONTRIBUTORS.rst`.

PR #1861, opened 2018-11-12
Settings: Use security settings in dev as well as prod

### LINKED ISSUE #1827: Having production only settings leads to deployment bugs
## Description
Move all security settings out of production.py and into base.py.

## Rationale
I have been bitten multiple times by security settings that are set only in production.py that I don't become aware of until deploying. If the setting is worth setting, it should be set in base.py, so that I can learn about clickjacking, CORS, HTTP only cookies, etc. when I'm still developing, rather than after I deploy and a feature is mysteriously broken. The only exception I can think of are secure cookies, and even that's only because it's a pain to get HTTPS working locally. (It still bit me when I was trying to "debug" my non-HTTPS staging site.)

carlmjohnson:
Fixes #1827

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -70,6 +70,7 @@ Listed in alphabetical order.
   Bouke Haarsma
   Brent Payne              `@brentpayne`_               @brentpayne
   Burhan Khalid            `@burhan`_                   @burhan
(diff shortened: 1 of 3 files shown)


browniebroke:
Thanks

carlmjohnson:
🎆 🍾 💯

PR #1865, opened 2018-11-14
strip dots in default project_slug

mrcoles:

[//]: # (Before you proceed:)

## Description

This removes periods from the auto-generated project_slug (and also runs trim for good measure). 

## Rationale

Otherwise, if you have a "." in your project name (e.g., you used a website domain name as your project name), you will encounter an error about it failing for a `project_slug.isidentifier()` check.

In the old code, if you set project name to "example.com", the default for `project_slug` will appear as such:

```bash
> cookiecutter https://github.com/pydanny/cookiecutter-django
project_name [My Awesome Project]: example.com
project_slug [example.com]:
```

And if you use the default, at the end of the all the decisions you just made, you’ll encounter:

```
AssertionError: 'example.com' project slug is not a valid Python identifier.
```

This isn’t perfect, as other characters can fail too, but I’ve seen dots used enough to see value in adding this. Also, I stylistically chose to replace the dot with an empty string instead of an underscore, it seems to make more sense to me for dots.


mrcoles:
OK, sure. I amended the prior commit and pushed it (to keep the PR at 1 commit).

webyneter:
@mrcoles it's best you keep the PR history continuous -- we're squashing branch commits before merging to master anyway, no need to worry.

PR #1914, opened 2019-01-29
Hides 'sign up' elements when ACCOUNT_ALLOW_REGISTRATION is disabled

wgordon17:

[//]: # (Before you proceed:)

## Description

I am proposing an update to whether or not the "Sign Up" links are shown when `ACCOUNT_ALLOW_REGISTRATION` is disabled for a project.

## Rationale

Currently, the "Sign Up" links still show, and end up linking immediately to a "Sign up is closed" page. This is a poor UX, and it would be nice to provide this out of the box.

Umm, same as above 👍


wgordon17:
Screenshot of the proposed update to the signup page:

image

demestav:
@wgordon17 the URLs are still active though, right?

wgordon17:
@demestav You're referring to the /accounts/signup/ URL, correct? Yes, that link would still work...this just hides the visual elements and makes it much more obvious that Signup's are closed. Otherwise, user's would still see the Sign Up elements in the Nav and Sign In pages, which is just a poor UX IMO.

demestav:
@wgordon17 Yes I agree with you. I was thinking that maybe it makes sense to remove the URLs as well to keep it consistent.

wgordon17:
@demestav At the very least, we would still need the `/accounts/social/signup/` URL, because when using the social login, if the account doesn't already exist in the DB, then the user get's the Signup Closed page, which would be expected.

What's your preference? I'm good either way 👍

demestav:
It would be better to wait from one of the maintainers to tell us their preference for the project. I was just commenting in case you missed it.

wgordon17:
@demestav Agreed! Thanks for commenting 👍

browniebroke (approved):

PR #1922, opened 2019-01-31
Add setting to hide the intermediary logout page

wgordon17:

[//]: # (Before you proceed:)

## Description

I would like to avoid showing the unnecessary, intermediary confirmation signout page.

The need for an intermediary page is documented in the `django-allauth` docs: https://django-allauth.readthedocs.io/en/latest/views.html#logout-account-logout

This change will still use the intermediary logout page, but instead auto-submit the form via JavaScript. This has the added benefit of still degrading to using the form in the case that JavaScript is disabled.

The concept comes from an old Django issue: https://code.djangoproject.com/ticket/15619#comment:38

## Rationale

This is something that I believe would benefit the larger community. It's an optional feature with the default value ensuring that it does not change the expected behavior.

This hides an unnecessary, intermediary page and increases user UX by preventing clicks where not necessary.


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/context_processors.py
```diff
@@ -0,0 +1,8 @@
+from django.conf import settings
+
+
+def exposed_settings(request):
```
KrunchMuffin:
```suggestion
def expose_settings(request):
```

wgordon17: Sorry for the long lead time. This has now been addressed

BoPeng:
I have been searching for a solution for this... I am merging this to my project now.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/account/logout.html
```diff
@@ -5,18 +5,28 @@
 {% block head_title %}{% trans "Sign Out" %}{% endblock %}
 
 {% block inner %}
+
+{% if ACCOUNT_HIDE_INTERMEDIARY_LOGOUT %}
+  {# Hide form - prevents user confusion by seeing an auto-submitting form #}
+  <script>document.getElementsByClassName('container')[0].style.display = 'none';</script>
+{% endif %}
+
```
bioworkflows: The document might have multiple `container` so if there is another container (e.g. in the header) this script will fail. I would suggest wrapping the form in a div 

```
<div style='display:none'>
<form ...> </form>
</div>
```
instead.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/account/logout.html
```diff
@@ -5,18 +5,28 @@
...
 <h1>{% trans "Sign Out" %}</h1>
```
bioworkflows: If you indeed use a `display:none` `div`, this can be left out of the `div` and becomes something like

```
<h1>{% trans "Signing Out" %}</h1>
```
so that users can see a `Sinning Out` message instead of a blank page.

PR #1945, opened 2019-02-21
Change eager celery setting in local Docker

keithjeb:

[//]: # (Before you proceed:)

## Description

Added a note around CELERY_TASK_ALWAYS_EAGER = True in docker config for local development. This causes tasks to be executed on the 'main' thread rather than by the workers. I understand why that might be desirable, but thought it worth calling out incase (like me) it makes people think something is broken.

## Rationale

Ease of use/troubleshooting

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -76,7 +76,7 @@ INSTALLED_APPS += ['django_extensions']  # noqa F405
 
 # Celery
 # ------------------------------------------------------------------------------
-{% if cookiecutter.use_docker == 'n' - %}
+{% if cookiecutter.use_docker == 'n' -%}
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-always-eager
 CELERY_TASK_ALWAYS_EAGER = True
 {%- endif % }
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -79,7 +79,7 @@ INSTALLED_APPS += ['django_extensions']  # noqa F405
 {% if cookiecutter.use_docker == 'n' -%}
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-always-eager
 CELERY_TASK_ALWAYS_EAGER = True
-{%- endif % }
+{%- endif %}
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-eager-propagates
 CELERY_TASK_EAGER_PROPAGATES = True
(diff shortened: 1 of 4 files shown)


keithjeb:
If this is a holdover from the non docker dev environment and people are in agreement it shouldn't be necessary with docker happy to submit a change vs the actual code instead.

browniebroke:
Thanks for opening this. 

I agree, I can understand why we would need `CELERY_TASK_ALWAYS_EAGER = True` in bare metal as running Celery next to the web server isn't super convenient, but I don't see why we need to run Celery in eager mode with Docker.

If you feel like changing the code, that would be very much appreciated. I _think_ we need to:

- Keep using `CELERY_TASK_ALWAYS_EAGER = True` when docker option is NOT selected
- Not set it if Docker is selected, it's disabled by default
- Add the documentation you wrote to [the local development documentation](https://cookiecutter-django.readthedocs.io/en/latest/developing-locally.html#setting-up-development-environment)

Feel free to add your name to the list of contributors.

keithjeb:
@browniebroke 
Done. apologies for the multiple commits rather than one, still a noob with git in VSCode and it was easier to do on the webui than un-fubar it.

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -76,8 +76,10 @@
 
 # Celery
 # ------------------------------------------------------------------------------
+{% if cookiecutter.use_docker == 'n' - %}
```
browniebroke: The build failed as there is a syntax error here. Should not be any spaces between the `-` and the closing `%}`:

```suggestion
{% if cookiecutter.use_docker == 'n' -%}
```

keithjeb:
Doh. Sorry about that. In my defence this is actually my first ever contribution to an open-source project.

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -76,8 +76,10 @@
 
 # Celery
 # ------------------------------------------------------------------------------
+{% if cookiecutter.use_docker == 'n' -%}
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-always-eager
 CELERY_TASK_ALWAYS_EAGER = True
+{%- endif % }
```
browniebroke: Just realised that there is another extra space causing a syntax error here:

```suggestion
{%- endif %}
```

browniebroke: The rest looks good now.

soccer99:
Thank you for this change.  I was confused for over an hour about why celery seemed to be broken.  I'm still new to celery and didn't see anything wrong.

PR #1946, opened 2019-02-23
Update Gulp & other NPM dependencies

### LINKED ISSUE #1937: Upgrade to Gulp 4
There are several npm dependencies with known security issues or that aren't compatible with the latest version of browserslist or gulp (the version of cssnano this project uses is quite old and doesn't support all the browserslist queries that might be put in `package.json` now.).  It would be great to upgrade to Gulp 4 - is there interest for this from the maintainers?  I'm happy to do the work to make a pull request.

takkaria:
## Description

Upgrade to Gulp 4 and up-to-date versions of the packages the gulpfile uses.  Changes include:
* moving from `gulp-uglify` to `gulp-uglify-es` (which supports modern JS syntax)
* requiring a newer version of Node (current is 0.8) and using newer JS syntax in the gulpfile as a result
* putting the browserslist definition in package.json instead of hidden in the gulpfile

Feedback requested on:
* package.json no longer lists everything in alphabetical order.  Since JSON doesn't allow trailing commas in lists, there would have been some icky templating logic to work out where to put the final comma in the package list.  I can put this in but it seemed cleaner not to.

## Rationale

It's good to keep dependencies up to date!

Fixes #1937.

### DIFF
--- a/{{cookiecutter.project_slug}}/gulpfile.js
+++ b/{{cookiecutter.project_slug}}/gulpfile.js
@@ -1,34 +1,32 @@
+////////////////////////////////
+// Setup
+////////////////////////////////
 
-////////////////////////////////
-		//Setup//
-////////////////////////////////
+// Gulp and package
+const { src, dest, parallel, series, watch } = require('gulp')
+const pjson = require('./package.json')
 
 // Plugins
-var gulp = require('gulp'),
-      pjson = require('./package.json'),
-      gutil = require('gulp-util'),
-      sass = require('gulp-sass'),
-      autoprefixer = require('gulp-autoprefixer'),
-      cssnano = require('gulp-cssnano'),
-      {% if cookiecutter.custom_bootstrap_compilation == 'y' %}
-      concat = require('gulp-concat'),
-      {% endif %}
-      rename = require('gulp-rename'),
-      del = require('del'),
-      plumber = require('gulp-plumber'),
-      pixrem = require('gulp-pixrem'),
-      uglify = require('gulp-uglify'),
-      imagemin = require('gulp-imagemin'),
-      spawn = require('child_process').spawn,
-      runSequence = require('run-sequence'),
-      browserSync = require('browser-sync').create(),
-      reload = browserSync.reload;
-
+const autoprefixer = require('autoprefixer')
+const browserSync = require('browser-sync').create()
+{% if cookiecutter.custom_bootstrap_compilation == 'y' %}
+const concat = require('gulp-concat')
+{% endif %}
+const cssnano = require ('cssnano')
+const imagemin = require('gulp-imagemin')
+const pixrem = require('pixrem')
+const plumber = require('gulp-plumber')
+const postcss = require('gulp-postcss')
+const reload = browserSync.reload
+const rename = require('gulp-rename')
+const sass = require('gulp-sass')
+const spawn = require('child_process').spawn
+const uglify = require('gulp-uglify-es').default
 
 // Relative paths function
-var pathsConfig = function (appName) {
-  this.app = "./" + (appName || pjson.name);
-  var vendorsRoot = 'node_modules/';
+function pathsConfig(appName) {
+  this.app = "./" + (appName || pjson.name)
+  var vendorsRoot = 'node_modules/'
 
   return {
     {% if cookiecutter.custom_bootstrap_compilation == 'y' %}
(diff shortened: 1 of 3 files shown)


browniebroke:
Thanks for your contribution! I want to take some time to try this out.

> requiring a newer version of Node (current is 0.8)

I think it's fine, 0.8 we were using is quite old AFAIK, and upgrading sounds sensible (and overdue). Apparently, the current LTS is v10, so using >=8 as you did should not be a problem.

> package.json no longer lists everything in alphabetical order.

That's not a biggie IMO, NPM sorts it when running the first install, it's more important to make sure the file is valid with and without each template block.

{{cookiecutter.project_slug}}/gulpfile.js
```diff
@@ -67,72 +74,80 @@ gulp.task('styles', function() {
...
-
-});
-
-// Default task
-gulp.task('default', function() {
-    runSequence(['styles', 'scripts', {% if cookiecutter.custom_bootstrap_compilation == 'y' %}'vendor-scripts', {% endif %}'imgCompression'], ['runServer', 'browserSync', 'watch']);
-});
+function watchPaths() {
+  watch(paths.sass + '/*.scss', styles)
+  watch(paths.js + '/*.js', scripts).on("change", reload)
```
browniebroke: I'm not sure if this was a problem until now, but it looks like it's also matching the minified version, and therefore it's stuck in a loop when the `project.js` is edited:

```
[15:56:40] Finished 'scripts' after 6.81 ms
[15:56:40] Starting 'scripts'...
[15:56:40] Finished 'scripts' after 5.94 ms
[15:56:40] Starting 'scripts'...
[15:56:40] Finished 'scripts' after 6.21 ms
[15:56:41] Starting 'scripts'...
[15:56:41] Finished 'scripts' after 5.34 ms
[15:56:41] Starting 'scripts'...
[15:56:41] Finished 'scripts' after 13 ms
[15:56:41] Starting 'scripts'...
[15:56:41] Finished 'scripts' after 6.05 ms
[15:56:42] Starting 'scripts'...
[15:56:42] Finished 'scripts' after 5.44 ms
[15:56:42] Starting 'scripts'...
[15:56:42] Finished 'scripts' after 5.5 ms
[15:56:42] Starting 'scripts'...
[15:56:42] Finished 'scripts' after 5.86 ms
[15:56:43] Starting 'scripts'...
[15:56:43] Finished 'scripts' after 5.96 ms
[15:56:43] Starting 'scripts'...
[15:56:43] Finished 'scripts' after 8.68 ms
[15:56:43] Starting 'scripts'...
[15:56:43] Finished 'scripts' after 5.88 ms
```

1. Source is updated
2. Triggers watch task
3. Regenerate minified version
4. Triggers watch task
5. Regenerate minified version
6. etc...

The first argument should be changed to a list of patterns with an exclusion pattern (starts with '!') as the second one. Since we are using modern JS in this file, we can also use templated strings. Putting all together:

```suggestion
  watch([`${paths.js}/*.js`,`!${paths.js}/*.min.js`], scripts ).on("change", reload)
```

{{cookiecutter.project_slug}}/gulpfile.js
```diff
@@ -67,72 +74,80 @@ gulp.task('styles', function() {
...
+function imgCompression() {
+  return src(paths.images + '/*')
     .pipe(imagemin()) // Compresses PNG, JPEG, GIF and SVG images
-    .pipe(gulp.dest(paths.images))
-});
+    .pipe(dest(paths.images))
+}
+
+// Generate all assets
+const generateAssets = parallel(
```
browniebroke: I noticed that this task was added. By looking deeper it's replacing run-sequence, which has in its readme:

> This was intended to be a temporary solution until the release of gulp 4.0 which should have support for defining task dependencies similarly.

That looks like a nice addition to Gulp 4.0 👍 

I'm thinking that this task could be defined further down, just above the `dev` one: they are both "meta" tasks running other tasks in parallel.

takkaria: Yep I'm happy to do that; it makes sense to me either way around.

takkaria:
I hadn't seen the script issue before, which is weird.  I had seen a similar issue with image compression though, where sometimes the image watcher would get stuck in a loop, but as that doesn't put images in a different place, I've removed the 'watch' image compression.

I also exported 'generate-assets' and 'dev' as different gulp tasks.  I've found this useful when using CI, as I needed to build assets to run tests.

browniebroke:
Thanks!

PR #1949, opened 2019-03-03
Added tornado <6.0 as an explicit dependency

keithjeb:
## Description

Add Tornado <6 to dependencies.

## Rationale

Tornado 6 breaks flower 0.9.2. They have fixed their requirements on github but not released to pypi. This is a bodge round it till they do.

Newly built docker images will pickup tornado 6.0 and therefore flower doesn't run. This might be better as an issue incase upstream is fixed, but they haven't released since 2017 so maybe not quick.

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -13,6 +13,7 @@ redis>=2.10.6, < 3  # pyup: < 3 # https://github.com/antirez/redis
 celery==4.2.1  # pyup: < 5.0  # https://github.com/celery/celery
 {%- if cookiecutter.use_docker == 'y' %}
 flower==0.9.2  # https://github.com/mher/flower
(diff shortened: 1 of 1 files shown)


cesarmanzo:
I was about to commit this, good job!

luzfcb:
@keithjeb please, can you add yourself in the https://github.com/pydanny/cookiecutter-django/blob/master/CONTRIBUTORS.rst

keithjeb:
@luzfcb I'm already there ;)

browniebroke:
Looks good, thanks for this. 

They plan to do a new release [in a few days](https://github.com/mher/flower/issues/878#issuecomment-469195853). I'll wait for next week before merging this, if there is no new release by then, we'll implement this in the template. Until then, people with the problem can implement in their generated project.

browniebroke:
A new version of Flower [has been released](https://pypi.org/project/flower/0.9.3/) with the fix. 

We should receive a pull request from Pyup soon with the fix. Strange that there is still nothing ~2 days after though... We should look into it.

browniebroke:
Closing as we've upgraded to the latest version of Flower with the fix.

Thanks for the report!

PR #1951, opened 2019-03-06
Add automatic migrations to heroku deploys

yunti:
heroku now has a new feature for running tasks as part of deployment. Perfect for automatic migrations.
https://devcenter.heroku.com/articles/release-phase#specifying-release-phase-tasks

### DIFF
--- a/docs/deployment-on-heroku.rst
+++ b/docs/deployment-on-heroku.rst
@@ -47,7 +47,6 @@ Run these commands to deploy the project to Heroku:
 
     git push heroku master
 
-    heroku run python manage.py migrate
     heroku run python manage.py createsuperuser
     heroku run python manage.py collectstatic --no-input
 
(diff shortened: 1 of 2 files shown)


browniebroke:
I think it's a good addition. However, I think we should also update the docs, we don't need to mention migrations in this page:

https://github.com/pydanny/cookiecutter-django/blob/master/docs/deployment-on-heroku.rst#L50

yunti:
Done see #1955

browniebroke:
Can you please add it to this same pull request? There is no need to create a new one...

browniebroke:
...Alternatively, you can directly edit the file on the right branch in the Github UI:

https://github.com/yunti/cookiecutter-django/edit/patch-7/docs/deployment-on-heroku.rst

yunti:
Thanks, I didn’t manage to move the other git before you.  Thanks for finishing it off

PR #1955, opened 2019-03-11
Update deployment docs for heroku

yunti:
As per #1951

Update heroku docs to reflect automatic migrations in procfile


browniebroke:
As per the other PR, please update the existing branch rather than create a new pull request. 
Thanks.

Issue #591, opened 2016-06-05
Write test that creates/checks a matrix of all y/n option combinations


audreyr:
Ensure that the project generates without errors for every possible combination of:
- y/n values entered
- choices entered when selecting from a list, e.g. test all license choices

(Similarly filed in Cookiecutter PyPackage: https://github.com/audreyr/cookiecutter-pypackage/issues/184)

jeffshek:
@audreyr i'll give this a shot, i'll start with the cc-django one and port it over if it's what we're going for

jeffshek:
Sorry, it has been stupid busy, but I have this bookmarked and every few days I remind myself of the guilt that I didn't do this yet! I'm going to take a look into this week to give it a shot (if not already done)

webyneter:
@jeffshek, any news on that?

kenlavoie:
@browniebroke - is this still an open issue?

PR #1957, opened 2019-03-11
Run Black on Travis

### LINKED ISSUE #1603: Teach Travis to black Cookiecutter Django code
[black VCS integration](https://github.com/ambv/black#version-control-integration).

x-ref #1602.

### LINKED ISSUE #1866: mypy not found when running tests in docker
## What happened?
When running `sh tests/test_docker.sh` as per the documentation, the following error shows up:
> /entrypoint.sh: exec: line 44: mypy: not found

Note that after the error, the process proceeds, it does not break.

## Steps to reproduce
Just run `sh tests/test_docker.sh`.

browniebroke:
## Description

Run black on the CI to make sure template scripts, as well as generated projects, are following the Black code style, which fix #1603.

I reworked the Travis build matrix to run things in parallel as well as test docker with and without Celery, this should help our coverage (refs #591).

I also added an option to the bash script to make sure it fails the whole script as soon as one command fails. Should fix #1866.

## Rationale

This is to make sure we don't introduce any regressions when adding things to the project, it should stick to black style guide.

Some builds in previous commits failed and therefore demonstrate that violations are properly spotted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,10 @@ before_install:
 
 matrix:
   include:
-    - name: Tox test
+    - name: Test
       script: tox -e py36
+    - name: Black
+      script: tox -e black
     - name: Basic Docker
       script: sh tests/test_docker.sh
     - name: Docker with Celery
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,16 +7,18 @@ language: python
 
 python: 3.6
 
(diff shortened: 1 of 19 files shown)


browniebroke:
Unless there is any objections from the other maintainers, I'm planning on merging this in a few days.

Issue #1954, opened 2019-03-08
Kombu==4.4.0 breaks celery / redis implementation Docker


rich5851:
## What happened before?

Celery Workers started 

## What happens now?

kombu==4.4.0 released on the 3rd March 2019

Celery Workers fail to start with VersionMismatch
celeryworker_1  | kombu.exceptions.VersionMismatch: Redis transport requires redis-py versions 3.2.0 or later. You have 2.10.6

Updating to redis 3.2 not an option as it causes django_celery_beat that I am also using to no longer change the periodic tasks file when tasks are updated.

## Steps to reproduce
Deployment local or production with docker
docker compose build and up

Fixed the issue by adding kombu==4.3.0 at bottom of requirements/base.txt

rolep:
Then better to use kombu==4.4.0 cause 4.3.0 also has bug which I have encoutered: celery/kombu#1006
The bug is fixed in 4.4.0

Also am using django_celery_beat, but with database scheduler, seemingly with no such issues.

If using redis 2.x is preferred - then better stick with previous kombu version 4.2.xxx, but imho thats not in spirit of this project

PR #1869, opened 2018-11-18
 Pinning redis version to >=2.10.6, <3

Egregors:
Fixes #1868

Issue #1868, opened 2018-11-18
New version of redis package is the cause of the Error


Egregors:
## What happened?

After redeploy, we got `redis==3.0.1` instead `redis==2.10.6`. So, when you try to work with cache for example, we got the exception: 

```
DataError: Invalid input of type: 'CacheKey'. Convert to a byte, string or number first.
  File "django/core/handlers/exception.py", line 35, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 128, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 126, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "django/utils/decorators.py", line 134, in _wrapped_view
    result = middleware.process_request(request)
  File "django/middleware/cache.py", line 133, in process_request
    cache_key = get_cache_key(request, self.key_prefix, 'GET', cache=self.cache)
  File "django/utils/cache.py", line 360, in get_cache_key
    headerlist = cache.get(cache_key)
  File "django_redis/cache.py", line 32, in _decorator
    return method(self, *args, **kwargs)
  File "django_redis/cache.py", line 81, in get
    client=client)
  File "django_redis/client/default.py", line 202, in get
    value = client.get(key)
  File "redis/client.py", line 1207, in get
    return self.execute_command('GET', name)
  File "redis/client.py", line 754, in execute_command
    connection.send_command(*args)
  File "redis/connection.py", line 619, in send_command
    self.send_packed_command(self.pack_command(*args))
  File "redis/connection.py", line 659, in pack_command
    for arg in imap(self.encoder.encode, args):
  File "redis/connection.py", line 124, in encode
    "byte, string or number first." % typename)
```
The problem exactly in the `redis==3.0.1`, as we may see there: https://stackoverflow.com/questions/53331405/django-compress-error-invalid-input-of-type-cachekey

It is because of this pkg sets in the `requiriments/base.txt` as `redis>=2.10.5  # https://github.com/antirez/redis`

## What should've happened instead?

All should work well

## Steps to reproduce

Just rebuild compose. 

(Ubuntu 18.04, Docker 17.05.0-ce)

browniebroke:
Hi, thanks for the report. I believe it was fixed upstream in `django-redis`. I've just merged #1870 which should solve this.

[...]

PR #1958, opened 2019-03-12
Redis and Kombu update

rolep:

[//]: # (Before you proceed:)

## Description

Use recent versions or redis==3.2.0
and kombu==4.4.0

Fixes: pydanny/cookiecutter-django#1954

## Rationale

Previously redis version was pinned by pydanny/cookiecutter-django#1869
due to bug pydanny/cookiecutter-django#1868

Related bug in kombu celery/kombu#947
was fixed at least in 4.3.0

kombu 4.3.0 although contains another bug celery/kombu#1006
fixed in 4.4.0

kombu 4.4.0 strictly requires redis>=3.2.0

Pinning to older version of kombu / redis does not feel to be in the spirit 
of this project, and older redis version was pinned due to bug in kombu which
was fixed

New installs will get recent versions of kombu which require redis 3.2.0

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -11,6 +11,7 @@ whitenoise==4.1.2  # https://github.com/evansd/whitenoise
 redis>=3.2.0  # https://github.com/antirez/redis
 {%- if cookiecutter.use_celery == "y" %}
 celery==4.2.1  # pyup: < 5.0  # https://github.com/celery/celery
+kombu==4.4.0  # https://github.com/celery/kombu
 {%- if cookiecutter.use_docker == 'y' %}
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/requirements/base.txt
```diff
@@ -8,9 +8,10 @@ argon2-cffi==19.1.0  # https://github.com/hynek/argon2_cffi
 {%- if cookiecutter.use_whitenoise == 'y' %}
 whitenoise==4.1.2  # https://github.com/evansd/whitenoise
 {%- endif %}
-redis>=2.10.6, < 3  # pyup: < 3 # https://github.com/antirez/redis
+redis>=3.2.0  # https://github.com/antirez/redis
```
browniebroke:
```suggestion
redis>=3.2.1  # https://github.com/antirez/redis
```

PR #1993, opened 2019-03-25
Test all possible template combinations

### LINKED ISSUE #591: Write test that creates/checks a matrix of all y/n option combinations

browniebroke:
## Description

Fixes #591.

## Rationale

We are currently not testing many combinations, we run Flake8 on the generated project with default options, but that rarely catch any issues.

Catch problems with invalid combinations, for instance, it fails due to Whitenoise being incompatible with `django-compressor` because `STATIC_URL` was undefined here:

https://github.com/pydanny/cookiecutter-django/blob/b91c70d75578e56c2ca74400684c5b5523cd0cd4/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L185


luzfcb:
hey @browniebroke, thanks a lot to make this pull-request.

tests/test_cookiecutter_generation.py
```diff
@@ -60,33 +93,49 @@ def test_default_configuration(cookies, context):
...
+@pytest.mark.parametrize("windows", BINARY_CHOICES)
+@pytest.mark.parametrize("use_docker", BINARY_CHOICES)
+@pytest.mark.parametrize("use_celery", BINARY_CHOICES)
+@pytest.mark.parametrize("use_mailhog", BINARY_CHOICES)
+@pytest.mark.parametrize("use_sentry", BINARY_CHOICES)
+@pytest.mark.parametrize(
+    # These 2 cannot be used together, but test the other combinations
+    ["use_compressor", "use_whitenoise"],
+    [("y", "n"), ("n", "n"), ("n", "n")],
+)
```
browniebroke: I'm not really satisfied with this duplication of parameters. If we want to avoid it, it looks like we need to combine [fixtures and parametrization](https://docs.pytest.org/en/latest/fixture.html#fixture-parametrize) or use the [`pytest_generate_tests` hook](https://docs.pytest.org/en/latest/parametrize.html#pytest-generate-tests).

PR #2016, opened 2019-04-02
Refactor Celery integration according to current best practices

### LINKED ISSUE #865: Changes on celery implementation
According to [celery docs](http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html#using-celery-with-django):

> Note that this example project layout is suitable for larger projects, for simple projects you may use a single contained module that defines both the app and tasks, like in the First Steps with Celery tutorial.

With that in mind, the fact that the celery setup file is inside of a folder in the path where we hold our apps is a little misleading. It is configured so it discovers tasks with the @shared_task decorator inside of any app, but it is inside of a folder named taskapp, again, misleading.

Another thing that i would like to point is that celery.py is a setup file, where we create the instance for the celery application. In that way it is very similar to the wsgi.py file, where we setup the wsgi app. Different things, i know, but with the same purpose.

Following that line of thought i would like to propose removing the taskapp and move the celery.py file (and the changes to **init**.py) to the config folder. Simpler structure and a more coherent setup in my opinion.

browniebroke:
## Description

- Change celery app to not be a Django app, more like a WSGI app
- Define a Celery task in the Django users app
- Write a test to execute the task
- Update scripts to use the new app to start workers
- Update documentation

## Rationale

Fix #865

Follow the Celery docs on how to [integrate with Django](http://docs.celeryproject.org/en/stable/django/first-steps-with-django.html)


sfortunato:
Hello! Very excited for this refactor! 

Not sure if this is kosher, but took a look at the diffs of this PR and implemented them on a project I'd kicked off from cookiecutter without celery integration. In order to run the celery worker from the root dir (same location as manage.py - as per the updated README in this PR) I had to add the following lines from manage.py to celery_app.py.

`current_path = os.path.dirname(os.path.abspath(__file__))
sys.path.append(os.path.join(current_path, '..', "vessel_health"))`

May only be necessary for my frankenstein implementation - but thought I'd share.

browniebroke:
@sfortunato thanks for the feedback. I don't think I can reproduce your issue, I tried to deploy a version generated from this pull request on Heroku and I was able to start the Celery worker & get the dummy task executed asynchronously. I assume your problem is related to your project, but hard to tell without seeing the whole thing...

Unless someone else has any further objections or want more time for review, I'll merge this in the next few days.

cdvv7788:
Finally. No more manual changes to the base cookiecutter to have this structure 👍 .
Good job guys!

PR #2017, opened 2019-04-02
Upgrade to Django 2.1

### LINKED ISSUE #1654: Things to keep in mind before upgrading to Django 2.1
* [UserCreationForm and UserChangeForm no longer need to be rewritten for a custom user model](https://docs.djangoproject.com/en/dev/releases/2.1/#django-contrib-auth) (I believe, this has already been pointed out by some around here but could find any mention though...)

browniebroke:
## Description

Although 2.2 has just been released, we should do this long-awaited upgrade first, as the 3rd party libs we use might take a bit of time to be compatible. 

I've implemented a few changes suggested in #1654, but I couldn't find the simplifications for user forms.

## Rationale

Fixes #1654


luzfcb:
I'm hoping I can get, sometime in the near future, some free weekend to review this and the other pull-requests

luzfcb:
we need to create a git tag before merge this

foarsitter:
Are there any specific features need to be tested? I installed this branch with cookiecutter and it looks good to me. Besides, I'm using 2.1 already in production for a while. 

The simplification of the user forms suggested in #1654 can be done in a different pull request I suppose.  After playing around a bit with the two forms it looks like there isn't any difference.

browniebroke:
I'd say we should test sign-up/sign-in and that the admin work. I did a quick check and it seems to work fine. I was able to sign-up, verify my account & then sign-in. I've navigated the admin and performed some changes without a problem.

I agree that the simplifications can be done later.

vegerta:
Hi guys,
Is there any progress in here?

browniebroke:
Good point. @luzfcb did you get a chance to take a look?

Unless anyone has any objection, I propose we merge this and get to the next step with the 2.2 upgrade. I should have a bit of free time in the coming days in case of problems.

Issue #2004, opened 2019-03-29
Docs are out of date


browniebroke:
## What happened?

The docs are not being updated anymore. It looks like RTD hasn't been building [in the past 3 months](https://readthedocs.org/projects/cookiecutter-django/builds/). Looks like it's due to a change in how they receive webhooks: https://blog.readthedocs.com/webhook-deprecations/

We'll need to update the webhooks in the repo settings to the new URL.

browniebroke:
I don't have access to update the webhook, so in the meantime, I've deployed the latest version here: https://cookiecutter-django-mirror.readthedocs.io/en/latest/index.html

sfdye:
![image](https://user-images.githubusercontent.com/1016390/56255107-2aee6a80-60f6-11e9-9f25-705bd5d26c79.png)
Fixed by @luzfcb

PR #2027, opened 2019-04-08
Use Netlify to host the docs

browniebroke:
## Description

Experiement to use [Netlify](https://www.netlify.com/) to build the docs

## Rationale

The documentation is currently hosted on ReadTheDocs. This worked great until they updated their webhooks APIs a few months ago which stopped the docs from being updated (see #2004).

While this can be fixed easily, Netlify gives a very nice feature that ReadTheDocs currently lacks: deploy previews on each pull request. When docs need to be updated, we can check that it builds and it renders well.

This is the version deployed from my fork: https://frosty-galileo-8b7cac.netlify.com/

## Problems

I'm not sure how people feel about moving away from a community supported project like RTD to a platform like Netlify. While they have this nicer feature, I have mixed feeling about it.

If we were to transition, we would also need to make sure the old location redirects people to the new one.

Thoughts on this?


sfdye:
I host with my blog with netlify, but I prefer docs on RTD though. We can ask @pydanny to grant us the admin access on RTD, so that we can fix the webhooks.

PR #2047, opened 2019-05-06
 Edit STATIC and MEDIA urls for GCE production.

cmargieson:

[//]: # (Before you proceed:)

## Description

Edit STATIC and MEDIA urls in production settings to include a trailing / when Google Cloud Platform is used as a storage service.

## Rationale

An error occurs on deployment if the / is not included in the environment variable. (Tested on Heroku)

tox gives cake, unsurprisingly.


browniebroke:
Thnaks for the fix 🎉

PR #2059, opened 2019-05-17
Add LocaleMiddleware

tanoabeleyra:
As said couple years ago in #792:

> The settings comes with `USE_I18N` and `USE_L10N` set as `True`, but we need to add `django.middleware.locale.LocaleMiddleware` to make the localization and internationalization work.

The middleware was added here (https://github.com/pydanny/cookiecutter-django/commit/464fb56263a4c5de39d6d333250f797c96a1265a) but it seems that it wasn't merged (https://github.com/pydanny/cookiecutter-django/pull/793).

Is any reason for not adding the middleware by default?


tanoabeleyra:
I also added [LOCALE_PATHS](https://docs.djangoproject.com/en/dev/ref/settings/#locale-paths)

tanoabeleyra:
Thank you!

PR #2066, opened 2019-05-18
traefik toml config - storageFile is deprecated, use storage instead

mknapper1:

[//]: # (Before you proceed:)

## Description

The `storageFile` configuration variable for ACME has been deprecated, and replaced with `storage`

## Rationale

This pull request implements the updated configuration variable name and stops the warning from displaying when starting a cookiecutter-django project that uses Docker.


browniebroke:
Thanks for the fix! 🎉

PR #2067, opened 2019-05-18
Make cloud_provider optional

tanoabeleyra:
Some projects may not need AWS or GCS.

Also use `GCS` as [Google Cloud Storage](https://cloud.google.com/storage/) acronym everywhere (don't alternate between GCE and GCS).

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -33,7 +33,8 @@
   ],
   "cloud_provider": [
     "AWS",
-    "GCE"
+    "GCE",
+    "None"
   ],
   "custom_bootstrap_compilation": "n",
   "use_compressor": "n",
--- a/README.rst
+++ b/README.rst
@@ -185,7 +185,8 @@ Answer the prompts with your own desired options_. For example::
     Select cloud_provider:
     1 - AWS
     2 - GCS
-    Choose from 1, 2 [1]: 1
+    3 - None
+    Choose from 1, 2, 3 [1]: 1
     custom_bootstrap_compilation [n]: n
     Select open_source_license:
     1 - MIT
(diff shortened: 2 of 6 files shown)


browniebroke:
Thanks for this!

I'm not sure about the use case for no cloud provider, Static could be optional with Whitenoise (once #2068 is merged), but how would media storage work?

For the GCE vs GCS naming, there is indeed some inconsistencies. GCE is definitely not appropriate, but now that you raise this, GCS is the storage service of Google. Maybe we should use GCP instead? It would be more in line with what AWS stands for on the Amazon side of things, we don't call it `S3`...

PS: I generally prefer separate pull requests for separate concerns, one of these 2 changes is welcome but the other requires a bit more discussion... Right now, both are on hold, while one could be merged if part of a separate pull request.

tanoabeleyra:
Hi Bruno,

You're right, there will be no media serving if `"None"` was selected, but maybe the project doesn't need it or the user wants to use a different storage. I think this make the config more flexible and doesn't harm those who are using AWS/GCP.

I agree with using **GCP** as acronym, I will open another PR for that.

cookiecutter.json
```diff
@@ -32,6 +32,7 @@
     "Gulp"
   ],
   "cloud_provider": [
+    "None",
```
browniebroke: I think this will make it the default option, wouldn't it? We usually add new options at the end of the list.

Can you change this and the associated pieces of doc please?

tanoabeleyra: Sure, I'll change it

tanoabeleyra:
Here we go, I made the proposed changes and passed every test.

tanoabeleyra:
Hi @browniebroke, any update on this?

PR #2071, opened 2019-05-20
1989 - reworked postgres version options

### LINKED ISSUE #1989: Add PostgreSQL 11 as an option
## Description

It's been out for a few months. I also suggest removing older version of PostgreSQL 10, and just keep the latest version. They changed their [versioning policy in v10](https://www.postgresql.org/support/versioning/), which means that 10.1 vs 10.2 is the same type of change as 9.6.1 vs 9.6.2 was. 

## Rationale

We should provide recent versions, and not overload with minor versions.

Update options in `cookiecutter.json`, and documentation where applicable.

curtisstpierre:

[//]: # (Before you proceed:)

## Description

This is a change to address #1989

## Rationale

As is pointed out in this ticket, the new postgres versioning is different, so removing older 10.x versions makes sense as well as adding postgres 11.3

Fixes #1989


README.rst
```diff
@@ -169,14 +169,11 @@ Answer the prompts with your own desired options_. For example::
...
-    5 - 10.1
-    6 - 9.6
-    7 - 9.5
-    8 - 9.4
+    1 - 11.3
+    2 - 10.8
+    3 - 9.6
+    4 - 9.5
+    5 - 9.4
     Choose from 1, 2, 3, 4, 5, 6, 7, 8 [1]: 1
```
browniebroke:
```suggestion
     Choose from 1, 2, 3, 4, 5 [1]: 1
```

CONTRIBUTORS.rst
```diff
@@ -7,19 +7,20 @@ Core Developers
...
-Burhan Khalid               `@burhan`_       @burhan
-Nikita Shupeyko             `@webyneter`_    @webyneter
-Bruno Alla                  `@browniebroke`_ @_BrunoAlla
-Wan Liuyang                 `@sfdye`_        @sfdye
-=========================== ================ ===========
+Burhan Khalid               `@burhan`_         @burhan
+Nikita Shupeyko             `@webyneter`_      @webyneter
+Bruno Alla                  `@browniebroke`_   @_BrunoAlla
+Wan Liuyang                 `@sfdye`_          @sfdye
+Curtis St Pierre            `@curtisstpierre`_ @cstpierre1388
```
browniebroke: This list is the core maintainers who can merge pull requests and commit to the repo. All the contributors are listed in the following list, that where you should add your name.

Also, the second column is [a reStrusturedText hyperlink](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks), you should also define its target somewhere in the file, we do that at the bottom of it, for instance:

https://github.com/pydanny/cookiecutter-django/blob/98012829f3c5a2dbfa7962f744590685513de956/CONTRIBUTORS.rst#L195

README.rst
curtisstpierre: thanks for catching this. It's fixed in the newest commit.

CONTRIBUTORS.rst
curtisstpierre: Fixed in the newest commit, and added the link. Thank you!

PR #2075, opened 2019-05-21
Log errors are sent to Sentry

### LINKED ISSUE #2073: Improve logging to Sentry
## Description
In production, when something unexpected happens we should be notified. Since there is already Sentry integration in place, it would be good to be able to utilize the logging functionality of Sentry.

According to [Sentry documentation](https://docs.sentry.io/platforms/python/logging/ ) logging should be enabled by default. Maybe I am missing something but using the following code, we should be able to log it to Sentry.
```python
import logging
logging.error("I am an event")
```

I tried it, but I couldn't get it to work, probably because I am failing to understand how the loggers configuration in cookiecutter-django.

demestav:
## Description
Fixes #2073 

## Rationale
It is efficient to utilize an already existing reporting channel (Sentry) for logging errors.

### DIFF
--- a/docs/deployment-with-docker.rst
+++ b/docs/deployment-with-docker.rst
@@ -35,7 +35,15 @@ Configuring the Stack
 
 The majority of services above are configured through the use of environment variables. Just check out :ref:`envs` and you will know the drill.
 
-To obtain logs and information about crashes in a production setup, make sure that you have access to an external Sentry instance (e.g. by creating an account with `sentry.io`_), and set the ``SENTRY_DSN`` variable.
+To obtain logs and information about crashes in a production setup, make sure that you have access to an external Sentry instance (e.g. by creating an account with `sentry.io`_), and set the ``SENTRY_DSN`` variable. Logs of level `logging.ERROR` are sent as Sentry events. Therefore, in order to send a Sentry even use:
+
+.. code-block:: python
+
+    import logging
+    logging.error("This event is sent to Sentry", extra={"<example_key>": "<example_value>"})
+
+The `extra` parameter allows you to send additional information about the context of this error.
+
 
 You will probably also need to setup the Mail backend, for example by adding a `Mailgun`_ API key and a `Mailgun`_ sender domain, otherwise, the account creation view will crash and result in a 500 error when the backend attempts to send an email to the account owner.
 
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -288,7 +288,7 @@ class MediaRootS3Boto3Storage(S3Boto3Storage):
 
 sentry_logging = LoggingIntegration(
     level=SENTRY_LOG_LEVEL,  # Capture info and above as breadcrumbs
-    event_level=None,  # Send no events from log messages
+    event_level=logging.ERROR  # Send errors as events
```
browniebroke: Build is failing the Black code formatting due to the missing comma. I wish the output was better, but this should fix it:

```suggestion
    event_level=logging.ERROR,  # Send errors as events
```

docs/deployment-with-docker.rst
```diff
@@ -35,7 +35,15 @@ Configuring the Stack
 
 The majority of services above are configured through the use of environment variables. Just check out :ref:`envs` and you will know the drill.
 
-To obtain logs and information about crashes in a production setup, make sure that you have access to an external Sentry instance (e.g. by creating an account with `sentry.io`_), and set the ``SENTRY_DSN`` variable.
+To obtain logs and information about crashes in a production setup, make sure that you have access to an external Sentry instance (e.g. by creating an account with `sentry.io`_), and set the ``SENTRY_DSN`` variable. Logs of level `logging.ERROR` are sent as Sentry events. Therefore, in order to send a Sentry even use:
```
browniebroke:
```suggestion
To obtain logs and information about crashes in a production setup, make sure that you have access to an external Sentry instance (e.g. by creating an account with `sentry.io`_), and set the ``SENTRY_DSN`` variable. Logs of level `logging.ERROR` are sent as Sentry events. Therefore, in order to send a Sentry event use:
```

{{cookiecutter.project_slug}}/config/settings/production.py
demestav: Thank you for the correction. 

For future reference, how did you spot the error? Going to Travis shows many failures all around. Searching for the code (line 291) I couldn't find it in the report.

docs/deployment-with-docker.rst
demestav: Thank you for the correction.

{{cookiecutter.project_slug}}/config/settings/production.py
browniebroke: I simply knew how Black wanted this to be formatted by experience. The output is pretty bad for this in Travis, if you know way to improve it, help is welcome.

browniebroke:
Thanks 🎉

Issue #2070, opened 2019-05-20
docker-compose -f local.yml up fails on windows


shireenrao:
## What happened?

I am trying to start a new project on windows using docker. Here are my settings

```
cookiecutter https://github.com/pydanny/cookiecutter-django
You've downloaded C:\Users\srao.cookiecutters\cookiecutter-django before. Is it okay to delete and re-download it? [yes]: yes
project_name [My Awesome Project]:
project_slug [my_awesome_project]:
description [Behold My Awesome Project!]:
author_name [Daniel Roy Greenfeld]:
domain_name [example.com]:
email [daniel-roy-greenfeld@example.com]:
version [0.1.0]:
Select open_source_license:
1 - MIT
2 - BSD
3 - GPLv3
4 - Apache Software License 2.0
5 - Not open source
Choose from 1, 2, 3, 4, 5 (1, 2, 3, 4, 5) [1]:
timezone [UTC]:
windows [n]: y
use_pycharm [n]:
use_docker [n]: y
Select postgresql_version:
1 - 10.5
2 - 10.4
3 - 10.3
4 - 10.2
5 - 10.1
6 - 9.6
7 - 9.5
8 - 9.4
Choose from 1, 2, 3, 4, 5, 6, 7, 8 (1, 2, 3, 4, 5, 6, 7, 8) [1]:
Select js_task_runner:
1 - None
2 - Gulp
Choose from 1, 2 (1, 2) [1]:
Select cloud_provider:
1 - AWS
2 - GCE
Choose from 1, 2 (1, 2) [1]:
custom_bootstrap_compilation [n]:
use_compressor [n]:
use_celery [n]:
use_mailhog [n]: y
use_sentry [n]:
use_whitenoise [n]:
use_heroku [n]:
use_travisci [n]:
keep_local_envs_in_vcs [y]:
debug [n]:
[SUCCESS]: Project initialized, keep up the good work!
```

The build succeeds with `docker-compose -f local.yml build` and the django_1 container fails after I execute `docker-compose -f local.yml up`
The error is

```
django_1 | Running on http://0.0.0.0:8000/ (Press CTRL+C to quit)
django_1 | Restarting with stat
': No such file or directorycute 'python
my_awesome_project_django_1 exited with code 127
```

I am on Docker version 18.09.2, build 6247962 and docker-compose version 1.23.2, build 1110ad01
Any ideas on what I am doing wrong?

## Steps to reproduce

[//]: # (Any or all of the following:)
[...]

PR #2076, opened 2019-05-21
Revert Werkzeug back to 0.14

shireenrao:
reverting back Werkzeug to version 0.14 based on discussion on #2070
I did the change locally on my windows laptop and can confirm that this is now working.

[//]: # (Before you proceed:)

## Description

## Rationale

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/local.txt
+++ b/{{cookiecutter.project_slug}}/requirements/local.txt
@@ -1,6 +1,6 @@
 -r ./base.txt
 
-Werkzeug==0.15.4  # https://github.com/pallets/werkzeug
+Werkzeug==0.14  # https://github.com/pallets/werkzeug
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/requirements/local.txt
```diff
@@ -1,6 +1,6 @@
 -r ./base.txt
 
-Werkzeug==0.15.4  # https://github.com/pallets/werkzeug
+Werkzeug==0.14  # https://github.com/pallets/werkzeug
```
browniebroke: Can we use 0.14.1 instead? Also, we would want to prevent pyup from updating to 0.15 and above by adding a filter:

```suggestion
Werkzeug==0.14.1  # pyup: < 0.15  # https://github.com/pallets/werkzeug
```

browniebroke:
Thanks

PR #2084, opened 2019-05-30
Use django-celery-beat for scheduled tasks

keyvanm:

[//]: # (Before you proceed:)

## Description

[`django-celery-beat`](https://pypi.org/project/django-celery-beat/) is an extension enables you to store the periodic task schedule in the database.

The periodic tasks can be managed from the Django Admin interface, where you can create, edit and delete periodic tasks and how often they should run.

## Rationale

`django-celery-beat` allows for an easy interface to define periodic tasks. There already is a docker-compose service defined for `django-celery-beat`, but celery isn't configured to read 

Configuring scheduled tasks using celery and celery beat in Django admin, and running them in docker.

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/local/django/celery/beat/start
+++ b/{{cookiecutter.project_slug}}/compose/local/django/celery/beat/start
@@ -5,4 +5,4 @@ set -o nounset
 
 
 rm -f './celerybeat.pid'
-celery -A config.celery_app beat -l INFO
+celery -A config.celery_app beat -l INFO --scheduler django_celery_beat.schedulers:DatabaseScheduler
--- a/{{cookiecutter.project_slug}}/compose/local/django/celery/beat/start
+++ b/{{cookiecutter.project_slug}}/compose/local/django/celery/beat/start
@@ -5,4 +5,4 @@ set -o nounset
 
 
 rm -f './celerybeat.pid'
-celery -A config.celery_app beat -l INFO --scheduler django_celery_beat.schedulers:DatabaseScheduler
+celery -A config.celery_app beat -l INFO
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -75,6 +75,9 @@ THIRD_PARTY_APPS = [
     "allauth.account",
     "allauth.socialaccount",
     "rest_framework",
+{% if cookiecutter.use_celery == 'y' -%}
+    "django_celery_beat",
+{%- endif %}
 ]
 LOCAL_APPS = [
     "{{ cookiecutter.project_slug }}.users.apps.UsersConfig",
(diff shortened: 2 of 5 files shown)


{{cookiecutter.project_slug}}/compose/local/django/celery/beat/start
```diff
@@ -5,4 +5,4 @@ set -o nounset
 
 
 rm -f './celerybeat.pid'
-celery -A config.celery_app beat -l INFO
+celery -A config.celery_app beat -l INFO --scheduler django_celery_beat.schedulers:DatabaseScheduler
```
browniebroke: That scheduler comes from the `django_celery_beat` extension published as [an external package](https://pypi.org/project/django-celery-beat/) which isn't installed in cookiecutter-django. There are a couple more things to do [according to the documentation](http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#using-custom-scheduler-classes).

Personally, I would prefer to use the [Django setting](http://docs.celeryproject.org/en/latest/userguide/configuration.html?highlight=CELERYBEAT_SCHEDULER) rather than the command line option, this would avoid non-Docker deployments to forget this (like you forgot the Heroku deployment).

keyvanm: Thanks for brining up the Django setting. Didn't know it existed.

Should we add a cookiecutter config option to install `django_celery_beat`? Or assume that someone who wants celery also wants beat?

browniebroke: > Should we add a cookiecutter config option to install `django_celery_beat`?

I don't think it's required, sounds a bit overkill.

{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -76,6 +76,11 @@
     "allauth.socialaccount",
     "rest_framework",
 ]
+# http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#using-custom-scheduler-classes
+{% if cookiecutter.use_celery == 'y' -%}
+THIRD_PARTY_APPS += ["django_celery_beat"]
+{%- endif %}
```
browniebroke: Do you know why it failed when you added it to the list above directly?

keyvanm: Not sure. I think the file in the generated project didn't have the right formatting.

browniebroke: Yea, our error reporting is really bad on CI at the moment, I will try to improve this... It has to do with the position of the dash in the Jinja tags. You wrote:

```
    "rest_framework",
{% if cookiecutter.use_celery == 'y' -%}
    "django_celery_beat",
{%- endif %}
]
```

Which is compiling to:

```
    "rest_framework",
"django_celery_beat",
]
```

Instead of:

```
    "rest_framework",
    "django_celery_beat",
{%- endif %}
]
```

Which is compiling to, and makes the linters happy:

```
    "rest_framework",
    "django_celery_beat",
]
```

keyvanm: Gotcha! I was wondering what the dashes do. Fixed in the new commit

browniebroke:
Tanks 🎉 !!

PR #2122, opened 2019-06-20
Remove redundant template debug setting

danihodovic:
> 'debug': a boolean that turns on/off template debug mode. If it is
> True, the fancy error page will display a detailed report for any
> exception raised during template rendering. This report contains the
> relevant snippet of the template with the appropriate line highlighted.

> It defaults to the value of the DEBUG setting.

https://docs.djangoproject.com/en/2.2/topics/templates/#module-django.template.backends.django

I could be wrong about this, but it seems like setting the template
DEBUG setting is redundant, since it should follow whatever the DEBUG
variable is set to.


browniebroke:
Good find! Thanks for the simplification 🎉

PR #2125, opened 2019-06-24
Add django-health-check

danihodovic:

[//]: # (Before you proceed:)

## Description

[django-health-check](https://github.com/KristianOellegaard/django-health-check) offers a health check  so that you can configure automatic probes to your app and ensure things are up and running at all times.

It supports:
- cache
- database
- storage
- disk and memory utilization (via psutil)
- AWS S3 storage
- Celery task queue
- RabbitMQ

I'm unsure if I need to test this by automatically deploying the app, so please let me know.

```
$ https https://findwork.dev/health/ --json
HTTP/1.1 200 OK

{
    "DatabaseBackend": "working"
}
```

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -73,6 +73,7 @@ SECURE_CONTENT_TYPE_NOSNIFF = env.bool(
 INSTALLED_APPS += ["storages"]  # noqa F405
 {%- endif -%}
 {% if cookiecutter.cloud_provider == 'AWS' %}
+INSTALLED_APPS += ["health_check.contrib.s3boto_storage"]
 # https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
 AWS_ACCESS_KEY_ID = env("DJANGO_AWS_ACCESS_KEY_ID")
 # https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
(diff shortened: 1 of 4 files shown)


BoPeng:
This looks really useful but I am not sure if `django-watchman` should be included as well. It might be useful to add both and this option as

```
Server health check:
1: none
2: health-check
3: watchman
```

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -73,6 +73,7 @@
 INSTALLED_APPS += ["storages"]  # noqa F405
 {%- endif -%}
 {% if cookiecutter.cloud_provider == 'AWS' %}
+INSTALLED_APPS += ["health_check.contrib.s3boto_storage"]
```
BoPeng: Should not it be `"health_check.contrib.s3boto3_storage"`?

danihodovic: Feel free to propose the change in this PR and I will update it

BoPeng: Sorry, this PR has conflict and cannot be patched from my fork. A "PRable" branch is https://github.com/BoPeng/cookiecutter-django/tree/health , which uses `health_check.contrib.s3boto3_storage` because this is what works for me.

{{cookiecutter.project_slug}}/requirements/base.txt
```diff
@@ -24,6 +24,7 @@ django-environ==0.4.5  # https://github.com/joke2k/django-environ
 django-model-utils==3.1.2  # https://github.com/jazzband/django-model-utils
 django-allauth==0.39.1  # https://github.com/pennersr/django-allauth
 django-crispy-forms==1.7.2  # https://github.com/django-crispy-forms/django-crispy-forms
+django-health-check==3.10.2
```
VadymKhodak, line 27:
```suggestion
django-health-check==3.10.2. # https://github.com/KristianOellegaard/django-health-check
```

browniebroke:
Thanks for the suggestion, but let's try to not add any more extra deps if they can be avoided - one could argue that we're already a bit past this point. Each dependency might slow down new Python/Django upgrades and we're already lagging a bit behind.

PR #2147, opened 2019-07-08
Always use test settings in pytest

danihodovic:

[//]: # (Before you proceed:)

## Description

You can alternatively use DJANGO_SETTINGS_MODULE, but this is overridden
by the environment variable of the same name. In order to always run it
with test settings, regardless of the environment variable, use --ds.

https://pytest-django.readthedocs.io/en/latest/configuring_django.html

## Rationale

### DIFF
--- a/{{cookiecutter.project_slug}}/pytest.ini
+++ b/{{cookiecutter.project_slug}}/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
-DJANGO_SETTINGS_MODULE=config.settings.test
+addopts = --ds=config.settings.test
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-DJANGO_SETTINGS_MODULE=config.settings.test
+addopts = --ds=config.test
```
browniebroke: I think it should be:

```suggestion
addopts = --ds=config.settings.test
```

danihodovic:
Good catch @browniebroke . It should be fixed now.

PR #2151, opened 2019-07-09
Remove gunicorn from installed apps

danihodovic:
I can't find the requirement of adding it to INSTALLED_APPS on the
Django or the Gunicorn docs.

https://docs.gunicorn.org/en/latest/run.html?highlight=django#django
https://docs.djangoproject.com/en/2.2/howto/deployment/wsgi/gunicorn/

[//]: # (Before you proceed:)

## Description

I can't find the requirement of adding it to INSTALLED_APPS on the
Django or the Gunicorn docs.

https://docs.gunicorn.org/en/latest/run.html?highlight=django#django
https://docs.djangoproject.com/en/2.2/howto/deployment/wsgi/gunicorn/

## Rationale

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -179,10 +179,6 @@ ANYMAIL = {
     "MAILGUN_API_URL": env("MAILGUN_API_URL", default="https://api.mailgun.net/v3"),
 }
 
-# Gunicorn
-# ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


browniebroke:
I think I found out why it was installed at some point, Gunicorn used to provide a `run_gunicorn` command which apparently required it [to be installed](https://stackoverflow.com/a/16859467/2261637) like this. 

It [was deprecated back in 2014](https://docs.gunicorn.org/en/latest/2014-news.html?highlight=run_gunicorn#deprecations), and this is indeed no longer needed.

PR #2161, opened 2019-07-14
Delete user_list.html

2O4:
Following on Remove user list view #2062

## Description

Since everything else is deleted
Probably don't need this template anymore.


browniebroke:
Good catch, thanks 🎉

PR #2167, opened 2019-07-17
Rename 403_csrf.html to 403.html

### LINKED ISSUE #2163: 403_csrf.html should be 403.html, no?
## What happened?
When testing a custom 403 page, I would expect the 403_csrf.html template to map to 403_csrf in debug mode.

 However instead this only works and maps to 403 if it is renamed to 403.html

## Steps to reproduce
Make a django-cookiecutter site. Change the default 403_csrf template.

step21:
403.html is the name that django expects, unless it is overriden. 403_csrf is a separate case, which according to django docs and mailing list should get a separate view/template. Fixes #2163 

- Docs: https://docs.djangoproject.com/en/2.1/ref/csrf/#rejected-requests
- Mailing list: https://groups.google.com/forum/#!msg/django-developers/O2raF0RVGA0/44-3Jz0jCwAJ
- Stackoverflow: https://stackoverflow.com/questions/31981239/django-custom-403-template


browniebroke:
Thanks for the fix 🎉

PR #2169, opened 2019-07-20
Always set REDIS_URL in env, regardless of celery use

danihodovic:

[//]: # (Before you proceed:)

## Description

I'm guessing that REDIS_URL should always be set, even if celery is not used in the project as they are not related.

## Rationale


demestav:
Why though? Please check https://github.com/pydanny/cookiecutter-django/pull/1956

browniebroke:
As per @demestav answer, this is intentional. There is no Redis service locally if Celery is not selected:

https://github.com/pydanny/cookiecutter-django/blob/60ce5ce2d561053b4c7f64c5971c1f9d78f03166/%7B%7Bcookiecutter.project_slug%7D%7D/local.yml#L45-L48

PR #2173, opened 2019-07-21
Add pre-commit for generated Django project

### LINKED ISSUE #2171: Add pre-commit for generated project
## Description

add pre-commit for generated django project.
add `.pre-commit-config.yaml` to project root directory, and also install pre-commit with pip.

## Rationale

This tool can help with pointing out issues in code, which also helpful when doing code review, do not have to wasting time with trivial style nitpicks.

As far as I know, there is only one case I've seen.
[bk-bcs-saas](https://github.com/Tencent/bk-bcs-saas/blob/master/.pre-commit-config.yaml)

leollon:

[//]: # (Before you proceed:)

## Description

add pre-commit for generated django project

## Rationale

`pre-commit` is a tool that can points out issues in code, it is helpful before doing code review, do not have to waste time with trivial style nitpicks.

[werkzeug](https://github.com/pallets/werkzeug)
[bk-bcs-saas](https://github.com/Tencent/bk-bcs-saas/blob/master/.pre-commit-config.yaml)

## Related issues

Fixes #2171

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -42,6 +42,6 @@
   "use_heroku": "n",
   "use_travisci": "n",
   "keep_local_envs_in_vcs": "y",
-
+  "use_pre_commit": "n",
   "debug": "n"
 }
--- a/docs/developing-locally.rst
+++ b/docs/developing-locally.rst
@@ -26,6 +26,8 @@ First things first.
 #. Install development requirements: ::
 
     $ pip install -r requirements/local.txt
+    # Optional: install git hook if using pre-commit
+    $ pre-commit install
 
 #. Create a new PostgreSQL database using createdb_: ::
 
--- a/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
+++ b/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
@@ -1,0 +1,32 @@
+exclude: .+/migrations/.+\.py
+default_stages: [commit]
+fail_fast: true
+
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: master
+    hooks:
+      - id: trailing-whitespace
+        files: (^|/){{cookiecutter.project_slug}}/.+\.(py|html|sh|css|js)$
+
+-   repo: local
+    hooks:
+      - id: black
+        name: black
+        entry: black
+        args: [--line-length=80]
+        language: python
+        types: [python]
+        files: (^|/){{cookiecutter.project_slug}}/.+\.py$
+        verbose: true
+
+-   repo: local
+    hooks:
+      - id: flake8
+        name: flake8
+        entry: flake8
+        args: [--max-line-length=120]
+        language: python
+        types: [python]
+        files: (^|/){{cookiecutter.project_slug}}/.+\.py$
+        verbose: true
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -136,6 +136,7 @@ Listed in alphabetical order.
   Keyvan Mosharraf         `@keyvanm`_
   Krzysztof Szumny         `@noisy`_
   Krzysztof Żuraw          `@krzysztofzuraw`_
+  Leo Wong                 `@leollon`_
   Leo Zhou                 `@glasslion`_
   Leonardo Jimenez         `@xpostudio4`_
   Lin Xianyi               `@iynaix`_
@@ -286,6 +287,7 @@ Listed in alphabetical order.
 .. _@kevgathuku: https://github.com/kevgathuku
 .. _@keyvanm: https://github.com/keyvanm
 .. _@knitatoms: https://github.com/knitatoms
+.. _@leollon: https://github.com/leollon
 .. _@krzysztofzuraw: https://github.com/krzysztofzuraw
 .. _@MathijsHoogland: https://github.com/MathijsHoogland
 .. _@mattayes: https://github.com/mattayes
--- a/README.rst
+++ b/README.rst
@@ -84,6 +84,7 @@ Optional Integrations
 .. _PythonAnywhere: https://www.pythonanywhere.com/
 .. _Traefik: https://traefik.io/
 .. _LetsEncrypt: https://letsencrypt.org/
+.. _pre-commit: https://github.com/pre-commit/pre-commit
 
 Constraints
 -----------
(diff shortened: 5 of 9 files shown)


{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+    hooks:
+      - id: trailing-whitespace
+        files: (^|/){{cookiecutter.project_slug}}/.+\.(py|html|sh|css|js)$
+
+-   repo: local
+    hooks:
+      - id: black
+        name: black
+        entry: black
+        args: [--line-length=80]
```
browniebroke: This should probably match [what we test for](https://github.com/pydanny/cookiecutter-django/blob/19eb42a7cf1bdf41fa3deef0af976ba2c7f2d1d3/tests/test_docker.sh#L23):

```suggestion
        args: [--exclude 'migrations']
```

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+        language: python
+        types: [python]
+        files: (^|/){{cookiecutter.project_slug}}/.+\.py$
```
browniebroke: Looking at the [Black documentation](https://black.readthedocs.io/en/stable/version_control_integration.html), I don't think we need this line:

```suggestion
```

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+        verbose: true
```
browniebroke: Looking at this:

> forces the output of the hook to be printed even when the hook passes

Seems a bit much, can we remove it?

```suggestion
```

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+-   repo: local
+    hooks:
+      - id: flake8
+        name: flake8
+        entry: flake8
+        args: [--max-line-length=120]
```
browniebroke: If we don't specify this, would the hook use the project config from the `setup.cfg`? That would be nice to avoid duplication...

https://github.com/pydanny/cookiecutter-django/blob/19eb42a7cf1bdf41fa3deef0af976ba2c7f2d1d3/%7B%7Bcookiecutter.project_slug%7D%7D/setup.cfg#L2

As a side note, I realised that the flake8 config should probably be updated to be [in line with Black's](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length) (out of scope of this pull request, but feel free to update it).

cookiecutter.json
```diff
@@ -42,6 +42,6 @@
   "use_heroku": "n",
   "use_travisci": "n",
   "keep_local_envs_in_vcs": "y",
-
+  "use_pre_commit": "n",
```
browniebroke: Looking at the diff now, I'm no longer sure we should make it an option, it's adding complexity to the template for not much benefit. Let's make it the default and mention it in our README, [under the list of features](https://github.com/pydanny/cookiecutter-django/blob/master/README.rst#features).

docs/developing-locally.rst
```diff
@@ -26,6 +26,8 @@ First things first.
 #. Install development requirements: ::
 
     $ pip install -r requirements/local.txt
+    # Optional: install git hook if using pre-commit
+    $ pre-commit install
```
browniebroke: Can we expand this section a bit? We could link to the pre-commit documentation for details.

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
leollon: Yes, I agree to remove it.

cookiecutter.json
leollon: You mean that `pre_commit` exists in final project as default and mention that, is that right?

browniebroke: Yes

docs/developing-locally.rst
leollon: I will add a note which is a link referred to the usage of `pre-commit` in its docs under the `pip`, like the process of creating database:

```#. Create a new PostgreSQL database using createdb_: ::

    $ createdb <what you have entered as the project_slug at setup stage> -U postgres --password <password>

   .. note::

       if this is the first time a database is created on your machine you might need an
       `initial PostgreSQL set up`_ to allow local connections & set a password for
       the ``postgres`` user. The `postgres documentation`_ explains the syntax of the config file
       that you need to change.
```
The note I will add as follows:

```
    .. note::

        the `pre-commit` will exist in the generated project as default.
        so, for the usage of `pre-commit`, take this [link](https://pre-commit.com/).
```

what do you think?

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
leollon: I am going to remove `args`, meanwhile update max-line-length to 88 for flake8 config in setup.cfg.

leollon: It can't be set to 88.

there is a one line whose length is greater than 88.
[config/settings/base.py](https://github.com/pydanny/cookiecutter-django/blob/19eb42a7cf1bdf41fa3deef0af976ba2c7f2d1d3/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/base.py#L122)

or, it has to break like this:
```
AUTH_PASSWORD_VALIDATORS = [
      {
          "NAME": "django.contrib.auth.password_validation.\
                  UserAttributeSimilarityValidator"
      },
      {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator"},
      {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"},
      {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator"},
  ]
```

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
leollon: so, I will remove `black` in the end.

leollon: Sorry that I respond so slowly. Since the two options in that test don't write the files back, and I add `pre-commit` for the generated project, not for this project(namely cookiecutter-django itself), so that when submission occurs, `pre-commit` can automatically format the code in file.  Later, I will make another PR. @browniebroke

PR #2202, opened 2019-08-21
Remove EMAIL_HOST, EMAIL_PORT for console backend

danihodovic:
django.core.mail.backends.console.EmailBackend is an in memory backend
so EMAIL_HOST and EMAIL_PORT does not need to be specified in this case.


browniebroke:
Thanks for this change.

PR #2205, opened 2019-08-25
Fix APPS path when using pytest

danihodovic:
Pytest does not use manage.py when executing tests. This means that the
trick we use to append app directories inside of the main project won't
work since the logic exists in manage.py.

If we instead place the same logic in base.py, things work as expected
when running pytest.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -1,6 +1,7 @@
 """
 Base settings to build other settings files upon.
 """
+import sys
 
 import environ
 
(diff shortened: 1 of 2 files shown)


danihodovic:
It should be noted that this breaks relative imports so the below won't work

```
from .models import Page
```

```
RuntimeError: Model class web.models.Page doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

browniebroke:
I'm not sure I understand the problem. Can you elaborate a bit on what breaks exactly? Aren't the test passing?

danihodovic:
You're absolutely correct. I was missing the project prefix when including the
app urls. It appeared as if the app wasn't in the python path.

```
path("", include("web.urls")),
```

fix

```
path("", include("my_project.web.urls")),
```


The stacktrace for future Google searches:
```
name = 'web', import_ = <function _gcd_import at 0x7f2eb3343e18>

>   ???
E   ModuleNotFoundError: No module named 'web'

<frozen importlib._bootstrap>:965: ModuleNotFoundError
```

PR #2206, opened 2019-08-26
Remove redundant DEBUG = False in test.py

danihodovic:
This is already set to False by default in base.py


browniebroke:
Also note that `pytest-django` sets it to `False` in a session fixture, so this is really not used:

https://github.com/pytest-dev/pytest-django/blob/d2973e21c34d843115acdbccdd7a16cb2714f4d3/pytest_django/plugin.py#L471

Issue #1689, opened 2018-06-20
React support through Webpack


anchpop:
## Description

It'd be really nice to have React as a built-in option, or have documentation on how to hack it in. I see this was suggested in #466, and seems to have been [merged](https://github.com/pydanny/cookiecutter-django/pull/610), but I saw no option for it when running cookiecutter (it also does not appear in [cookiecutter.json](https://github.com/pydanny/cookiecutter-django/blob/master/cookiecutter.json)). I've seen [ccokiecutter-webpack](https://github.com/goldhand/cookiecutter-webpack) but it hasn't been updated in 2 years which makes me nervous.

## Rationale

Lots of people prefer Webpack over Gulp. Also React is pretty good, but it's but a pain to use with Gulp. Even a working webpack setup, with no built-in react support, would be nice.

webyneter:
> have documentation on how to hack it in

This is a workable alternative, to my mind, write an external documentation somewhere and link it from our docs. Once again, we, the contributors and the community as a whole have yet to find a one-size-fits-all step-by-step third-party Cookiecutter Django-React integration guide -- why not to pioneer, @anchpop? :)

pydanny:
Just use `create-web-app` and build it in a new directory or even a separate repo. That's the best way, as it means you aren't dependent on us maintainers to configure Webpack for you, instead you get the best webpack people in the world doing it for you.

The same technique applies to VueJS. Just use Vue CLI instead of having a tight integration with this project.

pydanny:
Oops, this should be documented somewhere as @webyneter suggested.

browniebroke:
Good recent post: https://fractalideas.com/blog/making-react-and-django-play-well-together-single-page-app-model/

anchpop:
https://github.com/anchpop/annotator-tots/blob/master/tutorial/howToSwitchToWebpack.md I got something working, this is how I did it. I'm no react expert so this is very likely not the best way. Pretty painless, though

anchpop:
update: I now use [this](https://github.com/anchpop/cookiecutter-django), which I wrote and use in production. It is very similar to pydanny's cookiecutter-django.

pydanny:
This is awesome, @anchpop. Any chance you could do it for VueJS? 😜

anchpop:
@pydanny I will once I eventually learn Vue.js 😉

[...]

Issue #1153, opened 2017-05-01
Use webpack to compile scss/sass to css and add babel support


geekodour:
I am using cookiecutter-django with webpack and it's working nicely. I only have to keep running webpack on watch mode and everything else just works. Can I propose the change? It uses npm scripts instead of grunt.

webyneter:
@geekodour, for historical reference on `webpack` support, see #774

pydanny:
Inclusion of a light Webpack setup would be nice, but it would have to be easily maintainable for us.

geekodour:
I'll just post my folder structure webpack included here, not sure if it's good bad. it's probably bad.

```
.
├── contrib
├── __init__.py
├── media
├── node_modules <--
├── package.json <--
├── postcss.config.js <--
├── __pycache__
├── static
├── taskapp
├── templates
├── users
├── videos (app)
├── vlog (app)
└── webpack.config.js <--
```
inside static
```
.
├── fonts
├── images
│   └── favicon.ico
├── js
│   └── project.js
├── public
│   ├── Main.css
│   └── Main.js
├── sass
│   └── project.sass
└── vendor
    ├── responsive-nav.css
    └── responsive-nav.min.js
```
the `project.js` in `/js` @imports `project.sass` from `/sass` and I am using webpack extractTextPlugin to put the `Main.css` and `Main.js` into `/public`

bluesurfer:
@geekodour

Is webpack worth using in a multiple page application? 

I have never used webpack in practice but I read articles mainly about how cool it is :smile: and also how to use it to optimize the delivery of static content in a single page application. I understood the benefits there but what about a traditional Django app?

Braintelligence:
Well I have problems currently including Vue.js plugins without Webpack... It would be awesome if I could have just spun up Webpack for this.

natecox:
Just throwing this out there: I use https://github.com/owais/django-webpack-loader for managing webpack inside of Django, which is a really solid solution overall.

kiran-capoor94:
I have built a very simple integration with Webpack and it works in a very basic setup. It's my first time contributing, forgive me for the incomplete project. [https://github.com/kiran-capoor94/cookiecutter-django-webpack](url)

Issue #774, opened 2016-09-02
Consider removing Webpack/React option


pydanny:
We are getting a lot of support requests for our Webpack integration. Unfortunately, it doesn't work with all setups configurations all the time. None of the core devs of this project are that knowledgable of the toolchain, and have difficulty helping users. Community assistance with the issue has also been negligible, and we're not even sure who understands the toolchain.

Please share your opinions.

Paging @goldhand, @jayfk, @luzfcb, @audreyr, @theskumar

jayfk:
I'd love to keep support for webpack in Cookiecutter Django, but only if we find a way to provide a strong integration.

I think @goldhands work is fantastic, but a loosely coupled integration via post_gen hook is not enough in my opinion. Supporting Webpack and React/Redux pulls in a whole eco system of stuff that needs support. To make things worse, they are constantly moving (and breaking) things.

luzfcb:
@pydanny since I never use Webpack, and I only had time and reason to learn Webpack and ecosystem if the team that I work begin to use it (I do not see any chance of that happening in the short and medium term), I'm +1 to remove Webpack support from cookiecutter-django, except if someone takes responsibility to keep working integration with Webpack.

goldhand:
Awww this is sad but I understand. I've been trying to keep up with any webpack related issues but I've been a bit busy adjusting to my new employer and may have missed them. I'm still maintaining hzdg's cookiecutter-webpack but I'm no longer with them so thats another thing to consider in support of removing it. 

However, if there's a way to divert those support requests to cookiecutter-webpack instead of here maybe that would help?

Supporting all of cookiecutter-django's many deployment options would be difficult and too much of an undertaking for me alone. I could start working on a couple of the configurations, like heroku so that deployment works out of the box. I guess the next big issue would be getting docker working...

pydanny:
@goldhand, diverting support requests is only going to help so much because a lot of them are environment specific. Thanks for letting us know about your work status, as it is a factor. 

Here are the options I'm considering:

1. Maintain the status quo. Try to ramp up community support for the integration.
2. Add a fork or replacement of **hzdg/cookiecutter-webpack** to someone or some group with a vested interest in constantly maintaining it.
3. Remove direct Webpack/React integration and just specify how to integrate **hzdg/cookiecutter-webpack** in the docs.
4. Remove Webpack/React altogether

goldhand:
[...]

PR #2223, opened 2019-09-13
Add StimulusJS integration.

jeromecaisip:

[//]: # (Before you proceed:)

## Description

I am proposing of adding an option to use [stimulus JS](https://stimulusjs.org/) and setup the integration via webpack for those who usually use it and those who wants to get started using this library.

## Rationale

Because this could help devs who wants to start building multiple page applications
using stimulusJS, but don't have an idea of how to integrate the library in order to start. This could
also help awareness of this library and help devs with limited javascript experience to write better JS code.

I typically use this library, [stimulusJS](https://stimulusjs.org/), in order for me to write readable and maintainable javascript when building multiple page application.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -121,6 +121,7 @@ Listed in alphabetical order.
   Jan Van Bruggen          `@jvanbrug`_
   Jelmer Draaijer          `@foarsitter`_
   Jens Nilsson             `@phiberjenz`_
+  Jerome Caisip            `@jeromecaisip`_
   Jerome Leclanche         `@jleclanche`_                @Adys
   Jimmy Gitonga            `@afrowave`_                  @afrowave
   John Cass                `@jcass77`_                   @cass_john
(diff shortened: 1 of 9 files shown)


browniebroke:
Thanks for proposing this, looks interesting. I'm afraid I don't have any experience with this library, so I personally don't feel confortable maintaining this.

It would be nice to modernise our front-end setup, but I think this has been discussed in the past as a problem to keep things up to date (see #1689 #1153 #774)

Finally, the choice of framework is pretty subjective, others might have a preference for Angular, React, Vue (which @pydanny likes), Svelte, and it would be tricky for us to provide more than 1 option.

I would argue that it's outside the scope of this project, especially given that most of them have great CLI to get you started ([Angular CLI](https://cli.angular.io/), [create-react-app](https://github.com/facebook/create-react-app), [Vue CLI](https://cli.vuejs.org/)...).

It would be nice, however, to have a documented route for our users who would like to go down there.

jeromecaisip:
Hi. Thanks for the response. I agree that this particular is quite opinionated and the choice of frontend tools should be up to the user. 

Thanks.

PR #2225, opened 2019-09-18
Add option to use behave-django.

jeromecaisip:

[//]: # (Before you proceed:)

## Description

Add option to add django-behave for BDD.

## Rationale

Giving an option to use django-behave will make devs consider using this tool to do BDD with django ecosystem.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -33,6 +33,7 @@
     "GCP",
     "None"
   ],
+  "use_behave_django": "n",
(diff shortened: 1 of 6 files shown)


morenoh149:
looks interesting. there are conflicts

browniebroke:
Thanks for the suggestion, but let's try to not add any more extra deps if they can be avoided - one could argue that we're already a bit past this point. Each dependency might slow down new Python/Django upgrades and we're already lagging a bit behind.

PR #2234, opened 2019-09-23
Test users detail view

### LINKED ISSUE #1020: UserDetailView fails on mixed-case usernames
UserDetailView fails to find the user based on slug if the username is not stored in lowercase, which is likely to be the case given that django-allauth preserves case when storing usernames by default.

highpost:
## Description

Add a test for UsersDetailView. See https://github.com/pydanny/cookiecutter-django/issues/1020

## Rationale

https://github.com/pydanny/cookiecutter-django/issues/1020 says that a test is needed to track down a bug where case is sometimes not retained in usernames. I'm not sure if this bug is still present, because my test is currently passing. But I think this is a start.

Please let me know if this test needs anymore work.

Fixes #1020

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -6,7 +6,7 @@ from django.urls import reverse
 from {{ cookiecutter.project_slug }}.users.tests.factories import UserFactory
 from {{ cookiecutter.project_slug }}.users.views import (
     UserDetailView,
-    UserRedirectView, 
+    UserRedirectView,
     UserUpdateView,
 )
 
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -3,7 +3,7 @@ from django.conf import settings
 from django.test import RequestFactory
 from django.urls import reverse
 
-from .factories import UserFactory
+from {{ cookiecutter.project_slug }}.users.tests.factories import UserFactory
 from {{ cookiecutter.project_slug }}.users.views import (
     UserDetailView
     UserRedirectView, 
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -59,6 +59,7 @@ class TestUserRedirectView:
 
 
 class TestUserDetailView:
+    # PR #2234
     # Django usernames are case-sensitive.
     # See https://code.djangoproject.com/ticket/2273
     def test_username_case_detail_view(self, client, rf):
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -1,8 +1,10 @@
 import pytest
 from django.conf import settings
 from django.test import RequestFactory
+from django.urls import reverse
 
-from {{ cookiecutter.project_slug }}.users.views import UserRedirectView, UserUpdateView
+from {{ cookiecutter.project_slug }}.users.tests.factories import UserFactory
+from {{ cookiecutter.project_slug }}.users.views import UserDetailView, UserRedirectView, UserUpdateView
 
 pytestmark = pytest.mark.django_db
 
@@ -50,3 +52,30 @@ class TestUserRedirectView:
         view.request = request
 
         assert view.get_redirect_url() == f"/users/{user.username}/"
+
+
+class TestUserDetailView:
+    def test_user_detail_view(self, rf):
+        user          = UserFactory(username = 'tEsTcAsE')
+        request       = rf.get(
+                             path  = reverse(
+                                       viewname  = 'users:detail',
+                                       kwargs    = {'username': user.username}
+                                     )
+                           )
+        request.user  = user
+        response = UserDetailView.as_view()(request, **{'username': user.username})
+
+        assert response.status_code == 200
+
+    def test_user_detail_view_url(self, client):
+        user          = UserFactory(username = 'tEsTcAsE')
+        response      = client.get(
+                                 path  = reverse(
+                                           viewname  = 'users:detail',
+                                           kwargs    = {'username': user.username}
+                                         )
+                               )
+
+        assert response.status_code == 302
+        assert response.url == f'/accounts/login/?next=/users/{user.username}/'
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -62,7 +62,9 @@ class TestUserDetailView:
     def test_user_detail_view(self, rf):
         user = UserFactory(username="tEsTcAsE")
         request = rf.get(
-            path=reverse(viewname="users:detail", kwargs={"username": user.username.lower()})
+            path=reverse(
+                viewname="users:detail", kwargs={"username": user.username.lower()}
+            )
         )
         request.user = user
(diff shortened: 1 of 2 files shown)


highpost:
I modified the import statements to use relative imports because I was getting an error from black:

> error: cannot format test_views.py: Cannot parse: 6:5: from {{ cookiecutter.project_slug }}.users.tests.factories import UserFactory

This is probably a Jinja2 problem, but I couldn't figure it out quickly.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -1,8 +1,10 @@
 import pytest
 from django.conf import settings
 from django.test import RequestFactory
+from django.urls import reverse
 
-from {{ cookiecutter.project_slug }}.users.views import UserRedirectView, UserUpdateView
+from .factories import UserFactory
+from ..views import UserDetailView, UserRedirectView, UserUpdateView
```
browniebroke:
```suggestion
from {{ cookiecutter.project_slug }}.users.views import (
    UserDetailView
    UserRedirectView, 
    UserUpdateView,
)
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -1,8 +1,10 @@
 import pytest
 from django.conf import settings
 from django.test import RequestFactory
+from django.urls import reverse
 
-from {{ cookiecutter.project_slug }}.users.views import UserRedirectView, UserUpdateView
+from .factories import UserFactory
```
browniebroke:
```suggestion
from {{ cookiecutter.project_slug }}.users.tests.factories import UserFactory
```

highpost: Thanks for your review. Does this test resolve https://github.com/pydanny/cookiecutter-django/issues/1020?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -1,8 +1,14 @@
 import pytest
 from django.conf import settings
 from django.test import RequestFactory
+from django.urls import reverse
 
-from {{ cookiecutter.project_slug }}.users.views import UserRedirectView, UserUpdateView
+from {{ cookiecutter.project_slug }}.users.tests.factories import UserFactory
+from {{ cookiecutter.project_slug }}.users.views import (
+    UserDetailView
```
browniebroke: My bad, forgot a trailing comma in my previous suggestion:

```suggestion
    UserDetailView,
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
browniebroke: I think it is, yes. This a fairly old issue so it might have been resolved already.

browniebroke: Actually, it might be happening when the URL is accessed with lowercase?

```
reverse(viewname="users:detail", kwargs={"username": user.username.lower()})
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
highpost: I appended lower() to the path() strings and I think the test now demonstrates that there's a bug. Note that I have two versions of the test, one that calls the view directly and one that uses the test client fixture to traverse through, well, everything.

I added the comma ;).

highpost: I meant to say that the first test passes and the second test fails. So that suggests that the bug isn't in the view.

highpost:
Thanks for your review. Not sure where my other comments went, so I'll reiterate them here.

Thanks. I applied your changes and fixed a few other cosmetic issues. I think that the test now demonstrates that there is a bug. Note that I have two versions of the test, one that calls the view directly and one that uses the test client to exercise the full url request-response mechanism. Since the first version passes and the second version fails I *think* this demonstrates the presence of a bug.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -50,3 +56,28 @@ def test_get_redirect_url(
...
+    def test_user_detail_view_url(self, client):
+        user = UserFactory(username="tEsTcAsE")
+        response = client.get(
+            path=reverse(
+                viewname="users:detail", kwargs={"username": user.username.lower()}
+            )
+        )
+
+        assert response.status_code == 302
+        assert response.url == f"/accounts/login/?next=/users/{user.username}/"
```
browniebroke: I think this change should make the test pass:

```suggestion
        assert response.url == f"/accounts/login/?next=/users/{user.username.lower()}/"
```

I'm not sure there is a bug after all...

highpost: Sorry for the delay. If the canonical URL for the user is /users/tEsTcAsE, then shouldn't the response URL also be '/users/tEsTcAsE'?

browniebroke: >  I don't know how to check the equivalent response URL for the first test case. I just get a 200.

Actually, I realise that this test is probably not testing what we should. It's redirecting because the `UserDetailView` inherits from the `LoginRequiredMixin`, so this test is not even entering the code we should test IMO.

https://github.com/pydanny/cookiecutter-django/blob/2459392a986a193450144aa3fe14228910183f47/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/users/views.py#L11

To correct this and actually test the view, we should log-in the client before calling `client.get()`. The easier way to do that is to use [the `force_login()` method](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.Client.force_login):

```
client.force_login(user)
```

Then the response status code should either be 200 or 404. 

If it's 200, the next thing to check would be what happens when we have 2 users with the same username but with different cases, e.g. `tEsTcAsE` and `testcase`.

highpost: Thanks for your guidance on this PR. I created a second user with lowercase ('testcase') and added force_login() but I still got a 200. I also spent some time with pdb tracing through the Django view code.

I looked at the logic of using lower() in the result and I don't think that's the right way to go because we want to test *for* case sensitivity. Finally, I read through https://code.djangoproject.com/ticket/2273 which makes it clear that Django usernames are case sensitive and will remain so.

In the end I no longer think that this is a bug. If you want to keep the test I could modify it to be a vanilla view test becase I think it's still useful as an example. Either way...

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -50,3 +56,30 @@ def test_get_redirect_url(
...
+        )
+        request.user = user
+        response = UserDetailView.as_view()(request, **{"username": user.username})
+
+        assert response.status_code == 200
+
+    def test_username_case_detail_view_url(self, client):
+        _ = UserFactory(username="testcase")
+        user = UserFactory(username="tEsTcAsE")
+        response = client.get(
```
browniebroke:
```suggestion
		client.force_login(user)
        response = client.get(
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
browniebroke: Thanks for digging all this. I agree, doesn't seem like a bug. I think this could still be good to have some tests for this view. Could you please just make sure the client is logged in for **both tests** (see my latest suggestion). The `response.status_code` should not be 302 IMO...

browniebroke:
I agree that it doesn't seem like a bug. I'll close this and the associated issue, thanks a lot for your help!

PR #2251, opened 2019-10-02
Fix #2250 -- No staticfile storage with GCP without Whitenoise

### LINKED ISSUE #2250: No staticfile storage with GCP without Whitenoise
## What happened?

If the user chooses GCP as cloud provider & no Whitenoise, then `STATICFILES_STORAGE` is unset in production.

https://github.com/pydanny/cookiecutter-django/blob/164e2aa69c2bf58104de7ad3474d89995329b8b5/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L104-L111

The `STATICFILES_STORAGE` setting should be set to `"storages.backends.gcloud.GoogleCloudStorage"` or a subclass.

## Steps to reproduce

Reproducible using:

`cookiecutter https://github.com/pydanny/cookiecutter-django use_whitenoise=n cloud_provider=GCP`

caioariede:

[//]: # (Before you proceed:)

## Description

Fixes #2250

## Rationale

If the user chooses GCP as cloud provider & no Whitenoise, then STATICFILES_STORAGE should be set in production

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -92,7 +92,6 @@ AWS_DEFAULT_ACL = None
 # https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
 AWS_S3_REGION_NAME = env("DJANGO_AWS_S3_REGION_NAME", default=None)
 {% elif cookiecutter.cloud_provider == 'GCP' %}
-DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
 GS_BUCKET_NAME = env("DJANGO_GCP_STORAGE_BUCKET_NAME")
 GS_DEFAULT_ACL = "publicRead"
 {% endif -%}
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -107,6 +107,7 @@
 STATICFILES_STORAGE = "config.settings.production.StaticRootS3Boto3Storage"
 STATIC_URL = f"https://{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com/static/"
 {% elif cookiecutter.cloud_provider == 'GCP' -%}
+STATICFILES_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
```
browniebroke: I _beleive_ this will cause Static files to be stored in the same location as Media files (same as `DEFAULT_FILE_STORAGE` + `GS_BUCKET_NAME`).

Ideally, we should create 2 subclasses like we do for AWS below:

https://github.com/pydanny/cookiecutter-django/blob/f3b46ec4251358b7dd5ae24236b7cd9cc0a0b441/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L119-L129

caioariede: oh, cool! I'll update it

browniebroke:
Thanks! 🎉

PR #2252, opened 2019-10-02
Fix #2249 -- Broken static files with whitenoise=n & cloud_prov…

### LINKED ISSUE #2249: Broken static files with whitenoise=n & cloud_provider=none
## What happened?

If the user chooses `use_whitenoise=n` & `cloud_provider=None`, then the static files are not served properly.

This combination should not be allowed and show an error when trying to generate the project.

caioariede:

[//]: # (Before you proceed:)

## Description

Fixes https://github.com/pydanny/cookiecutter-django/issues/2249

## Rationale

If the user chooses use_whitenoise=n & cloud_provider=None, it should not be allowed and show an error when trying to generate the project.

### DIFF
--- a/tests/test_cookiecutter_generation.py
+++ b/tests/test_cookiecutter_generation.py
@@ -165,6 +165,7 @@ def test_invalid_slug(cookies, context, slug):
 
 
 def test_no_whitenoise_and_no_cloud_provider(cookies, context):
+    """It should not generate project if neither whitenoise or cloud provider are set"""
     context.update({"use_whitenoise": "n", "cloud_provider": "None"})
     result = cookies.bake(extra_context=context)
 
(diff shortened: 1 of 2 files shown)


tests/test_cookiecutter_generation.py
```diff
@@ -157,3 +162,11 @@ def test_invalid_slug(cookies, context, slug):
...
     assert result.exit_code != 0
     assert isinstance(result.exception, FailedHookException)
+
+
+def test_no_whitenoise_and_no_cloud_provider(cookies, context):
+    """It should not generate project if neither whitenoise or cloud provider are set"""
+    context.update({"use_whitenoise": "n", "cloud_provider": "None"})
+    result = cookies.bake(extra_context=context)
+
+    assert result.exit_code == 1
```
browniebroke: Looks like it's `-1` for some reason... I suppose it's good enough to check that it's non-zero, and that the exception type is what we want. We don't really need a specific number:

```suggestion
    assert result.exit_code != 0
    assert isinstance(result.exception, FailedHookException)
```

caioariede:
Thanks for the suggestion @browniebroke

browniebroke:
Thanks 🎉

PR #2256, opened 2019-10-04
Adding conda to get Cookiecutter

RichaldoElias:

[//]: # (Before you proceed:)

## Description

## Rationale

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -132,8 +132,15 @@ Let's pretend you want to create a Django project called "redditclone". Rather t
 and then editing the results to include your name, email, and various configuration issues that always get forgotten until the worst possible moment, get cookiecutter_ to do all the work.
 
 First, get Cookiecutter. Trust me, it's awesome::
+There are to ways get Cookiecutter:
+
+Using pip
 
     $ pip install "cookiecutter>=1.4.0"
+    
+Using conda
+
+    $ conda install -c cmckeague cookiecutter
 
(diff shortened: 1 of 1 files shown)


RichaldoElias:
pip is not the only way to get Cookiecutter, we can also use  conda

README.rst
```diff
@@ -132,8 +132,15 @@ Let's pretend you want to create a Django project called "redditclone". Rather t
 and then editing the results to include your name, email, and various configuration issues that always get forgotten until the worst possible moment, get cookiecutter_ to do all the work.
 
 First, get Cookiecutter. Trust me, it's awesome::
+There are to ways get Cookiecutter:
```
rosariopfernandes: Did you mean two?

browniebroke:
Correct, it's not the only way, there is also (at least):

- Homebrew
- pipx
- pipenv

Cookiecutter Django is a tempalate, I don't think we should list all of the ways to install the cookiecutter CLI, but rather document the preferred way and then link to the CLI documentation for other ways. Something along the lines of this maybe?

> First, get Cookiecutter via pip (or [via another installation method](https://cookiecutter.readthedocs.io/en/latest/installation.html#install-cookiecutter))

browniebroke:
Thanks for the effort in trying to get this better documented, but it's not in a state to be accepted here and this is quite well documented in the CLI documentation.

PR #2279, opened 2019-10-22
Prevent collectfast from guessing a wrong strategy

mapx:

[//]: # (Before you proceed:)

## Description
This PR fixes the following issue raised when one starts Django (with AWS) docker service:
Screenshot from 2019-10-22 08-16-36

## Rationale
If `COLLECTFAST_STRATEGY` is not provided in `settings`, `collectfast` has to guess a "strategy". For the case with AWS, it guesses `BotoStrategy` in stead of `S3Boto3Storage`, which is a wrong decision as we `pip install django-storages[boto3]` (`boto3` in favor of `boto`). 

Setting `COLLECTFAST_STRATEGY` is also encouraged as `collectfast` is not going to guess the strategy in the future.

Ref:
* https://github.com/antonagestam/collectfast/blob/f37dd1faa2046a2ebb88c098eb03e56a95534ec6/collectfast/management/commands/collectstatic.py#L33
* https://github.com/antonagestam/collectfast/blob/f37dd1faa2046a2ebb88c098eb03e56a95534ec6/collectfast/strategies/base.py#L170

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -200,6 +200,7 @@ Listed in alphabetical order.
   William Archinal         `@archinal`_
   Xaver Y.R. Chen          `@yrchen`_                   @yrchen
   Yaroslav Halchenko
+  Yuchen Xie               `@mapx`_
 ========================== ============================ ==============
 
 .. _@a7p: https://github.com/a7p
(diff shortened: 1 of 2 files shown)


luzfcb:
@mapx Please, add yourself to the CONTRIBUTORS.rst

mapx:
> @mapx Please, add yourself to the CONTRIBUTORS.rst

Thanks :) The commit was amended accordingly.

PR #2312, opened 2019-11-13
Use pytest-django --reuse-db for faster tests

danihodovic:
Pytest-django allows you to re-use database setups between tests for
quicker test runs. This is especially useful for large projects with many
migrations.

For example, here's a project of mine with 40+ migrations:

with --reuse-db
```
pytest --reuse-db  2.54s user 0.35s system 108% cpu 2.669 total
```

without --reuse-db
```
pytest 7.40s user 0.34s system 50% cpu 15.240 total
```

Caveat: if your model happens to change you need to manually --create-db
to ensure migrations are applied.

https://pytest-django.readthedocs.io/en/latest/database.html#reuse-db-reuse-the-testing-database-between-test-runs

[//]: # (Before you proceed:)

## Description

## Rationale

### DIFF
--- a/{{cookiecutter.project_slug}}/pytest.ini
+++ b/{{cookiecutter.project_slug}}/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
-addopts = --ds=config.settings.test
+addopts = --ds=config.settings.test --reuse-db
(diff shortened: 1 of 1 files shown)


browniebroke:
Good idea!

PR #2321, opened 2019-11-22
Add an option to choose AWS SES for email

joshtensibly:

[//]: # (Before you proceed:)

## Description

Add the option to choose SES as email provider. This will install `django-anymail[amazon_ses]` and set the `EMAIL_BACKEND` to `anymail.backends.amazon_ses.EmailBackend`

## Rationale

To allow users to start a project with SES for email instead of Mailgun of they wish.

Send emails with SES.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -160,6 +160,7 @@ Listed in alphabetical order.
   Meghan Heintz            `@dot2dotseurat`_
   Mesut Yılmaz             `@myilmaz`_
   Michael Gecht            `@mimischi`_                  @_mischi
+  Michael Karamuth         `@joshtensibly`_               
   Min ho Kim               `@minho42`_
   mozillazg                `@mozillazg`_
   Nico Stefani             `@nicolas471`_                @moby_dick91
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -17,22 +17,10 @@
   "windows": "n",
   "use_pycharm": "n",
   "use_docker": "n",
-  "postgresql_version": [
-    "11.3",
-    "10.8",
-    "9.6",
-    "9.5",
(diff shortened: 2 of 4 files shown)


foarsitter:
When you select a different value for `cloud_provider` then `AWS`, the `AWS_*` variables are not added to the settings. This breaks you configuration for SES since the credentials are not provided.  

https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.envs/.production/.django#L19

cookiecutter.json
```diff
@@ -17,22 +17,10 @@
...
-    "Gulp"
-  ],
-  "cloud_provider": [
-    "AWS",
-    "GCP",
-    "None"
-  ],
+  "postgresql_version": ["11.3", "10.8", "9.6", "9.5", "9.4"],
+  "js_task_runner": ["None", "Gulp"],
+  "cloud_provider": ["AWS", "GCP", "None"],
```
browniebroke: Could you please not change the lines wrapping and stick to the code style of the file? We tend to have 1 option per line as it's making diff easier to process.

CONTRIBUTORS.rst
```diff
@@ -160,6 +160,7 @@ Listed in alphabetical order.
   Meghan Heintz            `@dot2dotseurat`_
   Mesut Yılmaz             `@myilmaz`_
   Michael Gecht            `@mimischi`_                  @_mischi
+  Michael Karamuth         `@joshtensibly`_
```
browniebroke: This is creating a broken link, we need to add the target for `@joshtensibly` firther down in the file.

browniebroke:
Thanks for your effort! I forgot about this, but it's now superseded by #2435, closing

PR #2338, opened 2019-11-29
Upgrade to Traefik v2

blaxpy:
## Description
Upgrade to Traefik v2

Fixes #2378


foarsitter:
Merged this update by hand in an existing project and it looks good to me. 

The only thing I ran into was that the certificates generated by v1 are not compatible with v2. Removing the docker traefik volume solves this issue.

PR #2353, opened 2019-12-09
Speed up django image build process, and reduce its size

mapx:
## Description

Improve Django image build process, and reduce its size.

## Rationale
As @iMerica mentioned in https://github.com/pydanny/cookiecutter-django/issues/1157, this modification reduces hundreds of megabytes from the Docker image.
We also found the new build process a lot faster. The command `RUN chown -R django /app` takes too much time.


foarsitter:
Indeed, this gives the build process a major speed boost! Thx for this pull request!

browniebroke:
Thanks for this, gave it a go and it seems to work.

PR #2356, opened 2019-12-10
Allow rendering custom widgets from root templates dir

danihodovic:
Rendering custom widgets from the projects root template directory
doesn't work out of the box.

See: https://stackoverflow.com/questions/45844032/django-templatedoesnotexist-in-case-of-a-custom-widget/46208414#46-208414

This changes the setting to use the same template finding logic that the
TEMPLATES setting uses.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -175,6 +175,8 @@ TEMPLATES = [
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         # https://docs.djangoproject.com/en/dev/ref/settings/#template-dirs
         "DIRS": [str(APPS_DIR.path("templates"))],
+        # https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-TEMPLATES-APP_DIRS
+        "APP_DIRS": True,
         "OPTIONS": {
             # https://docs.djangoproject.com/en/dev/ref/settings/#template-loaders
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -195,6 +196,10 @@
         },
     }
 ]
+
+# https://docs.djangoproject.com/en/3.0/ref/settings/#form-renderer
```
browniebroke: We usually avoid mentioning a version in here:

```suggestion
# https://docs.djangoproject.com/en/dev/ref/settings/#form-renderer
```

browniebroke:
Ok, my mistake. The build just failed with:

```
django.core.exceptions.ImproperlyConfigured: app_dirs must not be set when loaders is defined.
```

Looking at the loaders doc, I now realise that what I suggested earlier is redundant with the [`django.template.loaders.app_directories.Loader`](https://docs.djangoproject.com/en/dev/ref/templates/api/#django.template.loaders.app_directories.Loader) that we already use:

> You can enable this loader by setting APP_DIRS to True

https://github.com/pydanny/cookiecutter-django/blob/f5aa06c160bccf037f9c87af21264965ce7a8ec2/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/base.py#L180-L183

Apologies for the wrong indicatons, if you could please revert the `APP_DIRS` change I think this is good to go.

browniebroke:
Thanks 🎉

PR #2378, opened 2019-12-22
Add Debian 10 (buster) OS dependencies

ddiazpinto:
## Description

They are all checked. Same packages names as bionic, except for `libjpeg`.


browniebroke:
Thanks!

Issue #2349, opened 2019-12-05
entrypoint script uses Bash features despite POSIX shell shebang


moseb:
## What happened?

At …
https://github.com/pydanny/cookiecutter-django/blob/efbf164e84998a8d450133bd00c14c07b25bf8c4/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/django/entrypoint#L1
… it reads `#!/bin/sh` so the script has no guarantee to deal with Bash (and not just a POSIX shell) but then uses Bash features, e.g. `set -o pipefail`.

## What should've happened instead?

I would expect
- a) use of `#! /usr/bin/env bash` or
- b) a true POSIX shell script without use of Bashisms

Thank you!

browniebroke:
As far as I remember, we intended to use `sh` because `bash` isn't installed in the alpine image we use. Feel free to send a pull request to cleanup the "bashism".

aguerreroec:
Hello,
I've recently changed my django/dockerfile from alpine to python:3.7-slim-buster.
When i try to run docker-compose -f local.yml up , appears the following error:

`django_1    | /entrypoint: 4: set: Illegal option -o pipefail`

do you know why occur this?

moseb:
> do you know why occur this?

Because `/bin/sh` in Debian buster understands `set -o errexit` and applies it and then `set -o pipefail`  not only fails loudly but also exits the shell.

PR #2391, opened 2020-01-07
Replaced Docker python:3.7-alpine image with python:3.7-slim-bu…

### LINKED ISSUE #2373: Proposal to use python-slim instead of python-alpine
## Description

Use the python-slim Docker image as a base instead of the python-alpine variant.

## Rationale

I believe you guys switched to the Alpine image to reduce space, but the python-slim variant is about the same size. Alpine does not support manylinux wheels from Pypi, so many packages have to be built from source, including cython, numpy, pandas (which takes FOREVER), and others. This is further complicated by the fact that Alpine doesn't contain undocumented build dependencies of some pip packages such as gcc , g++, etc. 

Further justification can be found here https://pythonspeed.com/articles/base-image-python-docker-images/

trungdong:
## Description

This PR is in response to proposal #2373 - moving from Alpine to Debian base images.

## Rationale

Please refer to #2373 for the original discussion.
(Also see https://pythonspeed.com/articles/base-image-python-docker-images/)

~~In addition, I chose not to use the `python:3.7-slim-buster` image because it contains 2 versions of Python (the lastest Python built from source and the system `python3` package, which gets installed from as a dependency of `python3-dev` package). As a result, it does not allow the use of Debian packages for numpy, scipy, etc., which would require installing various extra packages to build from source.~~
Update: Now using `python:3.7-slim-buster` to have the later version of Python.

## Notes

* (Before) Alpine-based Django image size: 609 MB
* (After) Debian-based Django image size: ~~762 MB~~ 738 MB

Fixes #2373 #2349


{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
```diff
@@ -1,19 +1,21 @@
-FROM python:3.7-alpine
+FROM debian:buster-slim
```
browniebroke: Using this image instead of `python:3.7-slim-buster` would make it much harder to upgrade to Python 3.8, isn't there a way to get the python image to work? 

The Python it's using (3.7.3) is 9 months, not very convenient if one wants to upgrade to the latest.

trungdong: I'm sure there is a way to make the `python:3.7-slim-buster` work. It's a matter of efficiency and choice.

That image built Python 3.7.6 from source and put it at `/usr/local/bin/python`. This Python version looks for Python packages in `/usr/local/lib/python3.7/site-packages`.

In our Dockerfile, we install the package `python3-dev` which also installs its dependency, `python3` package. This version of Python(3.7.3) lives at `/usr/bin/python3` and sees Python packages installed in `/usr/local/lib/python3.7/dist-packages` and `/usr/lib/python3/dist-packages`.

Running just `python` in the `python:3.7-slim-buster` image (expectedly) uses its own Python 3.7.6 version, which does not see any system-distributed Python packages you may want to install. In my case, these are `python3-numpy`, `python3-scipy` (which would be installed into `/usr/lib/python3/dist-packages`). If you don't need those, the `python:3.7-slim-buster` image works fine with the default requirements files, albeit the fact that there are two versions of Python on the image.

If you require, say, numpy in your project, you can still install its wheel files without building but some C libraries it uses (i.e. libblas3, libgfortran5, liblapack3) are not installed. It may still work, but I'm not sure (or whether it works as fast).

So I think this is down to personal preference. I am happy with the Debian-packaged version of Python and like to be able to use Debian-packaged Python libraries without having to worry about their dependencies like above. An extra bonus is that I save some space taken by the second version of Python on the image.

trungdong: Just another note to say that I've tried the `python:3.7-slim-buster` and it works (as in all tests passed). The image size is 857 MB though.

If one really wants to use the `python:3.7-slim-buster` image and also want a small image, I guess we could purge the whole system Python version and its related packages once everything is built and installed.

trungdong: Further to my previous message, I've managed to get the `python:3.7-slim-buster` without having to install the Debian-packaged Python version (see commit 4a71b0b below). The resulted image size is now slightly smaller, 738 MB.

browniebroke: Thanks for all these detailed explanations, good to clarify what was going on. And good work on keeping the image size down. I want to take a closer look, I hope to get some time this month to review all the outstading pull requests.

demestav:
> In addition, I chose not to use the python:3.7-slim-buster image

But is seems that you are using `python:3.7-slim-buster`, right?

trungdong:
> But is seems that you are using `python:3.7-slim-buster`, right?

Yes, I've changed to `python:3.7-slim-buster` in commit 4a71b0b above (after a discussion with @browniebroke in the outdated reviewed code block).

demestav:
Sorry, I completely missed that discussion. I read it now and makes sense. Thanks for extensive explanation.

PR #2393, opened 2020-01-09
Fix default pycharm django console config

NoNameItem:

[//]: # (Before you proceed:)

## Description

Remove "os.environ.setdefault(&quot;DATABASE_URL&quot;,&quot;postgres://{}:{}@{}:{}/{}&quot;.format(os.environ['POSTGRES_USER'], os.environ['POSTGRES_PASSWORD'], os.environ['POSTGRES_HOST'], os.environ['POSTGRES_PORT'], os.environ['POSTGRES_DB']))&#10;os.environ.setdefault(&quot;CELERY_BROKER_URL&quot;, os.environ['REDIS_URL'])" from default Django console start script in PyCharm

## Rationale

Currently django console launch ends with KeyError because of "os.environ['POSTGRES_USER'], os.environ['POSTGRES_PASSWORD']...". This env variables is never mentioned in cookiecutter-django docs and quite confusing. Actually it's unnecessary to set this env variable in this script

### DIFF
--- a/{{cookiecutter.project_slug}}/.idea/workspace.xml
+++ b/{{cookiecutter.project_slug}}/.idea/workspace.xml
@@ -2,12 +2,12 @@
 <project version="4">
     {%- if cookiecutter.use_celery == 'y' %}
         <component name="DjangoConsoleOptions"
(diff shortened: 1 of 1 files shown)


foarsitter:
When using docker these env variables do exists. De proper solution here is to[ wrap it all](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.idea/workspace.xml) in an if-statement `{%- if cookiecutter.use_docker == 'y' %}` or remove the file completely when `use_docker == 'n'`.

browniebroke:
Thanks for the report. Fixed in 8fd0d42de0e7c3c8fea92bb4d1d2df3e4af6aa31

PR #2406, opened 2020-01-19
Merge pull request #2406 from BoPeng/master

### LINKED ISSUE #2405: UserFactory ignores passed password
## What happened?

When using fixture,

```
from myapp.users.tests.factories import UserFactory
UserFactory(username='bob', password='password')
client.login(username='bob', password='password')
client.get('/url_needs_login')
```

would fail.

Login should be successful.

## Reason

https://github.com/pydanny/cookiecutter-django/blob/68f7268770057752609dd1f6beca4d432eb9f2a5/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/users/tests/factories.py#L13-L23
 
Does not following the [patten of `post_generation`](https://factoryboy.readthedocs.io/en/latest/reference.html) and ignores `extracted`, which is user specified password. Test would pass if I change the code to

```
password = extracted if extracted else Fake(
```

Question: Is always generating a random password (and ignores passed password) by design, or a bug?

BoPeng:

[//]: # (Before you proceed:)

## Description

`UserFactory` ignores user-provided password so `client` cannot use it to log in. The following does not work

```
UserFactory(username='bob', password='password')
client.login(username='bob', password='password')
```
because `UserFactory` always set a randomly generated password.

## Rationale

Whereas it is understandable that `user` fixture always have random password, direct use of `UserFactory(username, password)` should not silently ignore `password`, which causes confusion when users use `client.login()` to login with the same password.

#2405 has some a use case.

Disclaimer: I am not an expert on `factory_boy`. It seems that `create` should be used in this function to determine if the factory is called as `UserFactory` or `UserFactory.create`.

Fixes #2405

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/factories.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/factories.py
@@ -12,14 +12,18 @@ class UserFactory(DjangoModelFactory):
 
     @post_generation
     def password(self, create: bool, extracted: Sequence[Any], **kwargs):
-        password = extracted if extracted else Faker(
-            "password",
-            length=42,
-            special_chars=True,
-            digits=True,
-            upper_case=True,
-            lower_case=True,
-        ).generate(extra_kwargs={})
+        password = (
+            extracted
+            if extracted
+            else Faker(
+                "password",
+                length=42,
+                special_chars=True,
+                digits=True,
+                upper_case=True,
+                lower_case=True,
+            ).generate(extra_kwargs={})
+        )
         self.set_password(password)
 
     class Meta:
--- a/{{cookiecutter.project_slug}}/config/settings/test.py
+++ b/{{cookiecutter.project_slug}}/config/settings/test.py
@@ -32,7 +32,7 @@ PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
 
 # TEMPLATES
 # ------------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/factories.py
```diff
@@ -12,7 +12,7 @@ class UserFactory(DjangoModelFactory):
 
     @post_generation
     def password(self, create: bool, extracted: Sequence[Any], **kwargs):
-        password = Faker(
+        password = extracted if extracted else Faker(
```
browniebroke: The build is failing because this breaks black formatting. It's not clear from the output, but the right formatting is:

```python
        password = (
            extracted
            if extracted
            else Faker(
                "password",
                length=42,
                special_chars=True,
                digits=True,
                upper_case=True,
                lower_case=True,
            ).generate(extra_kwargs={})
        )
```

Would you mind updating this?

BoPeng: @browniebroke I think you just did it, right?

browniebroke: Yes, realised you gave me access to fix the branch. Thanks for that 🎉

BoPeng: Many thanks to you guys for creating and maintaining this wonderful project, which greatly reduced my effort to set up my new project.

PR #2412, opened 2020-01-24
Fixed the redis-py project URL

msamoylov:
## Description

Previously it was pointing to the Redis project repository.

## Rationale

It was misleading and confusing.


browniebroke:
Thanks for the fix! 🎉

PR #2417, opened 2020-01-29
Add django and postgres envs to celeryworker and celerybeat

BoPeng:

[//]: # (Before you proceed:)

## Description

It took me a while to get my celery worker and celery beat docker images working on the production environment. The error messages are similar to

```
Message Error: cannot add item to database

celerybeat raised exception <class 'dbm.error'>: error('cannot add item to database',)
```

and it turned out that the celery worker could not connect to `S3` and database because of missing `AWS` and postgres environment variables, and `celerybeat` could not connect to the database because of missing postgres environment variables.

# What's it you're proposing?

Adding 

```
+    env_file:
+      - ./.envs/.production/.django
+      - ./.envs/.production/.postgres
```
to `celery` etc in `production.yml`.

Note that it is possible that users' celery worker does not talk to S3 or the database, but I do not see any harm making the variables available to the worker.

### DIFF
--- a/{{cookiecutter.project_slug}}/production.yml
+++ b/{{cookiecutter.project_slug}}/production.yml
@@ -51,11 +51,16 @@ services:
     <<: *django
     image: {{ cookiecutter.project_slug }}_production_celeryworker
     command: /start-celeryworker
+    env_file:
+      - ./.envs/.production/.django
+      - ./.envs/.production/.postgres
 
   celerybeat:
     <<: *django
     image: {{ cookiecutter.project_slug }}_production_celerybeat
     command: /start-celerybeat
+    env_file:
+      - ./.envs/.production/.postgres
 
   flower:
     <<: *django


{{cookiecutter.project_slug}}/production.yml
```diff
@@ -51,11 +51,16 @@ services:
     <<: *django
     image: {{ cookiecutter.project_slug }}_production_celeryworker
     command: /start-celeryworker
+    env_file:
+      - ./.envs/.production/.django
+      - ./.envs/.production/.postgres
```
browniebroke: They should already be included in [the `django` anchor](https://yaml.org/spec/1.2/spec.html#id2785586) we define at the top:

https://github.com/pydanny/cookiecutter-django/blob/2698867cf8b949b03ed215883f51538fe2f04b38/%7B%7Bcookiecutter.project_slug%7D%7D/production.yml#L9-L19

Note the `django: &django` at the top, which expanded at the line 51 and 59 with `<<: *django`

BoPeng:
The same here. You are perhaps right, but my project was 'vanilla' and it did need these `env_files` to work ....

PR #2433, opened 2020-02-04
Flower now served by Traefik

### LINKED ISSUE #2394: Celery Flower is not configured for Traefik
## What happened?

Even though [production.yml](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/production.yml#L60) has Flower configured, [traefik.toml](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/traefik.toml) doesn't contain anything related.

Flower should be working with Docker.

demestav:

[//]: # (Before you proceed:)

## Description
Fixes #2394 . 

I believe it is better to place Flower behind Traefik that supports SSL, since Flower authenticates with Basic Auth.


browniebroke:
Do we need to update the documentation to explain how to access Flower in production? It's not much detailed at the moment, it points at the instructions for the local developemnt, but given this change, it would probably deserve a section specific to production?

https://github.com/pydanny/cookiecutter-django/blob/master/docs/deployment-with-docker.rst#L28

demestav:
Indeed we need to include some information about accessing Flower. Looking at the documentation structure, I am not sure where to add it though. Currently, Flower has a dedicated paragraph about what it is and how it works in the "Getting Up and Running Locally With Docker" section. Therefore we can't have a similar paragraph in the "Deployment with Docker" section, since that would be a duplicate. So the only option left, is to just add "How to connect in Flower" sentence. @browniebroke does that make sense?

I believe we could improve the structure of the documentation; a list of available features section, independent from Local or Production environment sections. However, that is a different issue irrelevant to this pull request.

browniebroke:
Agree, no need to duplicate what is Flower in "Deployment with Docker", I was more thinking about a sentence or section explaining how to access it on Prod.

demestav:
Just realized that the cookiecutter celery conditional is missing in Traefik configuration. I will add it and commit.

browniebroke:
You're right, let me know when it's updated.

demestav:
@browniebroke I have updated the configuration file.

jameswilliams1:
@demestav Doesn't the compose also need updating to map 5555:5555 on traefik? (https://github.com/pydanny/cookiecutter-django/blob/86c9b3cb282a4e0fc9a47b2ffa991fbc72486b20/%7B%7Bcookiecutter.project_slug%7D%7D/production.yml#L44)? Have just used your traefik.yml on my server and can confirm it works as well but I had to make that change.

demestav:
Thank you @jameswilliams1, indeed the 5555 mapping needs to be defined in the Traefik.

{{cookiecutter.project_slug}}/production.yml
```diff
@@ -42,6 +42,7 @@ services:
     ports:
       - "0.0.0.0:80:80"
       - "0.0.0.0:443:443"
+      - "0.0.0.0:5555:5555"
```
browniebroke: Should this be wrapped inside a conditional to use Celery?

demestav: Of course, you are right. I missed a lot of things in this PR. I should commit when I am less tired I guess!

browniebroke:
Thanks 🎉

PR #2434, opened 2020-02-06
Include `use_drf` option on project generation Options

xpostudio4:
## Description

Include `use_drf` option on project generation Options Documentation

## Rationale

`use_drf` is missing but it appears as an option while generating a new project.

Every time you create a new project.


browniebroke:
Good catch, was added recently, but this part of the doc was missed.

browniebroke:
Thanks!

PR #2442, opened 2020-02-11
Fix broken Django Compressor setup and enable static files minification

### LINKED ISSUE #2336: Broken setup with use_whitenoise=y, use_compressor=y and cloud_provider=none
## What happened?
Multiple things:
- Altough, we set cloud_provider to None, follwing is set in `config/settings/production.py`:
`COMPRESS_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"`
- Compressed static files are not available via Whitenoise, giving an error on page:

`compressor.exceptions.OfflineGenerationError: You have offline compression enabled but key "..." is missing from offline manifest. You may need to run "python manage.py compress"`

- Compress storage should be set to local configuration:
`COMPRESS_STORAGE = "compressor.storage.GzipCompressorFileStorage"`
- `compose/production/django/start` should be extended to this:
```
python /app/manage.py collectstatic --noinput
python /app/manage.py compress --force
python /app/manage.py collectstatic --noinput
```

See https://stackoverflow.com/questions/58712195/django-whitenoise-with-compressed-staticfiles for further information

## Steps to reproduce
Set up project with following settings:
- use_docker=y
- use_whitenoise=y
- use_compressor=y
- cloud_provider=3 (None)

jameswilliams1:

[//]: # (Before you proceed:)

## Description

To change the django compressor setup as well as the startup script for django when required (if using Whitenoise).

## Rationale

I am propsing to change the django compressor setup for 2 reasons:
1) It currently does not work except when chosing S3 cloud storage
2) Currently no minification is performed on files, making it mostly redundant as this is a major feature of compressor.

The start script is modified when whitenoise is also enabled as compress must then run at start (part of the reason it currently doesn't work).

It will improve site speed by minifiying static resources on app start and also fixes a broken setup.

Fixes #2336

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/django/start
+++ b/{{cookiecutter.project_slug}}/compose/production/django/start
@@ -6,9 +6,24 @@ set -o nounset
 
 
 python /app/manage.py collectstatic --noinput
-{%- if cookiecutter.use_whitenoise == 'y' and cookiecutter.use_compressor == 'y' %}
-if [ "${COMPRESS_ENABLED:-}" = "True" ];
-then
+{% if cookiecutter.use_whitenoise == 'y' and cookiecutter.use_compressor == 'y' %}
+compress_enabled() {
+python << END
+import sys
+
+from environ import Env
+
+env = Env(COMPRESS_ENABLED=(bool, True))
+if env('COMPRESS_ENABLED'):
+    sys.exit(0)
+else:
+    sys.exit(1)
+
+END
+}
+
+if compress_enabled; then
+  # NOTE this command will fail if django-compressor is disabled
   python /app/manage.py compress
 fi
 {%- endif %}
--- a/{{cookiecutter.project_slug}}/compose/production/django/start
+++ b/{{cookiecutter.project_slug}}/compose/production/django/start
@@ -6,4 +6,10 @@ set -o nounset
 
 
 python /app/manage.py collectstatic --noinput
+{%- if cookiecutter.use_whitenoise == 'y' and cookiecutter.use_compressor == 'y' %}
+if [ "${COMPRESS_ENABLED:-}" = "True" ];
+then
+  python /app/manage.py compress
+fi
+{%- endif %}
 /usr/local/bin/gunicorn config.wsgi --bind 0.0.0.0:5000 --chdir=/app
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -299,6 +299,7 @@ Listed in alphabetical order.
 .. _@howiezhao: https://github.com/howiezhao
 .. _@IanLee1521: https://github.com/IanLee1521
 .. _@ikhomutov: https://github.com/ikhomutov
+.. _@jameswilliams1: https://github.com/jameswilliams1
 .. _@ikkebr: https://github.com/ikkebr
 .. _@Isaac12x: https://github.com/Isaac12x
 .. _@iynaix: https://github.com/iynaix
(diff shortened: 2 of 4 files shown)


jameswilliams1:
Fixes issue #2336

{{cookiecutter.project_slug}}/compose/production/django/start
```diff
@@ -6,4 +6,10 @@ set -o nounset
 
 
 python /app/manage.py collectstatic --noinput
+{%- if cookiecutter.use_whitenoise == 'y' and cookiecutter.use_compressor == 'y' %}
+if [ "${COMPRESS_ENABLED:-}" = "True" ];
```
browniebroke: I'm not sure about that... As far as I can tell, this is unset in prod, and the setting default to `True` if absent. This bash `if` will never evaluate run, will it? Shouldn't it be that instead:

```suggestion
if [ "${COMPRESS_ENABLED:-True}" = "True" ];
```

Also, the way django-environ reads the environment, it [supports several values as truthy](https://django-environ.readthedocs.io/en/latest/index.html#environ.environ.Env.BOOLEAN_TRUE_STRINGS), but this supports only one.

I think I'd be OK to keep it simple with just the template branch from line 9, but I'm open to a better idea.

jameswilliams1: Of course, I hadn't thought of that as I'm using this as an env var on my site. The issue is if compress is disabled and this line runs it causes an error, so without some check here there's no way of disabling compressor using env vars (this file would also have to change). I think the default to true should mostly work for a lot of cases though, I will change this.

jameswilliams1: @browniebroke I think this solution using environ should cover any edge cases now. Broken link also fixed.

browniebroke:
I gave this a try, but it's not working out of the box. If I choose to use compressor, the compression doesn't run by default. I think we're just missing one piece, it's to change the default environment variable here, which is unset:

https://github.com/pydanny/cookiecutter-django/blob/530868d072248b4472fb09a42e511c6024f361dd/%7B%7Bcookiecutter.project_slug%7D%7D/.envs/.production/.django#L34-L38

Since the default value is [already specified in the settings](https://github.com/pydanny/cookiecutter-django/blob/530868d072248b4472fb09a42e511c6024f361dd/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L202), and this is not anything secret, I suggest to remove the above part from the `.envs/.production/.django` entirely. Thoughts?

jameswilliams1:
> 
> 
> I gave this a try, but it's not working out of the box. If I choose to use compressor, the compression doesn't run by default. I think we're just missing one piece, it's to change the default environment variable here, which is unset:
> 
> https://github.com/pydanny/cookiecutter-django/blob/530868d072248b4472fb09a42e511c6024f361dd/%7B%7Bcookiecutter.project_slug%7D%7D/.envs/.production/.django#L34-L38
> 
> Since the default value is [already specified in the settings](https://github.com/pydanny/cookiecutter-django/blob/530868d072248b4472fb09a42e511c6024f361dd/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L202), and this is not anything secret, I suggest to remove the above part from the `.envs/.production/.django` entirely. Thoughts?

Strange, hadn't noticed that part but it makes sense to just remove it from the .django entirely (can't see any reason why someone would want it unset? probably a bug). I'll update tomrrow when at a PC.

browniebroke:
Thanks for the quick reply.

browniebroke:
Thanks 🎉

PR #2457, opened 2020-02-23
Add option to serve media files locally using nginx

### LINKED ISSUE #2449: Serving media files without an external service
## Description
Currently, in order to serve media files, you need an external file storage service. Setting up an external file storage service introduces some complexity (especially for people who do not already use these services), as well as potentially extra costs. Even though it is a proper way to serve files in production, I believe it is not a requirement for all the use-cases that, in my opinion, are within the scope of cookiecutter-django. Deployment of small-scale projects could benefit by a simpler configuration.

## Proposition
I propose to have an option to serve media files without an external service.

We are using Traefik as a reverse proxy. Traefik does not serve files (by design) therefore a web server is needed to do this job. To achieve this, we can have an Nginx behind Traefik to handle the media files route **only**.

## Comments
- The trade-off is that we add yet another docker resource in the project, making it the complete system bigger (maintenance, resources, build time etc.)
- Nginx is a beast of its own
- Someone might argue why not use Nginx exclusively (why use Traefik); that was the state previously but the project switched to Caddy and later to Traefik, and part of the reason was because it is easier and safer to get automatic SSL certificates
- I am currently testing this setup with a small project and so far seems OK.

## Closing
Would the maintainers be interested to add this? Are there any major disadvantages that I am missing?

arrys:
## Description

This PR proposes an example solution to issue #2449.
It allows selecting a local nginx instance to serve media files.

This is just a bare bones implementation that could serve as a base for serving media files. 

I added an nginx container. This container has a volume mapped where django stores its media files. Routes and services are added in the traefik configuration to route `{{domain.name}}/media/*`
requests to the nginx instance which serves them.

## Rationale

It makes deploying very simple and small scale projects easier as @demestav eloquently expressed in his issue.

When using AWS or GCP is not a good option. Serve the `/media` files using nginx.

```
/media/ --> nginx -----------|
django ---> uwsgi -----------o--> user
/static/ -> whitenoise ------|
```

Fix #2449
Closes #3260


dahifi:
@ArkadiuszMichalRys, I pulled your version and got the following error: 
```
Creating my_awesome_project_django_1_499f9d077e92   ... error

ERROR: for my_awesome_project_django_1_499f9d077e92  Cannot create container for service django: invalid volume specification: 'my_awesome_project_production_django_media:my_awesome_project/media:rw': invalid mount config for type "volume": invalid mount path: 'my_awesome_project/media' mount path must be absolute

ERROR: for django  Cannot create container for service django: invalid volume specification: 'my_awesome_project_production_django_media:my_awesome_project/media:rw': invalid mount config for type "volume": invalid mount path: 'my_awesome_project/media' mount path must be absolute

```
Looks like there's a slash missing: `services:django:volumes - production_django_media:my_awesome_project/media` should be `services:django:volumes - production_django_media:my_awesome_project/media`.

ArkadiuszMichalRys:
@dahifi, thank you for the feedback. I will look into it. The two volume configurations you posted look identical so I guess there's a typo in the second one.

ArkadiuszMichalRys:
@dahifi, I should have tested my initial changes better. There were a few issues which should now be resolved.
This time I tested the generated files without modifying anything so I couldn't forget to commit any changes I made.

sorasful:
Hi ! Thank you, I was exactly looking for something like this. I tried your setup, but have you tried it on production with https enabled ? For now I couldn't find my media. I'm digging into Nginx.
Anyway, it's a good idea.

{{cookiecutter.project_slug}}/compose/production/nginx/default.conf
```diff
@@ -0,0 +1,8 @@
+server {
+	listen       80;
+	server_name  localhost;
+	location /media {
```
sorasful, line 4: I recommand puting a trailing / to avoid path traversal exploits even if traefik will only redirect if the /media/ rule is matched, it's better.
`location /media/ {`

{{cookiecutter.project_slug}}/compose/production/nginx/default.conf
```diff
@@ -0,0 +1,8 @@
...
+		autoindex on;
```
sorasful, line 5: By default I think you probably want to disable this behavior.

ArkadiuszMichalRys:
@sorasful Thank you for your suggestions! They were all justified so I implemented them as you suggested.

maltebeckmann:
> @sorasful Thank you for your suggestions! They were all justified so I implemented them as you suggested.

Great stuff, what's happened to this cool enhancement?

arrys:
I think it might be because this seems like a very niche application. I just added it because I was using it for my own server. The repository seems to be aimed at more advanced deployments where you would really benefit from hosting the media on Amazon S3 or Google Cloud Storage.

I am happy to fix any issues in case there is something I need to add to get this merged.

maltebeckmann:
> I think it might be because this seems like a very niche application. I just added it because I was using it for my own server. The repository seems to be aimed at more advanced deployments where you would really benefit from hosting the media on Amazon S3 or Google Cloud Storage.
> 
> I am happy to fix any issues in case there is something I need to add to get this merged.

INHO, Cookiecutter leaves a hole by not providing an option for local media serving. @browniebroke What say you?

mkrtchian:
Thanks a lot! This was helpful to me.

sorasful:
Do we have news about this ? This could really be a great feature.

sorasful:
Do we need to fork this to fix conflict and merge this ?

Afrowave:
Hey @sorasful,

I suppose we will need to fork and sort it out.

Braintelligence:
Thank you very much for this PR @arrys it helped me set this up for my project. Can you give advice on what needs to be further implemented so media links can only be accessed by logged in users?

arrys:
Hopefully this helps you, @Braintelligence.

I don't know if this would be the best approach so take this with a grain of salt.
Depending on your requirements you might just make the file names hard to guess. This is not exactly secure, just a deterrent.

You could also try [authentication based on subrequest results](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/). This would be slightly more work to get implemented and might not suit your access control model.

As far as I know there is no easy way to serve the media files through nginx while using Djangos access control.
You could make Django serve the media files but this is discouraged by the Django developers.

arrys:
@sorasful, @Afrowave, Is there something I should do?
There seems to be another take on this by @ghostforpy in a different pull request: #3260.

I am not entirely sure on how to proceed here.

browniebroke:
I've updated this with the latest changes from master. I changed to using nginx as soon as cloud provider is none, which I think is pretty much what folks need?

hooks/post_gen_project.py
```diff
@@ -492,10 +492,6 @@ def main():
         )
 
     if "{{ cookiecutter.cloud_provider }}" == "None":
-        print(
-            WARNING + "You chose not to use a cloud provider, "
-            "media files won't be served in production." + TERMINATOR
-        )
```
browniebroke, line 498: I've removed that but actually we should probably leave this warning if not using docker.

Afrowave: Thank you @browniebroke. We have been waiting for this to get into the `master` branch. I have maintained my own branch based on @arrys work. Let me see what we can add to it, especially in the documentation.

browniebroke:
I've updated the documentation, could folks proof read it and let me know whether what I wrote is clear, understandable, without typos, etc...? Thanks

arrys (commented):
The documentation is pretty clear to me.

I believe there are a few typos which I corrected in bold:
1. Make sure to keep this around to avoid **losing** any media files.
2. If you choose no cloud provider and docker, the media files will be served by **an** nginx Docker service in production. Without Docker, the media files won't work.

Maybe the 'in production' could be moved earlier in the sentence. Like:
1. If you choose no cloud provider and docker, the media files in production will be served by **an** nginx Docker service.

But that's just a personal preference/nit-picking.

browniebroke:
Thanks for contributing this feature!

PR #2471, opened 2020-03-02
Add rest_framework.authtoken on installed apps.

jeromecaisip:
Add rest_framework.authtoken on installed apps when the option to use DRF is selected,  because the endpoint '/auth-token/" depends on the said app.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -79,10 +79,9 @@ THIRD_PARTY_APPS = [
 {%- if cookiecutter.use_celery == 'y' %}
     "django_celery_beat",
 {%- endif %}
-{% if cookiecutter.use_drf == "y" -%}
+{%- if cookiecutter.use_drf == "y" %}
     "rest_framework.authtoken",
 {%- endif %}
-
 ]
 
 LOCAL_APPS = [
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -79,6 +79,9 @@
 {%- if cookiecutter.use_celery == 'y' %}
     "django_celery_beat",
 {%- endif %}
+{%- if cookiecutter.use_drf == "y" %}
+    "rest_framework.authtoken",
```
browniebroke: I didn't realise that before, but surely `"rest_framework"` should also be included in this conditional.

We should also probably make it conditional in the requirements:

https://github.com/pydanny/cookiecutter-django/blob/5e16589d8edba2ad5ea97a427ac5f9ed5a3e0226/%7B%7Bcookiecutter.project_slug%7D%7D/requirements/base.txt#L32-L33

Would you mind moving the line above (78) into this and add a similar condition in the `requirements/base.txt` please?

jeromecaisip: Sure!

browniebroke: Just missing to move the `"rest_framework",` line above 😄 Thanks

jeromecaisip: Alright. Will be checking this one as well :)

browniebroke: Made the changes and merged this. Thanks!

PR #2483, opened 2020-03-11
Add PYTHONDONTWRITEBYTECODE flag to local development Dockerfile

### LINKED ISSUE #2482: Add PYTHONDONTWRITEBYTECODE flag to local development image
Given that generating `*.pyc` files on local development environments is not necessary (and creates more problems than it solves) would it make sense to add a `PYTHONDONTWRITEBYTECODE` flag to `compose/local/django/Dockerfile`?

scaramagus:

[//]: # (Before you proceed:)

## Description

This PR adds the PYTHONDONTWRITEBYTECODE environment variable to the local development Dockerfile, according to [the official docs](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE).

## Rationale

Having `.pyc` files lying around with old changes has caused me (and I can safely assume many others) lots of headaches when coding across multiple branches. This change ensures these files are not created in a local environment and therefore, only `.py` files (with the desired changes) are executed each time.

When coding on a certain git branch and working on a different feature on a different branch, the Docker image prioritizes `*.pyc` files over `*.py` files, which can cause problems when running `manage.py runserver` or executing migrations.

Fixes #2482

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -49,6 +49,7 @@ Listed in alphabetical order.
   Adam Dobrawy             `@ad-m`_
   Adam Steele              `@adammsteele`_
   Agam Dua
+  Agustín Scaramuzza       `@scaramagus`_               @scaramagus
   Alberto Sanchez          `@alb3rto`_
   Alex Tsai                `@caffodian`_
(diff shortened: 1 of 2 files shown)


browniebroke:
Nice one, thanks!

PR #2498, opened 2020-03-17
Fixed pre-commit files selector regex

### LINKED ISSUE #2478: Potentially invalid file selector pattern for pre-commit
## What happened?
The pre-commit hook `trailing-whitespace` does not seem to run against any file. This is indicated by the pre-commit report during git staging, which is as follows:

> Trim Trailing Whitespace.............................(no files to check)Skipped

Looking at the `.pre-commit-config.yaml` file, the `files` hook parameter is `(^|/)a/.+\.(py|html|sh|css|js)$`. Therefore **I am assuming** that the indented behavior is to target all the `.py`, `.html`, `.sh`, `.css` and `.js` files.

## Steps to reproduce
1) Create a new project with cookiecutter
2) `git init`
3) `pre-commit install`
4) `git add .`
5) `git commit -m "Project created"` and you should be able to see the pre-commit report as specified above.

## Comments
Maybe I am missing something but looking at the regular expression for the `files` hook parameter `(^|/)a/.+\.(py|html|sh|css|js)$`, the `a/` does not seem to belong there as it matches the `a/` characters. Is there a purpose for that? By removing that, the hook report as `passed` rather than `Skipped`.

Let me know if this issue is valid and I can submit a pull request.

demestav:
Fixes #2478

### DIFF
--- a/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
+++ b/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
     rev: master
     hooks:
       - id: trailing-whitespace
-        files: (^|/)a/.+\.(py|html|sh|css|js)$
+        files: (^|/).+\.(py|html|sh|css|js)$
 
 -   repo: local
     hooks:


browniebroke:
Thanks @demestav

PR #2503, opened 2020-03-22
Add unsupported mail_service combinations

### LINKED ISSUE #2492: Improve mail service + cloud provider combinations
## Description

See #2435, if a user selects no cloud provider & Amazon SES to send emails, the project will result in a broken setup: the Anymail backend for SES requires the AWS credentials to be present in the environment, but that won't be the case.

## Possible Solutions

1. Implement a check in our pre-generation hook to tell users that it's not supported
2. Handle this case and include credentials if either SES or cloud provider is selected

cc @Andrew-Chen-Wang

Andrew-Chen-Wang:
## Description

Add unsupported mail_service combinations
* Added default yaml Loader for pre_gen_projects.py due to a deprecation warning

## Rationale

The Amazon SES mail_service should not be used with GCP and None cloud providers as it's financially not good and AWS code does not support it due to no credentials.

Also fixed the pytest yaml loader (it's the default, but it gets rid of the deprecation warning)

The deprecation warning:
```
=============================================== warnings summary ================================================
tests/test_cookiecutter_generation.py::test_travis_invokes_pytest
  /Users/andrewwork/Private/code/projects/django/cookiecutter-django/tests/test_cookiecutter_generation.py:185: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
    assert yaml.load(travis_yml)["script"] == ["pytest"]

tests/test_cookiecutter_generation.py::test_gitlab_invokes_flake8_and_pytest
  /Users/andrewwork/Private/code/projects/django/cookiecutter-django/tests/test_cookiecutter_generation.py:201: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
    gitlab_config = yaml.load(gitlab_yml)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

## Use cases

Fixes #2492

### DIFF
--- a/docs/project-generation-options.rst
+++ b/docs/project-generation-options.rst
@@ -73,15 +73,15 @@ cloud_provider:
 mail_service:
     Select an email service that Django-Anymail provides
 
-    1. Amazon SES_
-    2. Mailgun_
+    1. Mailgun_
+    2. `Amazon SES`_
     3. Mailjet_
     4. Mandrill_
     5. Postmark_
     6. SendGrid_
     7. SendinBlue_
     8. SparkPost_
-    9. Plain/Vanilla Django-Anymail_
+    9. `Other SMTP`_
 
 use_drf:
     Indicates whether the project should be configured to use `Django Rest Framework`_.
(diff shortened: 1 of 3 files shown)


browniebroke:
Thank you!

PR #2506, opened 2020-03-24
Add uvicorn and web sockets for native Django 3

### LINKED ISSUE #2314: add uvicorn with gunicorn and django channels
## Description
Hello well my last project I decided to implement the websocket, I think it will become a standard considering what it offers in django 3.0, personally I did not remove gunicorn by daphne but use the implementation gunicorn uvicorn. I just needed to add uvicorn to gunicorn to create an asgi file and pass it instead of the wsgi file. I have not yet detected any problem in the websocket connection of the browser. 

- **choice to use websocket in start project**

Add uvicorn to gunicorn for using an "asgi" server with "wsgi" and django channels

**It is necessary on the other hand to be careful not very compatible with whitenoise (personally I have disabled it) [whitenoise conflict]. and the use with the django-toolbar (Personally also deleted I never use it anyway:), there is an alternative that is fixed on django toolbar ) [fixed toolbar] requires some adjustments. add alpine-sdk in django dockerfile for compile uvloop and httptools** 

## Rationale

The websocket and the use of asynchronous is soon implemented in django 3.0, uvicorn allows to use gunicorn which has been proven without having to install daphne for example. Use of django channels which officially becomes a django package. 

## Link
https://channels.readthedocs.io/en/latest/ # django channels
https://github.com/encode/uvicorn # Uvicorn

[whitenoise conflict]: https://channels.readthedocs.io/en/latest/installation.html
[fixed toolbar]:https://stackoverflow.com/questions/39458513/debugging-django-channels

Andrew-Chen-Wang:
## Description

Add uvicorn to Docker with Django async capabilities

Optional based on cookiecutter.json

* Add use_async option to cookiecutter.json
* Add websocket configuration

## Rationale

Django is pushing for async especially with the thought of psycopg3 becoming async too to make connecting to the database thread safe. Thus, we should, too.

Most apps are going to need async at some point whether it be through chat applications (which you can easily do using web sockets and celery + redis already installed as a message queue + broker).

SLIGHTLY Fixes #2314. I refuse to add Django-channels because Django itself is moving towards thread-safe database access. And even if you want to create a chat app, you can simply use celery + redis, which is already added via docker if you turn on that configuration, and create objects through celery tasks. Django-channels creates a new thread; too many threads and you're gonna destroy your server. With a celery and native Django 3.0 based chat, you need not create that many threads since you're limited by memory in this case. It works better if you also use `bulk_create`.

Cookiecutter-django can now support ASGI which allows for HTTP/2, meaning faster speeds.


Andrew-Chen-Wang:
Adding proper configuration. Not done.

Andrew-Chen-Wang:
@browniebroke It should work work now. Heed one thing though: you can't run the server without setting an environment variable DJANGO_SETTINGS_MODULE=config.settings.local in the uvicorn command. I also noticed it wasn't in the `.django` env file (is this a security feature?).

Andrew-Chen-Wang:
If you're wondering about the use-cases for web sockets, I've written a [gist](https://gist.github.com/Andrew-Chen-Wang/d7d9f1f118a150185caef1aa5b2da71e) that talks about how to implement a one websocket per user system for passing information asynchronously. Basically, it can help with continuously feeding users information through a system of headers. To receive events from other clients, start an asyncio loop in the web sockets.

Additionally, #2510 would come in handy. Although, that would need approval first if I were to implement it here.

Andrew-Chen-Wang:
@browniebroke PR for ASGI is finished.

Note for the future, we should leave an example of using `asgiref.sync.sync_to_async` and Django channel's `channels.db.database_sync_to_async` for other users. Additionally, I found it more useful to have `websocket.py` inside of an app instead of the config file. For me, I prefer using `sync_to_async` with Celery and for chat apps I use Redis Pub/Sub with a connection singleton per Django.

wadkar:
> @browniebroke It should work work now. Heed one thing though: you can't run the server without setting an environment variable DJANGO_SETTINGS_MODULE=config.settings.local in the uvicorn command. I also noticed it wasn't in the `.django` env file (is this a security feature?).

I was puzzled as well at first when I saw no `DJANGO_SETTINGS_MODULE` in `./envs/.{local,production}/.django` and decided to add it explicitly to my project. Would like to know the rationale/usage or a pointer to the wiki/issue/PR where this was discussed/introduced.

Personally, I would add the variable to `./envs/.{local,production}/.django` instead of _"hardcoding"_ it in `../local/start` script file because I believe it is _the_ file to store environment variables.

Andrew-Chen-Wang:
@wadkar Currently, the uvicorn command includes the environment variable `DJANGO_SETTINGS_MODULE=config.settings.local` so if you run `docker-compose -f local.yml up`, it'll default to the local settings. I'm pretty sure the reason the variable is NOT in the `.envs` is because the current `start` file for Django uses the `runserver` which SKIPS over the `wsgi.py` and uses an internal __mock__ server since the `wsgi.py` re-sets the config to production https://github.com/pydanny/cookiecutter-django/blob/async/%7B%7Bcookiecutter.project_slug%7D%7D/config/wsgi.py#L30. When you run gunicorn in production (let's say without this PR), in the wsgi.py, the settings are re-set to a default of the production settings.

TL;DR It wasn't a security feature. It was more like cookiecutter didn't need to do it since `runserver` never actually uses `wsgi.py`.

wadkar:
> TL;DR It wasn't a security feature. It was more like cookiecutter didn't need to do it since `runserver` never actually uses `wsgi.py`.

Ah, I see. So correct me if I got it wrong, but this PR replaces `runserver` with `gunicorn` inside docker for local environment and thus needs to be passed the `-e DJANGO_SETTING_MODULE=settings.config.local` explicitly.

My question is then what would be our recommendation for people developing locally without docker then? Should they rely on `runserver.py` for this? Will `runserver.py` allow the devs to do async development?

Andrew-Chen-Wang:
This PR enables ASGI/async work with native Django 3.0. It replaces `python manage.py runserver` because the `runserver` command creates a mock __synchronous__ server. In order to run asynchronously, I had to use Uvicorn. In `cookiecutter.json`, I added the option "use_async" if you wanted the async files and commands, so I didn't ENTIRELY replaces the `runserver` command; it's just not used if you want to use async. If you want to run locally, you can use the following command: 
```
/usr/local/bin/gunicorn config.asgi --bind 0.0.0.0:8000 --chdir=/app -k uvicorn.workers.UvicornWorker -e DJANGO_SETTINGS_MODULE=config.settings.local --reload
```
Really, the commands I have in the two `start` files are the commands you should probably use for your local and production settings. Again, run server cannot do async development, [as noted in Django docs](https://docs.djangoproject.com/en/3.0/howto/deployment/asgi/), so I had to substitute it with running our own server, i.e. Uvicorn.

To be fair though, you only needed to do use Uvicorn for local development, but I found it better if the mock server of Gunicorn + Uvicorn would be better for simulation. However, for production, you should use the G+U combo.

wadkar:
Thanks a lot! This is super helpful. I am working on a branch with PostGIS support [here](https://github.com/wadkar/cookiecutter-django/tree/next). The branch works on top of this PR as my use case requires an async worker.

Andrew-Chen-Wang:
@wadkar A little off topic, but you can look at my repo sample-user-location for how I set up geolocation with PostGIS. Just make sure you add DATABASES[“default”][“engine”] = “postgis” and replace postgis with the actual thing and you need to add the geolocation dependencies to the Dockerfile which’re listed in the geodjango installation docs.

Additionally, be careful with async + Django ORM using psycopg2. Django ORM is NOT thread safe, and although you can use asgiref + Django channel’s database_sync_to_async method for db calling, I do not recommend it. I’d only recommend using a Celery task to perform it to be thread safe.

{{cookiecutter.project_slug}}/config/asgi.py
```diff
@@ -0,0 +1,44 @@
...
+# This application object is used by any ASGI server configured to use this
+# file. This includes Django's development server, if the ASGI_APPLICATION
+# setting points here.
+django_application = get_asgi_application()
+# Apply ASGI middleware here.
+# from helloworld.asgi import HelloWorldApplication
+# application = HelloWorldApplication(application)
+
+# Import websocket application here, so apps from django_application are loaded first
+from .websocket import websocket_application  # noqa isort:skip
```
browniebroke, line 35: I don't understand why we cannot import it at the top, the `websocket_application` doesn't require the application to be loaded, does it? Or is it because we can expect that it _will_ as soon as some meaningful code is in there?

{{cookiecutter.project_slug}}/config/websocket.py
```diff
@@ -0,0 +1,13 @@
+async def websocket_application(scope, receive, send):
```
browniebroke, line 1: Should this be defined outside of the `config`, in one of the apps maybe? It sounds like it's more part of the application than the config...

{{cookiecutter.project_slug}}/config/asgi.py
Andrew-Chen-Wang: I believe from past experience when importing code from the app, I would always run into errors like apps weren't loaded. When I created a chat app with Redis Pub/Sub, especially, I had to access celery tasks with `asgiref.sync.sync_to_async` to run some db commands (also possible with database_sync_to_async from Django channels) and also access some code from other places. Put it simply, it's best if the import is left there.

{{cookiecutter.project_slug}}/config/websocket.py
Andrew-Chen-Wang: I was debating on that, and yes, it's probably best if it's in the users app.

Andrew-Chen-Wang: I looked back at my previous code, and I'm not sure if I should (or can) import a little bit of what I've done. The gist of my dilemma is that I created a custom async_urlpatterns as a dict instead of a list with their values as dicts that represent each websocket type, i.e. connect, disconnect, and receive. This value dicts have functions that take the same scope, receive, and send parameters. That's probably the reason why I initially had `websocket.py` in config instead of users app.

However, that's websocket pioneering which probably isn't meant to be in this repo; on the other hand, I don't like having websocket_application in the users app simply because people will think a client (whether or not malicious) can connect to any websocket URL.

So I suppose I'm just going to leave it in config for now...

Andrew-Chen-Wang:
@browniebroke Thanks for taking a look and testing it! I updated the docs for local development. I didn't really test the Gulp configuration since I'm not exactly sure how to use Gulpfile.js. Should I remove my edits to it?

{{cookiecutter.project_slug}}/config/asgi.py
```diff
@@ -0,0 +1,44 @@
...
+
+# This allows easy placement of apps within the interior
+# {{ cookiecutter.project_slug }} directory.
+app_path = Path(__file__).parents[1].resolve()
+sys.path.append(str(app_path / "{{ cookiecutter.project_slug }}"))
+# We defer to a DJANGO_SETTINGS_MODULE already in the environment. This breaks
+# if running multiple sites in the same mod_wsgi process. To fix this, use
+# mod_wsgi daemon mode with each site in its own daemon process, or use
+# os.environ["DJANGO_SETTINGS_MODULE"] = "config.settings.production"
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.production")
```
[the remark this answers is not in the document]
browniebroke: I can see you've removed completely this line, but don't we need to default to local settings?

```python
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")
```

browniebroke:
Good spot with Gulp, I will take a look to that part as well.

{{cookiecutter.project_slug}}/config/asgi.py
```diff
@@ -0,0 +1,36 @@
...
+# This allows easy placement of apps within the interior
+# {{ cookiecutter.project_slug }} directory.
+app_path = Path(__file__).parents[1].resolve()
+sys.path.append(str(app_path / "{{ cookiecutter.project_slug }}"))
+
+# This application object is used by any ASGI server configured to use this file.
+django_application = get_asgi_application()
+# Apply ASGI middleware here.
+# from helloworld.asgi import HelloWorldApplication
+# application = HelloWorldApplication(application)
```
bahmdev, line 24: if `use_sentry` and `use_async` are enabled, [SentryAsgiMiddleware](https://docs.sentry.io/platforms/python/asgi/) needs to be used / applied. I can add this if needed. 👏

Andrew-Chen-Wang:
That'd be helpful! @bahmdev Can you make a PR in my repo? Not exactly sure how to integrate it since it's probably gonna be a bit of fixing up LOGGING...

bahmdev:
@Andrew-Chen-Wang I can do that, but I might not get to it until later this week.

{{cookiecutter.project_slug}}/config/asgi.py
browniebroke: Good point, I don't mind merging it wihtout, we can add it in a further PR I think.

browniebroke:
Thanks a lot for this @Andrew-Chen-Wang 🎉

Andrew-Chen-Wang:
@bahmdev Any updates? Or any tips on how I could do it myself? I'm just not sure how to using Sentry.

wadkar:
@Andrew-Chen-Wang Based on the documentation it seems all you need to do is add the sentry middleware after you get the application in [`asgi.py` file](https://github.com/pydanny/cookiecutter-django#diff-d617f4b4d9e30e21265c3228560b6228R21):

```python
import sentry_sdk
from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
# Not sure about the environment part in local/dev
# but typically you are expected to have SENTRY_DSN declared in the environment
SENTRY_DSN = env("SENTRY_DSN")
sentry_sdk.init(dsn="SENTRY_DSN")
...

django_application = get_asgi_application()
django_application = SentryAsgiMiddleware(django_application)
```

Andrew-Chen-Wang:
@wadkar It doesn't seem like it's even necessary. According to getsentry/python-sentry, it seems like the Django integration automatically looks for the handler (either WSGI or ASGI). I haven't tested it, but, since the connect method for websockets is essentially an HTTP request, I'm assuming it would work correctly for initial request. However, an internal error could be different. I'm pretty sure if you've got any kind of internal error, though, it would be caught by the Sentry middleware regardless.

So I guess the above wouldn't be necessary.

stefanitsky:
@Andrew-Chen-Wang great work man

PR #2509, opened 2020-03-27
docs: Fix simple typo, interpteter -> interpreter

timgates42:
There is a small typo in {{cookiecutter.project_slug}}/docs/pycharm/configuration.rst.

Should read `interpreter` rather than `interpteter`.


browniebroke:
Thanks 🎉

PR #2240, opened 2019-10-01
Support python 3.7

nicolas471:
[//]: # (Thank you for helping us out: your efforts mean great deal to the project and the community as a whole!)

[//]: # (Before you proceed:)

[//]: # (1. Make sure to add yourself to `CONTRIBUTORS.rst` through this PR provided you're contributing here for the first time)
[//]: # (2. Don't forget to update the `docs/` presuming others would benefit from a concise description of whatever that you're proposing)

## Description

[//]: # (What's it you're proposing?)

Update python version from 3.6 to 3.7
Described in issue #2236 

## Rationale

[//]: # (Why does the project need that?)

Fixes #2236 

## Use case(s) / visualization(s)

[//]: # ("Better to see something once than to hear about it a thousand times.")

PR #2514, opened 2020-03-29
Upgrade to Python 3.8

### LINKED ISSUE #2309: Upgrade to Python 3.8
Python 3.8 is out, and it would be nice to upgrade to it. We would need to check that all of our dependencies support it.

[The previous upgrade was done recently](https://github.com/pydanny/cookiecutter-django/pull/2240), we would need to do something similar to start with, and potentially fix issues that come up.

codnee:
## Description

Changes Python version to 3.8 everywhere where it was changed to 3.7 in the previous Python upgrade (https://github.com/pydanny/cookiecutter-django/pull/2240)

## Rationale

Closes #2309

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -115,7 +115,7 @@ Listed in alphabetical order.
   Emanuel Calso            `@bloodpet`_                  @bloodpet
   Eraldo Energy            `@eraldo`_
   Eric Groom               `@ericgroom`_
-  Ernesto Cedeno           `@codnee`
+  Ernesto Cedeno           `@codnee`_
   Eyad Al Sibai            `@eyadsibai`_
   Felipe Arruda            `@arruda`_
   Florian Idelberger       `@step21`_                    @windrush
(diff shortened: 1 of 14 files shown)


CONTRIBUTORS.rst
```diff
@@ -115,6 +115,7 @@ Listed in alphabetical order.
   Emanuel Calso            `@bloodpet`_                  @bloodpet
   Eraldo Energy            `@eraldo`_
   Eric Groom               `@ericgroom`_
+  Ernesto Cedeno           `@codnee`
```
browniebroke: If you want this to be a link to your profile (like the others) you should 

1) add an underscore after:

   ```suggestion
     Ernesto Cedeno           `@codnee`_
   ```

2) and define [the target later in the file](https://github.com/codnee/cookiecutter-django/blob/python_3.8/CONTRIBUTORS.rst#L273) (by adding this`.. _@codnee: https://github.com/codnee`)

browniebroke:
Looking good otherwise! 🎉

browniebroke:
Thanks a lot for your contibution!

PR #2522, opened 2020-04-03
Patch 1

gwiskur:

[//]: # (Before you proceed:)

## Description

Add ``compress`` command to docs for when using Django Compressor with a PaaS

## Rationale

I ran into this error when deploying to a PaaS; this PR should keep others from running into the same error.

> You have offline compression enabled but key "asdfa..." is missing from offline manifest

image


browniebroke:
Patch looks good, but it's done against the `docs-v2` branch, which is very old and out of date. Could you please base it on master? Thanks.

browniebroke:
Superseded by #2523

PR #2524, opened 2020-04-04
Traefik - Also redirect www. to service.

sorasful:

[//]: # (Before you proceed:)

## Description
I've recently tried to deploy my own project and noticed that some browsers like Google Chrome, Firefox or other redirect automatically when you type www.YOURDOMAIN, I believe they ignore the www and go directly to http://YOURDOMAIN or https://YOURDOMAIN. 
But for other browsers like IE for instance, if you go to www.YOURDOMAIN, you will get an certificate insecure error because www.YOURDOMAIN doesn't match the rule.

To add an exception for www subdomain in traefik rules.

## Rationale

More compatibility with browsers and people often try to access www.YOURDOMAIN .

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -200,6 +200,7 @@ Listed in alphabetical order.
   Sascha                   `@saschalalala`_             @saschalalala
   Shupeyko Nikita          `@webyneter`_
   Sławek Ehlert            `@slafs`_
+  Sorasful                 `@sorasful`_
   Srinivas Nyayapati       `@shireenrao`_
   stepmr                   `@stepmr`_
   Steve Steiner            `@ssteinerX`_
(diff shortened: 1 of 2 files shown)


wadkar:
Super useful, ended up manually adding it.

browniebroke:
Thanks for your contribution 🎉

wadkar:
Rare, but in case you end up entering sudomains instead of domain name during the cookiecutter prompts, say `domain_name: "api.mydomain.com"` you will get an extra `www.api.mydomain.com` level routing. Maybe a low priority issue can be added to patch this PR so that the additional `www` rule is added when the host has just one `.` character?

browniebroke:
Good point 🤔 

> the host has just one `.` character

That wouldn't work for TLD with 2 parts, for example `.co.uk`...

wadkar:
Good point as well 🤔

I have created a PR where we can discuss possible solutions. Unless we want to revert this PR?

PR #2528, opened 2020-04-07
DEBUG TOOLBAR - fix INTERNAL_IPS discovery for docker

dudanogueira:
A simple fix for the INTERNAL_IPS when using DOCKER

## Rationale

For users with multiple containers, the DEBUG TOOL BAR stops showing up if the django container receives an IP higher than 9, ex 172.19.0.10, it becomes 172.19.0.11, not complying with the INTERNAL_IPS requirement.

The actual code doesn't consider this scenario. This fix will work for all scenarios.

Users that has a stack with numerous containers


browniebroke:
Thanks for the fix 🎉

PR #2533, opened 2020-04-12
Store coverage config in setup.cfg

danihodovic:
Less is more. Unify tooling configuration in one file.

https://coverage.readthedocs.io/en/v4.5.x/config.html

### DIFF
--- a/{{cookiecutter.project_slug}}/.coveragerc
+++ b/{{cookiecutter.project_slug}}/.coveragerc
@@ -1,5 +1,0 @@
-[run]
-include = {{cookiecutter.project_slug}}/*
-omit = *migrations*, *tests*
-plugins =
-    django_coverage_plugin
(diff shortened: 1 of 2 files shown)


browniebroke:
I'm in favour of reducing the number of config files, so sounds good to me 👍 

One thing I learned recently is that the [latest version also supports `pyproject.toml` as config file](https://coverage.readthedocs.io/en/stable/config.html), which seems to become [the new standard](https://snarky.ca/what-the-heck-is-pyproject-toml/). Although it's not supported by all tools yet (at least mypy isn't). Anyway, this looks good.

danihodovic:
> One thing I learned recently is that the latest version also supports pyproject.toml as config file, which seems to become the new standard. Although it's not supported by all tools yet (at least mypy isn't). Anyway, this looks good.

Funny that you mention it. I was thinking about opening a PR for migrating to pyproject.toml (and using Poetry). While the community seems to be moving in that direction, it's not supported yet for mypy and pytest. We would end up in a strange intermediary state where we use pyproject.toml, but still maintain additional config files.

Let's wait until the major projects move to pyproject.toml before starting this effort.

browniebroke:
Yes, agree. Especially since after writing the above, I discovered that [adding support in flake8 is on hold for now](https://gitlab.com/pycqa/flake8/-/issues/428).

Issue #2100, opened 2019-06-06
Travis + Docker result in broken build on CI


browniebroke:
## What happened?

Try to generate a project with Docker and Tarvis enabled, the tarvis config is not suitable and fails the build:

<details>
  <summary>Build logs</summary>
  
```
[0Ktravis_fold:end:install
[0Ktravis_time:start:183a4568
[0K$ pytest
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/environ/environ.py", line 273, in get_value
    value = self.ENVIRON[var]
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'DATABASE_URL'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/bin/pytest", line 10, in <module>
    sys.exit(main())
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/_pytest/config/__init__.py", line 63, in main
    config = _prepareconfig(args, plugins)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/_pytest/config/__init__.py", line 207, in _prepareconfig
    pluginmanager=pluginmanager, args=args
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/hooks.py", line 289, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/manager.py", line 87, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/manager.py", line 81, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/callers.py", line 203, in _multicall
    gen.send(outcome)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/_pytest/helpconfig.py", line 94, in pytest_cmdline_parse
    config = outcome.get_result()
[...]

PR #2535, opened 2020-04-13
Add Docker support to Travis configuration for projects

Andrew-Chen-Wang:

[//]: # (Before you proceed:)

## Description

Add Docker support to the Travis configuration.

Edits allowed by maintainers meaning you can add commits to this patch, too.

## Rationale

Builds automatically fail on first push

No more fails on first push...

### DIFF
--- a/{{cookiecutter.project_slug}}/.travis.yml
+++ b/{{cookiecutter.project_slug}}/.travis.yml
@@ -1,4 +1,8 @@
 dist: xenial
+language: python
+python:
+  - "3.8"
+{% if cookiecutter.use_docker == 'y' %}
 services:
   - postgresql
 before_install:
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
Failures usually arise with the lack of a database URL, but... it seems like this needs more of a fork branch than a simple patch apparently.

browniebroke:
The patch looked pretty good, and it would solve #2100. I'm not sure I understand your comment, it seems to be about a different issue?

Andrew-Chen-Wang:
Whoops, I didn’t properly explain myself. The pre-gen test hook I believe ran into an error if you setup travis with those commands. My current configuration for Travis is a little different because if you were to only

`docker-compose build && docker-compose up`

Then you’d run into an error with celerybeat not having tables existing yet. I added entrypoint’s postgres_ready() function to beat’s start files and added a SQL statement to make sure its tables existed first before starting, but I was debating if that would take up too much costs and connection pool would increase substantially on a Blue/Green deployment for several clusters. You can take a look at https://github.com/Andrew-Chen-Wang/cookiecutter-django-ecs-deploy for what I did (also archiving it because I realized there’s probably no way ECS works with cookiecutter django... still looking).

PR #2547, opened 2020-04-17
Improve addition of www redirect to Traefik

wadkar:
Note: Ouch, I deleted the PR template stuff accidentally and a hard refresh won't reload it in the editor.

# Description

PR #2524 adds an additional `www.{{ cookiecutter.domain_name }}` to the traefik route. However, this helpful redirect may not be useful if the `domain_name` itself is a subdomain, e.g. `api.example.com`. In this case an additional `www.api.example.com` route is added to traefik. This PR provides a best effort check for detecting subdomain usage and conditionally adds `www.{{ cookiecutter.domain_name }}` to the traefik route.

# Rationale

As per [this discussion](https://github.com/pydanny/cookiecutter-django/pull/2524#issuecomment-615390917)

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
+++ b/{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
@@ -26,10 +26,9 @@ certificatesResolvers:
         entryPoint: web
 
 http:
-  {%- set domain_dots = cookiecutter.domain_name.count('.') %}
   routers:
     web-router:
-      {%- if domain_dots == 1 or (domain_dots == 2 and '.co.' in cookiecutter.domain_name) %}
+      {%- if cookiecutter.domain_name.count('.') == 1 %}
       rule: "Host(`{{ cookiecutter.domain_name }}`) || Host(`www.{{ cookiecutter.domain_name }}`)"
       {% else %}
       rule: "Host(`{{ cookiecutter.domain_name }}`)"
@@ -42,7 +41,7 @@ http:
       service: django
 
     web-secure-router:
-      {%- if domain_dots == 1 or (domain_dots == 2 and '.co.' in cookiecutter.domain_name) %}
+      {%- if cookiecutter.domain_name.count('.') == 1 %}
       rule: "Host(`{{ cookiecutter.domain_name }}`) || Host(`www.{{ cookiecutter.domain_name }}`)"
       {% else %}
       rule: "Host(`{{ cookiecutter.domain_name }}`)"
--- a/{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
+++ b/{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
@@ -26,9 +26,10 @@ certificatesResolvers:
         entryPoint: web
 
 http:
+  {%- set domain_dots = cookiecutter.domain_name.count('.') %}
   routers:
     web-router:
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
```diff
@@ -26,9 +26,14 @@ certificatesResolvers:
         entryPoint: web
 
 http:
+  {%- set domain_dots = cookiecutter.domain_name.count('.') %}
   routers:
     web-router:
+      {%- if domain_dots == 1 or (domain_dots == 2 and '.co.' in cookiecutter.domain_name) %}
```
browniebroke, line 32: I've done a bit of research, and I now believe that checking for `.co.` isn't enough. In Brasil, the `.com.br` is very common by [looking on Wikipedia](https://en.wikipedia.org/wiki/Second-level_domain), I can see there are a lot of variations across the world.

wadkar: Yeah, as I mentioned in the PR/commit this is a best effort detection. Note that the successful detection results in addition of `www` route, so we mostly err on the side of _not adding_ `www` route when the detection fails.

wadkar:
There exists a python package for getting the TLDs/STLDs: [tldextract](https://pypi.org/project/tldextract/) , but this adds a pip dependency which I think is out of question?

I am fine with reverting #2524 , or if we want to add the `www` route on best effort basis, we can merge this in.

@sorasful : thoughts?

browniebroke:
External dependency is out of question yes, it would be needed when running the cookiecutter CLI. 

I see your point, we'll default to the old behaviour if we're not sure. In this case I think we can just go with the most naive approach and add the `www` only if we have only one dot as you initially suggested.

IMO we should either treat all special second level TLDs or none.

wadkar:
Sounds good, I will add another commit that'll add `www` route if there's just one dot in the `domain_name` variable.

wadkar:
Not sure why the Labeler fails, I am not familiar with this project's CI/CD setup. Let me know if there's something wrong in my commits and I will be happy to fix it.

browniebroke:
Yes, you may ignore this. There was some syntax errors in the CI config file initially. If you merge merge master in you branch it'll be fixed.

sorasful:
Hi, I've read all your messages. I didn't think about the cases you mentionned, yep I agree with you with keeping things simple, even if that means reverting. The option to count 1 dot is cool IMO.

wadkar:
@browniebroke I rebased my branch off upstream/master but the labeler still chokes. I am going to ignore it and be happy that Travis-CI doesn't break.

I thought about adding tests and all, but this project is unusual in it's goal that writing tests for these things is rather impractical if not downright impossible. I mean _really "testing"_ if `www` domain resolves in a test environment is a long yak shaving process that leads into lets encrypt's testing facilities (I am not digging into [pebble](https://github.com/letsencrypt/pebble#docker) ) with custom `/etc/hosts` to ensure some variations of `curl https://www.fake.root` resolves and so on and so forth. Ugh, no thanks right now 😝 

On other hand, those tests could be nice infra tests for the resulting project. So maybe there's another PR in the making (someday, soon, hopefully, 😃 )

browniebroke:
Hum, odd, yes the error was different this time, I might just remove it actually. It looks like it's broken...

Anyway, thanks for the fix!

PR #2554, opened 2020-04-21
Hot-reload support to celery

yotamtal:

[//]: # (Before you proceed:)

## Description

This PR adds support for celery hot-reloading while developing, on changes to tasks.py and celery_app.py files when setting up a project with docker and celery.

## Rationale

I have been using this Cookiecutter template for many projects and while developing with celery tasks I had to restart the worker manually to test the changes. This minor change will allow developers to test their celery tasks faster

*-> Develop celery tasks -> notice a bug -> change the code -> save -> test the new code without restarting the worker manually*


foarsitter:
Nice addition! 

Is the pattern not to tight? Now the restarts do not happen when there are changes made to settings.py or models.py. So why not set it to *.py?

yotamtal:
@foarsitter I configured it that way on purpose, since I write all my tasks in those files.
I tried setting it as *.py before and then it restarts the worker on every save and the hot reload takes more time and mess up the logs, so I think this is the most efficient way to go :)

browniebroke:
Nice addition indeed. Good question for the pattern, I'm not sure I undertsnad the problem, but I probably just need to try it out 😄

pySilver:
Nice work! Consider using `watchgod` instead of `watchdog`.

{{cookiecutter.project_slug}}/compose/local/django/celery/worker/start
```diff
@@ -4,4 +4,4 @@ set -o errexit
 set -o nounset
 
 
-celery -A config.celery_app worker -l INFO
+watchmedo auto-restart --directory=./ --pattern="*tasks.py;*celery_app.py" --recursive -- celery -A config.celery_app worker -l INFO
```
browniebroke, line 7: Looks like we can do it with `watchgod` which is already installed if we use async:

```suggestion
watchgod celery.__main__.main --args -A config.celery_app beat -l INFO
```

foarsitter: `beat` has to be `worker` I guess?

browniebroke: Yes not sure why I suggested beat, that's definitely wrong

{{cookiecutter.project_slug}}/compose/local/django/celery/worker/start
```diff
@@ -4,4 +4,4 @@ set -o errexit
 set -o nounset
 
 
-celery -A config.celery_app worker -l INFO
+watchgod celery.__main__.main --args -A config.celery_app beat -l INFO
```
browniebroke, line 7:
```suggestion
watchgod celery.__main__.main --args -A config.celery_app worker -l INFO
```

scnerd:
This is a great feature idea. Is there anything still needed before this can be merged (besides simply resolving the merge conflicts)?

browniebroke:
Yes, I kind of forgot about it, I think it's good to go.

browniebroke:
Thanks for your contribution @yotamtal 🎉

PR #2559, opened 2020-04-25
Update django-redis links

tanoabeleyra:
**django-redis** was moved to the Jazzband organization ([related issue](https://github.com/jazzband/django-redis/issues/438))


browniebroke:
Thanks!

PR #2567, opened 2020-04-28
Add tests for DRF

### LINKED ISSUE #2415: Write tests for the DRF setup
## Description

A new option has been added to bring support for DRF in #2224.
However, this is currently not tested. We should write tests covering various components of the DRF option and remove them when the user opts out.

PilhwanKim:
## Description

Add test case, when `use_drf` option is on.

- Test for drf user urls
  - {{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_urls.py

- Tests for drf user views
  - {{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py

- If use_drf is off, these two test case files remove.

## Rationale

Fixes #2415

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -299,6 +299,8 @@ def remove_aws_dockerfile():
 def remove_drf_starter_files():
     os.remove(os.path.join("config", "api_router.py"))
     shutil.rmtree(os.path.join("{{cookiecutter.project_slug}}", "users", "api"))
+    os.remove(os.path.join("{{cookiecutter.project_slug}}", "users", "tests", "test_drf_urls.py"))
+    os.remove(os.path.join("{{cookiecutter.project_slug}}", "users", "tests", "test_drf_views.py"))
 
 
 def remove_storages_module():
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -299,8 +299,16 @@ def remove_aws_dockerfile():
 def remove_drf_starter_files():
     os.remove(os.path.join("config", "api_router.py"))
     shutil.rmtree(os.path.join("{{cookiecutter.project_slug}}", "users", "api"))
-    os.remove(os.path.join("{{cookiecutter.project_slug}}", "users", "tests", "test_drf_urls.py"))
-    os.remove(os.path.join("{{cookiecutter.project_slug}}", "users", "tests", "test_drf_views.py"))
+    os.remove(
+        os.path.join(
+            "{{cookiecutter.project_slug}}", "users", "tests", "test_drf_urls.py"
+        )
+    )
+    os.remove(
+        os.path.join(
+            "{{cookiecutter.project_slug}}", "users", "tests", "test_drf_views.py"
+        )
+    )
 
 
 def remove_storages_module():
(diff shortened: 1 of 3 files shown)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+from django.test import RequestFactory
+
+from {{ cookiecutter.project_slug }}.users.models import User
+from {{ cookiecutter.project_slug }}.users.api.views import UserViewSet
+from {{ cookiecutter.project_slug }}.users.api.serializers import UserSerializer
```
browniebroke, line 6: This is causing an isort error in the flake8 checks, should be ordered as below:

```suggestion
from {{ cookiecutter.project_slug }}.users.api.serializers import UserSerializer
from {{ cookiecutter.project_slug }}.users.api.views import UserViewSet
from {{ cookiecutter.project_slug }}.users.models import User
```

hooks/post_gen_project.py
```diff
@@ -299,6 +299,8 @@ def remove_aws_dockerfile():
 def remove_drf_starter_files():
     os.remove(os.path.join("config", "api_router.py"))
     shutil.rmtree(os.path.join("{{cookiecutter.project_slug}}", "users", "api"))
+    os.remove(os.path.join("{{cookiecutter.project_slug}}", "users", "tests", "test_drf_urls.py"))
+    os.remove(os.path.join("{{cookiecutter.project_slug}}", "users", "tests", "test_drf_views.py"))
```
browniebroke, line 303: This is failing some checks for black, the lines are too long. Can be fixed as follow:

```suggestion
    os.remove(
        os.path.join(
            "{{cookiecutter.project_slug}}", "users", "tests", "test_drf_urls.py"
        )
    )
    os.remove(
        os.path.join(
            "{{cookiecutter.project_slug}}", "users", "tests", "test_drf_views.py"
        )
    )
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_me(self, user: User, rf: RequestFactory):
+        view = UserViewSet()
+        request = rf.get("/fake-url/")
+        request.user = user
+
+        view.request = request
+
+        response = view.me(request)
+
+        assert response.data == UserSerializer(user, context={"request": request}).data
```
browniebroke, line 30: Ideally, this would not use the serializer to produce the expected data, can we compare to a dictionary instead? Something along these lines:

```suggestion
        assert response.data == {
            "username": user.username, 
            "email": user.email, 
            "name": user.name, 
            "url": f"/api/users/{user.username}/",
        }
```

PilhwanKim: OK. I agree with your idea. Even to me, this code is more explicit to check result.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py
PilhwanKim: OK. I mssed isort error.

browniebroke:
Thanks for your contribution 🎉

PR #2572, opened 2020-04-28
Fix reloading when using async

### LINKED ISSUE #2571: gunicorn is not reload itself
Not sure whether I should report this to `cookiecutter-django` or `gunicorn`
Currently my `gunicorn` is not `reload` after code change like this [guy has](https://stackoverflow.com/questions/58958853/gunicorn-reload-not-detecting-changes-in-application-code)

`gunicorn==20.0.4          # via -r requirements.in`
I use OSX 10.15.3 and run `docker-compose -f local.yml up`

Are there any people else except I and him having this problem?

Andrew-Chen-Wang:
## Description

Since gunicorn plus uvicorn worker combo wouldn't reload. My fault; Uvicorn has a built-in StatReloader but I failed to make it work until today.

## Rationale

Fixes #2571

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/local/django/start
+++ b/{{cookiecutter.project_slug}}/compose/local/django/start
@@ -7,7 +7,7 @@ set -o nounset
 
 python manage.py migrate
 {%- if cookiecutter.use_async == 'y' %}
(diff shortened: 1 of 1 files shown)


browniebroke:
Looks good, tested and it works. Thanks!

PR #2576, opened 2020-04-29
Add rabbitmq for celery

Andrew-Chen-Wang:

[//]: # (Before you proceed:)

## Description

Add RabbitMQ for Celery instead of Redis

## Rationale

Instead of using Redis as the broker, it's best to use RabbitMQ since there's disk failover.

To this day, the celery docs still say to use RabbitMQ. The result backend uses the performant RPC. If Redis was the broker, and your machine failed for some reason, you can still extract your tasks that you might've delayed for a long time and re-run it, whereas the Redis info would be lost.

### DIFF
--- a/{{cookiecutter.project_slug}}/.envs/.local/.django
+++ b/{{cookiecutter.project_slug}}/.envs/.local/.django
@@ -11,6 +11,9 @@ REDIS_URL=redis://redis:6379/0
 # Celery
 # ------------------------------------------------------------------------------
 
+# RabbitMQ
+RABBITMQ_URL=amqp://user:bitnami@rabbitmq:5672/
+
 # Flower
 CELERY_FLOWER_USER=!!!SET CELERY_FLOWER_USER!!!
 CELERY_FLOWER_PASSWORD=!!!SET CELERY_FLOWER_PASSWORD!!!
--- a/{{cookiecutter.project_slug}}/.envs/.production/.django
+++ b/{{cookiecutter.project_slug}}/.envs/.production/.django
@@ -65,6 +65,9 @@ REDIS_URL=redis://redis:6379/0
 # Celery
 # ------------------------------------------------------------------------------
 
+# RabbitMQ
+RABBITMQ_URL=amqp://user:bitnami@rabbitmq:5672/
+
 # Flower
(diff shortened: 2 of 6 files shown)


browniebroke:
I'm not sure about this change...

I've used Redis with Celery in a couple of Django projects and it has never been an issue. The fact that I can reuse this for caching it a big plus for me, I can avoid to introduce another component to the stack. With this change, we would need both Redis and RabbitMQ with this change.

Andrew-Chen-Wang:
Hmm, it was just PTSD for me due to a humongous Celery workload that I just needed to dump to RabbitMQ since one of my instances had failed when using Redis. If anything, Celery pushes everyone to use RabbitMQ, and it's very clear in their docs. But I can understand a heavy component is not something everyone would like, especially RabbitMQ on OS X can be... well not small (230 MB image but huge volume data being GBs large).

wadkar:
Maybe we can provide both options and let the user decide? For example,

```python
>>> do you want to enable celery?
1. No
2. Yes with Redis (default/recommended for easy setup)
3. Yes with RabbitMQ (heavy/doesn't provide caching)
```

And the templates, will have a conditional on this?

Andrew-Chen-Wang:
Sure, I've been having trouble with the bitnami/rabbitmq set up, anyways, so I'll just leave the PR as it. For example, in the flower start file, I had to add a function that checks if the celery workers are up and running first or else the containers exit. Plus, I don't know if it's just me, but RabbitMQ will start up then shutdown, making celery drop connection which is slightly annoying.

wadkar:
Yeah, the last time I followed Celery docs and tried to setup rabbitmq it didn't go that well (the other kind of scars). So I decided to stick with redis.

Let's just leave this PR here in case we want to revisit this in future.

bahmdev:
I believe Redis provides some form of persistence through snapshots, but if that's not enough, one could also configure an Append Only File. All memory changes would get logged. On failure, Redis would essentially run through the logged operations and reconstruct the state in memory.

browniebroke:
I'm going to close this PR for the reasons mentioned above. We can always re-open if we change our mind later. Thanks for the suggestion @Andrew-Chen-Wang

PR #2577, opened 2020-04-29
Use hiredis

Andrew-Chen-Wang:

[//]: # (Before you proceed:)

## Description

Using hiredis, which is developed by the redis community, too, is probably best performance wise.

Benchmarks (came when I was using aioredis for websocket pubsub): https://github.com/popravich/python-redis-benchmark

## Rationale

Since Redis is already used in this project, why not use hiredis? Windows is not supported by Redis and Hiredis; however, Redis could be supported via Docker.

Another thing is most production-ready Django projects don't use Windows servers. So I've put hiredis as a mandatory thing in production (since celery uses redis in local, too, so separate occasions for the hiredis requirements).

[More performant redis](https://github.com/popravich/python-redis-benchmark)

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -36,6 +36,7 @@ CACHES = {
             # Mimicing memcache behavior.
             # http://jazzband.github.io/django-redis/latest/#_memcached_exceptions_behavior
             "IGNORE_EXCEPTIONS": True,
+            "PARSER_CLASS": "redis.connection.HiredisParser",
         },
     }
 }
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -36,7 +36,6 @@ CACHES = {
             # Mimicing memcache behavior.
             # http://jazzband.github.io/django-redis/latest/#_memcached_exceptions_behavior
             "IGNORE_EXCEPTIONS": True,
-            "PARSER_CLASS": "redis.connection.HiredisParser",
         },
     }
 }
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -9,7 +9,9 @@ argon2-cffi==19.2.0  # https://github.com/hynek/argon2_cffi
 whitenoise==5.0.1  # https://github.com/evansd/whitenoise
 {%- endif %}
 redis==3.4.1  # https://github.com/andymccurdy/redis-py
+{%- if cookiecutter.windows == "n" %}
 hiredis==1.0.1  # https://github.com/redis/hiredis-py
+{%- endif %}
 {%- if cookiecutter.use_celery == "y" %}
 celery==4.4.2  # pyup: < 5.0  # https://github.com/celery/celery
 django-celery-beat==2.0.0  # https://github.com/celery/django-celery-beat
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -9,7 +9,7 @@ argon2-cffi==19.2.0  # https://github.com/hynek/argon2_cffi
 whitenoise==5.0.1  # https://github.com/evansd/whitenoise
 {%- endif %}
 redis==3.4.1  # https://github.com/andymccurdy/redis-py
-{%- if cookiecutter.windows == "n" %}
+{%- if cookiecutter.use_docker == "y" or cookiecutter.windows == "n" %}
 hiredis==1.0.1  # https://github.com/redis/hiredis-py
 {%- endif %}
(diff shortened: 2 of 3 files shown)


jcass77:
I was always under the impression that redis-py would revert to [using the `HiredisParser`](https://github.com/andymccurdy/redis-py#parsers) automatically if it was available?

Are there some specific implementation details in the Django cache that requires this to be done explicitly? (I haven't checked).

Andrew-Chen-Wang:
redis-py would, but [I don't think Django-redis does](http://jazzband.github.io/django-redis/latest/#_pluggable_parsers). I've never tested it, but I couldn't find an issue saying Django-redis was default using it.

jcass77:
Looks like django-redis leaves this up to redis-py: https://github.com/jazzband/django-redis/blob/dceb86fb0a25edbd966a27f797121e3dd60674b6/django_redis/pool.py#L82-L86

...and `DefaultParser` reverts to HiRedis if available: https://github.com/andymccurdy/redis-py/blob/09da3ad07f26350d093929fe1ea6eaaca79ab944/redis/connection.py#L485-L488

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -36,6 +36,7 @@
             # Mimicing memcache behavior.
             # http://jazzband.github.io/django-redis/latest/#_memcached_exceptions_behavior
             "IGNORE_EXCEPTIONS": True,
+            "PARSER_CLASS": "redis.connection.HiredisParser",
```
browniebroke, line 39: If the user choose `windows = 'y'` [this would raise an error](https://github.com/andymccurdy/redis-py/blob/09da3ad07f26350d093929fe1ea6eaaca79ab944/redis/connection.py#L370). 

Given that the point raised by @jcass77, it's probably simpler to leave this out.

{{cookiecutter.project_slug}}/requirements/base.txt
```diff
@@ -9,6 +9,9 @@ argon2-cffi==19.2.0  # https://github.com/hynek/argon2_cffi
 whitenoise==5.0.1  # https://github.com/evansd/whitenoise
 {%- endif %}
 redis==3.5.0 # https://github.com/andymccurdy/redis-py
+{%- if cookiecutter.windows == "n" %}
```
browniebroke, line 12: I think this condition should be also considering Docker, we can install `hiredis` if it's selected:

```suggestion
{%- if use_docker == "y" or cookiecutter.windows == "n" %}
```

We could also specify it as extra of redis:

```
redis[hiredis]
```

{{cookiecutter.project_slug}}/requirements/production.txt
```diff
@@ -12,6 +12,9 @@ Collectfast==2.1.0  # https://github.com/antonagestam/collectfast
 {%- if cookiecutter.use_sentry == "y" %}
 sentry-sdk==0.14.3  # https://github.com/getsentry/sentry-python
 {%- endif %}
+{%- if cookiecutter.windows == "y" %}
```
browniebroke, line 15: Same here:

```suggestion
{%- if use_docker == "n" and cookiecutter.windows == "y" %}
```

{{cookiecutter.project_slug}}/requirements/base.txt
Andrew-Chen-Wang: I don't use a windows machine anymore, but I honestly wouldn't risk it. I wouldn't know if the build would succeed.

{{cookiecutter.project_slug}}/requirements/production.txt
```diff
@@ -12,6 +12,9 @@ Collectfast==2.1.0  # https://github.com/antonagestam/collectfast
 {%- if cookiecutter.use_sentry == "y" %}
 sentry-sdk==0.14.3  # https://github.com/getsentry/sentry-python
 {%- endif %}
+{%- if cookiecutter.use_docker == "n" and cookiecutter.windows == "y" %}
+hiredis==1.0.1  # https://github.com/redis/hiredis-py
```
wadkar, line 16: I am a bit confused here. The line tells me that if people are _not_ using docker **and** they have said yes to windows, we force the use of `hiredis` in production? Is this intentional?

Andrew-Chen-Wang: Yes, it's just wrong to use a Windows machine to deploy a server. If you take a look at base.txt, it has a different condition for if you use Windows or Docker. If you do use Windows but not docker, you cannot build hiredis. But, since you wanted to use Redis, you'll also get hiredis for production, assuming, with hope, that you deploy with a unix.

However, if you use Docker on Windows, you should be able to build hiredis, and thus it can go into base.txt.

wadkar: Ah, thanks - that clarifies a lot. I see this is mentioned in the description as well but it didn't click until now.

Thanks a lot!

browniebroke:
Thanks 🎉

PR #2578, opened 2020-04-29
Improve the ROOT_DIR definition

luzfcb:
## Description

- Make sure to resolve symbolic links, uses an absolute path in `ROOT_DIR`, and raise an error if the path does not exist by using the `resolve` method. https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve 
- Use `parent.parent.parent` because "Explicit is better than simplicity" and because it causes less cognitive overload then `parents[2]`.

## Rationale

[A similar change](https://github.com/django/django/blob/ba4389a36b5fb1afce0cddb4e28233138b6612b7/django/conf/project_template/project_name/settings.py-tpl#L16) will be sent when django 3.1 is released.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -5,8 +5,8 @@ from pathlib import Path
 
 import environ
 
-ROOT_DIR = Path(__file__).parents[2]
-# {{ cookiecutter.project_slug }}/)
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
Should `wsgi.py` and `asgi.py` with the `app_path` name also be changed, too?

luzfcb:
> Should wsgi.py and asgi.py with the app_path name also be changed, too?
By consistency, I think so

luzfcb:
Added fixes for `wsgi.py` and `asgi.py`

Issue #2226, opened 2019-09-19
.gitignore ignores scripts conflicts with runscript


bmihelac:
## What happened?

`.gitignore` ignores every `scripts` directory that can be used by `runscript` command of included `django-extensions` package. 

https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.gitignore#L332

## What should've happened instead?

`scripts` directory should not be ignored IMHO.

luzfcb:
@bmihelac the `scripts` directory is ignored to avoid commit the `scripts` or `Scripts` directory of the virtualenv. These directories only are created on Windows machine


Looking briefly at the implementation in django-extensions, I saw that there are some configurations that can be made in django settings, which are currently not properly documented in django-extensions. 

Have you ever tried any of them like `RUNSCRIPT_SCRIPT_DIR` and the others to check if is possible to customize the `scripts` directory name?

https://github.com/django-extensions/django-extensions/blob/f7eacd2263a1069c61bbee316e9c32cf22fa121f/django_extensions/management/commands/runscript.py#L94

bmihelac:
Hi @luzfcb  and thanks for looking into this.

Unfortunately I am not familiar with virtualevn on windows and have no idea if this could be handled in other ways.

While workaround with  `RUNSCRIPT_SCRIPT_DIR` is probably possible, I still feel  that ignoring every `scripts` (and others directories too) directory in project is maybe too broad and users would not be aware what and why is ignored. This is why I opened issue, but feel free to close it if you do not agree.

browniebroke:
I did some investigation, and I think we can remove these lines safely:

https://github.com/pydanny/cookiecutter-django/blob/9075bdb39b98f88e9210e04cc4cae5f5920baa68/%7B%7Bcookiecutter.project_slug%7D%7D/.gitignore#L326-L333

When creating a virtual environment, such directories are actually **inside** the top directory, which we aim at excluding from these lines:

https://github.com/pydanny/cookiecutter-django/blob/9075bdb39b98f88e9210e04cc4cae5f5920baa68/%7B%7Bcookiecutter.project_slug%7D%7D/.gitignore#L70-L73

Example with `venv`:

```
.
`- venv
   |- Lib
   |- Scripts
   `- pyvenv.cfg
```

It sounds to me like we are being extra careful...

PR #2583, opened 2020-05-01
Fix OfflineGenerationError with Heroku + Django Compressor

### LINKED ISSUE #2581: OfflineGenerationError when deploying with compressor on Heroku
## What happened?

While trying to reproduce #2328, I got a different error with Heroku. Deployment works fine, but got a server error when opening the website:

```
OfflineGenerationError

You have offline compression enabled but key "d61d41e4f5705bf7675fe8d6b372b4b43ea27d8bbbb4af00a2012fbc8ebc98fa" 
is missing from offline manifest. 
You may need to run "python manage.py compress". Here is the original content:


    <!-- This file stores project-specific CSS -->
    
    <link href="/static/css/project.a9bf09099778.css" rel="stylesheet">
```

## Steps to reproduce

I think it's down to the following combination:

- use_heroku: y
- use_whitenoise: y
- use_compressor: y
- cloud_provider: None

Extract from the replay file:

```
{
    "use_docker": "n",
    "postgresql_version": "11.3",
    "js_task_runner": "None",
    "cloud_provider": "None",
    "mail_service": "Mailgun",
    "use_async": "n",
    "use_drf": "n",
    "custom_bootstrap_compilation": "n",
    "use_compressor": "y",
    "use_celery": "n",
    "use_mailhog": "n",
    "use_sentry": "y",
    "use_whitenoise": "y",
    "use_heroku": "y"
}
```

browniebroke:
## Description

Add `post_compile` build hook to run offline compression when using Heroku + Django Compressor. The hook is copying what we do in Docker currently, and should be removed as soon as either Heroku or compressor are deselected.

## Rationale

Fixes #2581

### DIFF
--- a/{{cookiecutter.project_slug}}/.gitignore
+++ b/{{cookiecutter.project_slug}}/.gitignore
@@ -325,7 +325,6 @@ tags
 
 ### VirtualEnv template
 # Virtualenv
-[Bb]in
 [Ii]nclude
 [Ll]ib
 [Ll]ib64
--- a/.gitignore
+++ b/.gitignore
@@ -211,7 +211,6 @@ tags
 ### VirtualEnv template
 # Virtualenv
 # http://iamzed.com/2009/05/07/a-primer-on-virtualenv/
-[Bb]in
 [Ii]nclude
 [Ll]ib
 [Ll]ib64
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -75,6 +75,11 @@ def remove_heroku_files():
             # don't remove the file if we are using travisci but not using heroku
             continue
         os.remove(file_name)
+    remove_heroku_build_hooks()
+
+
+def remove_heroku_build_hooks():
+    shutil.rmtree("bin")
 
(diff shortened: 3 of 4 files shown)


{{cookiecutter.project_slug}}/.gitignore
```diff
@@ -323,13 +323,7 @@ Session.vim
 tags
 {% if cookiecutter.use_docker == 'n' %}
 
-### VirtualEnv template
 # Virtualenv
-[Bb]in
-[Ii]nclude
-[Ll]ib
-[Ll]ib64
-[Ss]cripts
```
browniebroke, line 332: I had to remove the `[Bb]in` entry otherwise it was ignored, and I ended up removing all these, but now I stumbled upon this one #2226 

I'm not a Windows user so I'm not sure how these work. Need to look into this.

browniebroke: I've changed this to be more conservative, and only remove the line we need for this PR: `[Bb]in`.

PR #2635, opened 2020-06-21
Replace User by get_user_model

AsheKR:
## Description

`users.api.serializers.py` has changed the code that brings users.

## Rationale

consistency of the code bringing in user


browniebroke:
Thanks!

Issue #629, opened 2016-06-16
Ask for development and production environments?


jayfk:
Maybe we should split the current configuration into environments and ask during project generation what configurations to use in dev/prod.

This would make it easier for users to decide how the want to build and deploy their projects.

Currently, it's a bit of a mess. On the dev side we have:

- vanilla python
- vanilla python + node
- docker

On the prod side we have:

- vanilla python
- docker
- heroku
- a guide for docker deployment on AWS
- maybe EBS, see #610 

All with possible side effects on each other and to make things worse, it's not even transparent. I think we have done a good job by keeping all possible configurations independent from each other until now, but the more things we add, it becomes more and more complex.

Thoughts?

PR #2637, opened 2020-06-21
Non Docker approach to CI to take advantage of cache

### LINKED ISSUE #2588: Add cache for Docker-Based Projects in CI
## Description

CI takes a really long time to build the docker containers. Approximately 4 minutes on just building with so many requirements. Although CI like Travis in their docs say they don't support caching since each build is always provisioned on a new machine, we can take advantage of GitHub packages to push docker images there using GH actions.

An example is detailed here: https://testdriven.io/blog/faster-ci-builds-with-docker-cache/

The other option would be to rewrite a bit of config/settings/test.py and rewrite the CI configuration files to run __without__ Docker and instead just "install" and run tests like a normal project.

## Rationale

CI takes a really long time to build for Docker projects.

Faster docker builds during CI or, in general, faster CI runs. Plus, less whining.

Andrew-Chen-Wang:
Fixes #2588 

[//]: # (Before you proceed:)

## Description

Non Docker approach to CI to take advantage of cache. Based on my change in Travis over a two month period of trials in different PRs, it's a safe assumption that the cache decreases time of building to test. @browniebroke Do you mind taking a look at the Travis configuration? I feel like there should be more jinja2 templates than just for celery... I don't know. I think it might have to do with the issue regarding the Postgres versioning.

If we want to improve on that, you can visit psycopg/psycopg3 for easy configuration in Travis on multiple Postgres versions.

@rocode Do you mind looking at the gitlab CI? that'd be really helpful so that something stupid doesn't pop up again. I know the Travis configuration is fine, but I'm not too sure about the non-Docker generation. It should be the same though. So long as all your needed environment variables are in your CI's environment.

## Rationale

Faster builds in the CIs for faster results.

Faster builds. Last time I benchmarked... maybe a month of two ago, it was a 1.5 minute cut.


Andrew-Chen-Wang:
Interestingly enough, looking back on this PR, it doesn't work? Works fine on my repo...

Andrew-Chen-Wang:
After experimenting, I understand the problem. For one, it should be env without the global. Secondly, the celery command works with 4.4.2. Testing for reasoning rn

browniebroke:
I think it's a tradeoff. 

Yes, it's faster, but it potentially increases the complexity with setting up the CI system. This issue with the latest Celery is a perfect example of the tradeoff. Docker enables you to eliminate these discrepencies and more easily reproduce problem locally.

I've had an opposite experiement at work where our CI is fully Dockerized. We've updated to a newer OS without any issues.

Finally, it goes a bit against how the template options are setup. Currently, when one selects Docker, the generated project truely goes full Docker. In my experience, it's very true for application container (Django/Celery), but less true of databases. Depending on experience, personal preference and use case it's quite common to use managed databases (Postgres, Redis, ...) on production. That's a bigger discussion covered in this old issue: #629 (I think)

Andrew-Chen-Wang:
@browniebroke I looked into the celery issue. It's an actual celery issue that should've been resolved with a recent PR for version 4.4.4 but wasn't (was broke starting from 4.4.3)... So I'll be making a new PR to resolve that. Otherwise, there really isn't a "celery multi start" as it'll just be multiple `celery start` commands instead. Simple fix, just not as elegant, but fast enough.

Regarding upgrading and such in the future for the template itself, I think people will appreciate if the CI runs faster. IIRC when I used to run the Docker way, it took at least 3 minutes. On the test repository (on 4.4.2 since 4.4.3 apparently broke celery multi command), it dropped to 1 minute 19 seconds.

Taking, for example, the database configuration, it's very easy to reconfigure Travis to accommodate for what you're looking for. If you take a look at psycopg/psycopg3's Travis configuration, we basically test on all Postgres versions except v9. It's as simple as adding 3-4 lines. Sure, there may be some additions that devs make in the Docker orchestration like the node configuration, but these are test cases; **they don't really require most of the Docker stuff**.

I think if a dev or group of devs are deciding between saving hours of CI by figuring out how to add maybe a couple lines of code or being lazy and setting up an entire Docker stack, they'd much rather choose to take a little bit of time to decrease the CI process.

Andrew-Chen-Wang:
@browniebroke Friendly ping that it works now! Just gotta create the folders now (didn't need to in 4.4.2 but I guess we have to now). https://travis-ci.com/github/Andrew-Chen-Wang/test-cookiecutter-django-cache/jobs/360069088

Not sure why it doesn't work now. I wonder if /var/run/celery and /var/log/celery was ever created before 4.4.3 (when it broke).

Andrew-Chen-Wang:
ping @browniebroke Travis implementation with celery is located here: https://github.com/Donate-Anything/Donate-Anything/blob/master/.travis.yml

Again, the CI for the celery portion is based on Celery's Travis CI itself. Let me know if you're ok with this PR (once I fix the conflicts).

Andrew-Chen-Wang:
Closing so I can also implement GitHub actions. Now that Travis doesn't give you unlimited build credits, and that most people are probably on a private repository so GitHub actions also gives you limited build credits, and Gitlab never really had a free CI/CD with its runners, this should be looked at again.

Andrew-Chen-Wang:
Then again, the above comment is correct about needing to add CI complexity. I just don't think anyone's really going to add THAT much complexity in a project...

PR #2648, opened 2020-06-30
Add author name and description to HTML meta

Andrew-Chen-Wang:

[//]: # (Before you proceed:)

## Description

Base.html has author name and description meta tags, but the cookiecutter generation never adds it in.

## Rationale

META tags just help a website out.

Why waste the initial user input?

Edit: sorry for the constant PRs. I continuously use cookiecutter-django, so I just keep stumbling upon areas to touch up or add on to.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
@@ -5,8 +5,8 @@
     <meta http-equiv="x-ua-compatible" content="ie=edge">
     <title>{% block title %}{% endraw %}{{ cookiecutter.project_name }}{% raw %}{% endblock title %}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="description" content="{{ cookiecutter.description }}">
-    <meta name="author" content="{{ cookiecutter.author_name }}">
+    <meta name="description" content="{% endraw %}{{ cookiecutter.description }}{% raw %}">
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
```diff
@@ -5,8 +5,8 @@
     <meta http-equiv="x-ua-compatible" content="ie=edge">
     <title>{% block title %}{% endraw %}{{ cookiecutter.project_name }}{% raw %}{% endblock title %}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="description" content="">
-    <meta name="author" content="">
+    <meta name="description" content="{{ cookiecutter.description }}">
+    <meta name="author" content="{{ cookiecutter.author_name }}">
```
browniebroke, line 9: The build failed because it's a Django template, but as cookiecutter uses Jinja under the hood, the project is itself a Jinja template. This file is therefore mixing Jinja + Django and Jinja processing is generally disabled with `raw` tag.

Here we want to inject data coming from cookiecutter, in Jinja processing should be enabled temporarily like in line 6:

```suggestion
    <meta name="description" content="{% endraw %}{{ cookiecutter.description }}{% raw %}">
    <meta name="author" content="{% endraw %}{{ cookiecutter.author_name }}{% raw %}">
```

Andrew-Chen-Wang:
I think what you said makes sense since the description is put in the README. I don't think a survey can really happen in this repo, but my first instinct when I see description during project initialization is the description of my product.

browniebroke:
Thanks again!

PR #2663, opened 2020-07-06
Add :z/:Z to mounted volumes in {local,production}.yml

westurner:

[//]: # (Before you proceed:)

## Description

Add `:z` or `:Z` suffixes to volumes in the `local.yml` and `production.yml` docker-compose configuration files.

## Rationale

Without these in local.yml, I was getting the following errors:

```bash
$ docker-compose -f local.yml up
[...]
django          | python: can't open file 'manage.py': [Errno 13] Permission denied
flower          | Unable to load celery application.
flower          | The module config was not found.
node            | Error: EACCES: permission denied, scandir '/app'
```

I've only cursorily tested that the modifications to local.yml do seem to work; I see the index page and the django-debug-toolbar sidebar when I navigate to http://localhost:8000/

- [x] (Have someone else?) test these modifications to `production.yml`
- [ ] Get feedback on whether relabeling is the only way to make this work on a platform with SELinux enabled (e.g. Fedora, CentOS, RHEL, SUSE)
- [x] Verify that `:z` and `:Z` are silently ignored on non-selinux platforms
- [x] Reference the docs: 
  - https://docs.docker.com/compose/compose-file/#volumes
  - https://docs.docker.com/storage/volumes/
  - https://danwalsh.livejournal.com/76016.html
    > :z will relabel with a shared label so other containers ran read and write the volume.
    > :Z will relabel with a private label so that only this specific container can read and write the volume.
- [x] Should `production_postgres_data_backups:/backups:z` be `:z` or `:Z`?

- Users developing in docker containers on systems with SELinux in enforcing mode (Fedora, CentOS, RHEL, SUSE)
- Users deploying to production with docker on systems with SELinux in enforcing mode (Fedora, CentOS, RHEL, SUSE)


{{cookiecutter.project_slug}}/production.yml
```diff
@@ -38,7 +38,7 @@ services:
     depends_on:
       - django
     volumes:
-      - production_traefik:/etc/traefik/acme
+      - production_traefik:/etc/traefik/acme:Z
```
browniebroke, line 41: Initially, I wasn't able to obtain a certificate from Let's Encrypt, it was showing up as self-signed cert from Traefik. 

I've changed this into `z` and it worked so I thought it was wrong... However, I later tried again with `Z` and it worked. I think my problem might be unrelated...

{{cookiecutter.project_slug}}/production.yml
```diff
@@ -25,8 +25,8 @@ services:
       dockerfile: ./compose/production/postgres/Dockerfile
     image: {{ cookiecutter.project_slug }}_production_postgres
     volumes:
-      - production_postgres_data:/var/lib/postgresql/data
-      - production_postgres_data_backups:/backups
+      - production_postgres_data:/var/lib/postgresql/data:Z
+      - production_postgres_data_backups:/backups:z
```
browniebroke, line 29: To answer one your point in the description, I think it's fine with `z`, since it will probably be accessed by the `awscli` service.

browniebroke (commented):
Tried on macOS, and it seems to work. Would be nice if someone on Windows could confirm that it doesn't break anything.

{{cookiecutter.project_slug}}/production.yml
westurner: If there's another container that needs to access the volume, it should be `:z` so that the volume has a shared label.

If, as is the case with traefik FWIU, only the one container needs to access the volume, it should be `:Z` so that the volume has a private label (which is unique to the container). For a multi-node setup, in order to share the cert with every load balancer container, the acme volume would need to be `:z` so that the volume has a shared labe.

westurner: I changed the label to `:z` so that you could run multiple traefik containers that share the acme config.
While this is more flexible, it is less secure.

browniebroke:
Ok let's try this. Let's keep an eye on issues reported from Windows users in the coming days.

Thank you for your contribution! 🎉

westurner:
Thanks!

professorhaseeb:
> Ok let's try this. Let's keep an eye on issues reported from Windows users in the coming days.
> 
> Thank you for your contribution! 🎉

Works properly on windows 👍

PR #2674, opened 2020-07-10
Add entrypoint for docs to not error on build

### LINKED ISSUE #2712: Docs fail to build due to missing CELERY_BROKER_URL environment variable
## What happened?

My project is configured to use Celery and Docker. When I run `docker-compose -f local.yml up` (after building) I get the following error message from the docs container:

```
docs            | Running Sphinx v3.1.2
docs            | /usr/local/lib/python3.8/site-packages/environ/environ.py:437: UserWarning: Engine not recognized from url: {'NAME': '', 'USER': '', 'PASSWORD': '', 'HOST': '', 'PORT': '', 'ENGINE': ''}
docs            |   warnings.warn("Engine not recognized from url: {0}".format(config))
docs            | 
docs            | Configuration error:
docs            | There is a programmable error in your configuration file:
docs            | 
docs            | Traceback (most recent call last):
docs            |   File "/usr/local/lib/python3.8/site-packages/environ/environ.py", line 273, in get_value
docs            |     value = self.ENVIRON[var]
docs            |   File "/usr/local/lib/python3.8/os.py", line 675, in __getitem__
docs            |     raise KeyError(key) from None
docs            | KeyError: 'CELERY_BROKER_URL'
docs            | 
docs            | During handling of the above exception, another exception occurred:
docs            | 
docs            | Traceback (most recent call last):
docs            |   File "/usr/local/lib/python3.8/site-packages/sphinx/config.py", line 319, in eval_config_file
docs            |     execfile_(filename, namespace)
docs            |   File "/usr/local/lib/python3.8/site-packages/sphinx/util/pycompat.py", line 88, in execfile_
docs            |     exec(code, _globals)
docs            |   File "/docs/conf.py", line 22, in <module>
docs            |     django.setup()
docs            |   File "/usr/local/lib/python3.8/site-packages/django/__init__.py", line 19, in setup
docs            |     configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
docs            |   File "/usr/local/lib/python3.8/site-packages/django/conf/__init__.py", line 76, in __getattr__
docs            |     self._setup(name)
docs            |   File "/usr/local/lib/python3.8/site-packages/django/conf/__init__.py", line 63, in _setup
docs            |     self._wrapped = Settings(settings_module)
docs            |   File "/usr/local/lib/python3.8/site-packages/django/conf/__init__.py", line 142, in __init__
docs            |     mod = importlib.import_module(self.SETTINGS_MODULE)
docs            |   File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
docs            |     return _bootstrap._gcd_import(name[level:], package, level)
docs            |   File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
docs            |   File "<frozen importlib._bootstrap>", line 991, in _find_and_load
docs            |   File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
docs            |   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
docs            |   File "<frozen importlib._bootstrap_external>", line 783, in exec_module
docs            |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
docs            |   File "/app/config/settings/local.py", line 1, in <module>
docs            |     from .base import *  # noqa
docs            |   File "/app/config/settings/base.py", line 259, in <module>
docs            |     CELERY_BROKER_URL = env("CELERY_BROKER_URL")
docs            |   File "/usr/local/lib/python3.8/site-packages/environ/environ.py", line 123, in __call__
docs            |     return self.get_value(var, cast=cast, default=default, parse_default=parse_default)
docs            |   File "/usr/local/lib/python3.8/site-packages/environ/environ.py", line 277, in get_value
docs            |     raise ImproperlyConfigured(error_msg)
docs            | django.core.exceptions.ImproperlyConfigured: Set the CELERY_BROKER_URL environment variable
```

Tracing back through the code I can see that indeed the CELERY_BROKER_URL environment variable is not set in the docs container, because it usually set in the entrypoint script that the django container uses. I have not used sphinx before so I don't understand why it needs this environment variable to proceed. 

Manually setting the environment variable in the docs Dockerfile solves the issue, but I don't know if this proliferation of environment variable setting is best for the project so I won't submit that as a fix at this point.

I would appreciate any input on a decent way of solving this problem, I am happy to write a PR following that guidance or leave it to someone else if they want to write one.

Thanks,
Adam

Andrew-Chen-Wang:

[//]: # (Before you proceed:)

## Description

Docker building the docs error unless you add entry point to configure the database.

## Rationale

You see nothing unless you can build the docs... so you need to make sure Django is actually setup.

To see the docs.

~~Note: I wonder if we can do what we do for celery which is using anchors. The command: sphinx-autobuild -b html -H 0.0.0.0 -p 7000 --watch /app -c . ./_source ./_build/html~~ Still would need production.txt...

Fixes #2712

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/local/docs/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/local/docs/Dockerfile
@@ -24,6 +24,12 @@ COPY ./requirements /requirements
 # All imports needed for autodoc.
 RUN pip install -r /requirements/local.txt -r /requirements/production.txt
 
+COPY ./compose/production/django/entrypoint /entrypoint
+RUN sed -i 's/\r$//g' /entrypoint
+RUN chmod +x /entrypoint
+
 WORKDIR /docs
+
+ENTRYPOINT ["/entrypoint"]
 
 CMD make livehtml
--- a/{{cookiecutter.project_slug}}/compose/local/docs/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/local/docs/Dockerfile
@@ -28,8 +28,10 @@ COPY ./compose/production/django/entrypoint /entrypoint
 RUN sed -i 's/\r$//g' /entrypoint
 RUN chmod +x /entrypoint
 
+COPY ./compose/local/docs/start /start-docs
+RUN sed -i 's/\r$//g' /start-docs
+RUN chmod +x /start-docs
+
 WORKDIR /docs
 
 ENTRYPOINT ["/entrypoint"]
-
-CMD make livehtml
--- a/{{cookiecutter.project_slug}}/compose/local/docs/start
+++ b/{{cookiecutter.project_slug}}/compose/local/docs/start
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+make livehtml
--- a/{{cookiecutter.project_slug}}/local.yml
+++ b/{{cookiecutter.project_slug}}/local.yml
@@ -43,6 +43,8 @@ services:
     build:
       context: .
       dockerfile: ./compose/local/docs/Dockerfile
+    depends_on:
+      - postgres
     env_file:
       - ./.envs/.local/.django
       - ./.envs/.local/.postgres
--- a/{{cookiecutter.project_slug}}/local.yml
+++ b/{{cookiecutter.project_slug}}/local.yml
@@ -45,6 +45,7 @@ services:
       dockerfile: ./compose/local/docs/Dockerfile
     env_file:
       - ./.envs/.local/.django
+      - ./.envs/.local/.postgres
     volumes:
       - ./docs:/docs
       - ./config:/app/config
--- a/{{cookiecutter.project_slug}}/local.yml
+++ b/{{cookiecutter.project_slug}}/local.yml
@@ -52,6 +52,7 @@ services:
       - ./{{ cookiecutter.project_slug }}:/app/{{ cookiecutter.project_slug }}
     ports:
       - "7000:7000"
+    command: /start-docs
     
   {%- if cookiecutter.use_mailhog == 'y' %}
 


Andrew-Chen-Wang:
Sure for consistency, it'd make sense. Considering the celery commands are also just one-liners, having the a start file would make more sense

browniebroke:
Just tried it and I'm getting:

```
> docker-compose -f local.yml up docs
...
docs        | Waiting for PostgreSQL to become available...
docs        | Waiting for PostgreSQL to become available...
docs        | Waiting for PostgreSQL to become available...
docs        | Waiting for PostgreSQL to become available...
docs        | Waiting for PostgreSQL to become available...
docs        | Waiting for PostgreSQL to become available...
```

It works if I add this to `local.yml`:

```diff
  docs:
    image: my_awesome_project_local_docs
    container_name: docs
+   depends_on:
+     - postgres
    build:
      context: .
```

I think I need to get back to your original comment though, I don't understand why we need the DB to build the docs.

Andrew-Chen-Wang:
I believe in the docs, we're running `django.setup()` which means we're trying to connect to the database.

browniebroke:
So we need `depends_on` to run the command to only start the docs container: `docker-compose -f local.yml up docs`. If you started your project before generating the docs, the DB might already be up, in which case no issue. 

If the `docs` container tries to connect to the DB, we need this `depends_on` to declare this dependency to docker compose. I think we can add it to fix this bug, and maybe see if we can cut this dependency separately.

Andrew-Chen-Wang:
Makes sense. Never thought about it like that. I've committed the necessary change.

browniebroke:
I've just tried by generating a project on the master branch, and I'm able to get the docs working. There is no error preventing it from running, but there is this warning:

```
docs        | /usr/local/lib/python3.8/site-packages/environ/environ.py:437: UserWarning: Engine not recognized from url: {'NAME': '', 'USER': '', 'PASSWORD': '', 'HOST': '', 'PORT': '', 'ENGINE': ''}
docs        |   warnings.warn("Engine not recognized from url: {0}".format(config))
```

Is that what you're trying to fix with this change?

To remove the dependency on the `postgres` container for the docs, maybe it would be simpler to set `DATABASE_URL` to a SQLite file for the `docs` container? I think we could do that by changing the `local.yml` as follows for instance:

```diff
  docs:
    image: my_awesome_project_local_docs
    container_name: docs
    build:
      context: .
      dockerfile: ./compose/local/docs/Dockerfile
    env_file:
      - ./.envs/.local/.django
    volumes:
      - ./docs:/docs
      - ./config:/app/config
      - ./my_awesome_project:/app/my_awesome_project
    ports:
      - "7000:7000"
+   environment:
+     DATABASE_URL: "sqlite:///database.db"
```

In which case we don't need to add the `entrypoint` or the postgres env file the `Dockerfile`.

Let me know if you were talking about another error, either I missed it or I'm able to reproduce.

Andrew-Chen-Wang:
Sure that'd work! The problem was with the DATABASE_URL and other variables set in the entrypoint plus Django was trying to connect to a database. Would people have to continuously add their environment variables to the docs service?

Either way works, but I feel like setting the entrypoint via the Dockerfile lets devs to forget about setting env vars from entrypoint in local.yml, especially for multiple contributors. I'm not... entirely sure what the best option is :P although setting the database to sqlite is clever and would work. Would it create a new file though...? Hopefully not since we're not creating tables in the docs start file.

browniebroke:
> Would people have to continuously add their environment variables to the docs service?

That's a good point, not sure about that. My assumption was that the `entrypoint` is not expected to change much, but it depends what you'd do after generating the template. If we set good defaults, we shouldn't need to override a lot, but while writing this I noticed that we also have `CELERY_BROKER_URL` in there...

> Would it create a new file though

It probably would, but only inside the container, which would be destroyed when the command has finished. I don't know if that's the case with the snippet I suggested, but at least that's the idea.

The main potential problem I foresee with the current approach is that in some environments, it might not be easy/cheap to get hold of a database just to generate the docs, but it might be helpful to use the `Dockerfile`. With the use of `entrypoint`, one needs to set 4 environment variables, and is tied to PostgreSQL. 

On the other hand, it makes the `docs` container a more alike the `django` one.

Andrew-Chen-Wang:
I think your approach is more viable with the explicit setting of env vars in the compose file, especially when you say:

> entrypoint is not expected to change much

I think I've only edited entrypoint a couple of times for multi-db, so point taken. I agree that the dependency on other services would drag some projects. I'll test if SQLite is created and continues existing post `compose down` when I get the chance. In that case, I'll close this PR and start a new one with the new suggestions. (remove entrypoint; add env vars from entrypoint to docs service environment; change DATABASE_URL to a sqlite url; remove dependencies that docs service depends on).

Perhaps a couple of comments in the compose file or entrypoint file saying something like `Docs service requires the file's env vars to be in local.yml`? Or perhaps somewhere in the cookie cutter django docs, specify this necessity?

PR #2694, opened 2020-07-24
Be explicit in context processor

### LINKED ISSUE #2614: selective settings in context
## Description

Change the settings_context processor to only pass a required subset of the settings variables instead of all. We could have a special settings variable called CONTEXT_SETTINGS and set as an array of settings variable names to pass to the templates, e.g. ['GOOGLE_ID', 'MY_OTHER_SETTINGS', 'PAGINATION_SIZE', 'SITE_TITLE']. settings_context() would read that and only pass the relevant settings variable to the context.

## Rationale

Am I the only one to feel a bit uneasy about the idea of passing the whole settings to the context? It is very convenient but is that a best practice? 

https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/utils/context_processors.py#L5

For various reasons:

1. it is unnecessary
2. it may clutter the context or debugging metadata with a lot of unneeded text
3. it is harder to track dependencies between templates and settings (i.e. we bypass the controller for a lot of stuff)
4. it spreads sensitive information (software stack, database connection, passwords, secret keys, ...) to the context and may accidentally end up on screen, in logs, ...

### LINKED ISSUE #2670: Implicit global settings  in utils/context_processors.py is a security risk
## What happened?

https://github.com/feldroy/two-scoops-of-django-3.x/issues/80

We shouldn't dump all settings into the context processor. All it takes is one mistake and a critical password or key is now public.

Solution in https://github.com/feldroy/django-crash-starter/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/utils/context_processors.py

Andrew-Chen-Wang:
Closes #2670 Closes #2614 

## Description

Explicit dict for context processor.

## Rationale

Although it poses a security risk if a developer accidentally exposes something like a key, I think a general risk would be if someone were using some CI and env vars were exposed like that.

General practicing though, it makes developers more keen and hopefully more cautious on what they're putting in a Django template.

Better Django dev habits.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
@@ -2,5 +2,7 @@ from django.conf import settings
 
 
 def settings_context(_request):
-    # Be explicit
+    """Settings available by default to the templates context."""
+    # Note: we intentionally do NOT expose the entire settings
+    # to prevent accidental leaking of sensitive information
     return {"DEBUG": settings.DEBUG}
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
@@ -2,4 +2,5 @@ from django.conf import settings
 
 
 def settings_context(_request):
-    return {"settings": settings}
+    # Be explicit
+    return {"DEBUG": settings.DEBUG}


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
```diff
@@ -2,4 +2,5 @@
 
 
 def settings_context(_request):
-    return {"settings": settings}
+    # Be explicit
```
browniebroke, line 5: I'm not sure it will make sense outside of the context of this pull request, maybe we could make it a full sentence?

I also suggested a docstring as this function has none currently.

```suggestion
    """Settings available by default to the templates context."""
    # Note: we intentionally do NOT expose the entire settings
    # to prevent accidental leaking of sensitive information
```

Andrew-Chen-Wang: Ah good point. Committing. Thanks!

browniebroke:
Thanks for the fix!

PR #2745, opened 2020-08-13
Add STATICFILES_STORAGE with compressor support

### LINKED ISSUE #2739: Compressor CACHE not publicly available on S3
## What happened?

Using Django compressor with AWS on S3 will cache outputs and send them to S3. Unfortunately, if you have CORS setup, the CACHE directory is still sending back 403 errors. If you go directly to the static directory of your S3, there's no problem.

The public new files should be publicly available. I believe this COULD help but I haven't tested it out yet: https://django-compressor.readthedocs.io/en/latest/remote-storages/

## Steps to reproduce

- Start a new cookiecutter-django project. 
- Enable django compressor
- If you'd like, enable compress in a different HTML file with a different static file.
- Go to your production website.

I could only fix it by making each file public.

Andrew-Chen-Wang:
Fixes #2739

## Description

`COMPRESSOR_STORAGE` should be set to `STATICFILES_STORAGE` if not using whitenoise and using something like AWS S3.

## Rationale

Compressor cache is stored in the AWS bucket in a separate file structure. For example, without this PR, the files would not have an ACL of public-read:

```
- CACHE
- static
- media
```

With this PR:

```
- static
    - CACHE
- media
```

The static files storage is set to S3, but the utils (in cookiecutter.project_slug.utils.storages) are set up only for the original collectstatic command, not for compressor support. Thus, compressor would store stuff inside its own directory that did not have public-read.

Allow compressor directory of CACHE to be publicly viewable in storage buckets.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -242,7 +242,7 @@ COMPRESS_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
 COMPRESS_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
 {%- elif cookiecutter.cloud_provider == 'AWS' or cookiecutter.cloud_provider == 'GCP' %}
 COMPRESS_STORAGE = STATICFILES_STORAGE
-COMPRESS_ROOT = STATIC_ROOT
+COMPRESS_ROOT = STATIC_ROOT  # noqa F405 
 {%- elif cookiecutter.cloud_provider == 'None' %}
 COMPRESS_STORAGE = "compressor.storage.GzipCompressorFileStorage"
 {%- endif %}
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -235,16 +235,11 @@ ANYMAIL = {}
 # ------------------------------------------------------------------------------
 # https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_ENABLED
 COMPRESS_ENABLED = env.bool("COMPRESS_ENABLED", default=True)
-# https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_STORAGE
-{%- if cookiecutter.cloud_provider == 'AWS' and cookiecutter.use_whitenoise == 'y' %}
-COMPRESS_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
-{%- elif cookiecutter.cloud_provider == 'GCP' and cookiecutter.use_whitenoise == 'y' %}
-COMPRESS_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
-{%- elif cookiecutter.cloud_provider == 'AWS' or cookiecutter.cloud_provider == 'GCP' %}
-COMPRESS_STORAGE = STATICFILES_STORAGE
-COMPRESS_ROOT = STATIC_ROOT  # noqa F405 
-{%- elif cookiecutter.cloud_provider == 'None' %}
+{%- if cookiecutter.cloud_provider == 'None' %}
 COMPRESS_STORAGE = "compressor.storage.GzipCompressorFileStorage"
+{%- elif cookiecutter.cloud_provider in ('AWS', 'GCP') and cookiecutter.use_whitenoise == 'n' %}
+# https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_STORAGE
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
You can view this implementation at https://github.com/Donate-Anything/Donate-Anything and view the current website at https://donate-anything.org/. Posting this comment to double check that the files are actually publicly available and the listed project structure from the initial comment is correct.

(I keep getting some favicon.ico error when accessing the admin. I will file a new issue for that, too, although it doesn't happen... often to say the least. I'm not entirely sure why this happens since Sentry doesn't pick up on it but my email logging does).

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -236,10 +236,13 @@
 # https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_ENABLED
 COMPRESS_ENABLED = env.bool("COMPRESS_ENABLED", default=True)
 # https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_STORAGE
-{%- if cookiecutter.cloud_provider == 'AWS' %}
+{%- if cookiecutter.cloud_provider == 'AWS' and cookiecutter.use_whitenoise == 'y' %}
 COMPRESS_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
```
browniebroke, line 240: I've tested quickly this combination and it's still broken. I know it's not the one you're trying to fix here, but I'm wondering if this value of `COMPRESS_STORAGE` makes sense.

With Whitenoise, the static files are hosted on the application server, why the compress files should be stored on AWS?

I did a quick test and it looks like we don't need to customise the storage when using whitenoise.

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -236,10 +236,13 @@
...
-{%- elif cookiecutter.cloud_provider == 'GCP' %}
+{%- elif cookiecutter.cloud_provider == 'GCP' and cookiecutter.use_whitenoise == 'y' %}
 COMPRESS_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
+{%- elif cookiecutter.cloud_provider == 'AWS' or cookiecutter.cloud_provider == 'GCP' %}
+COMPRESS_STORAGE = STATICFILES_STORAGE
+COMPRESS_ROOT = STATIC_ROOT  # noqa F405
```
browniebroke, line 245: The `COMPRESS_ROOT` setting defaults to `STATIC_ROOT`, we should not need this line:

https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_ROOT

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -236,10 +236,13 @@
...
 COMPRESS_ENABLED = env.bool("COMPRESS_ENABLED", default=True)
 # https://django-compressor.readthedocs.io/en/latest/settings/#django.conf.settings.COMPRESS_STORAGE
-{%- if cookiecutter.cloud_provider == 'AWS' %}
+{%- if cookiecutter.cloud_provider == 'AWS' and cookiecutter.use_whitenoise == 'y' %}
 COMPRESS_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
-{%- elif cookiecutter.cloud_provider == 'GCP' %}
+{%- elif cookiecutter.cloud_provider == 'GCP' and cookiecutter.use_whitenoise == 'y' %}
 COMPRESS_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
+{%- elif cookiecutter.cloud_provider == 'AWS' or cookiecutter.cloud_provider == 'GCP' %}
+COMPRESS_STORAGE = STATICFILES_STORAGE
```
browniebroke, line 244: This is the real fix. I think we need to set this value when: 

- NOT using Whitenoise 
AND 
- Cloud provider is either AWS or GCP.

browniebroke: We don't need to cover Whitenoise + no cloud provider as this tis not supported:

https://github.com/pydanny/cookiecutter-django/blob/ebd43bbd3c29ff29cdf2617e661b857b89eb520c/hooks/pre_gen_project.py#L63-L70

{{cookiecutter.project_slug}}/config/settings/production.py
Andrew-Chen-Wang: Hm I quickly glossed over white noise (never using a CDN before), but I assumed the files were hosted on the instance/server itself rather than the bucket. Taking a look again, are the CDN's files still in these buckets? Or are they just kinda... cached?

PR #2754, opened 2020-08-15
Add ReadTheDocs Support

### LINKED ISSUE #2733: RTD Support
## Description

ReadTheDocs support

## Rationale

Hosting an open-source website with documentation 

Also somewhat of a bug report because you need a .readthedocs.yml (I think) file in order to pip install the local.txt packages.

Developers for an open-source (or closed-source) website can easily read up on how the website functions.

Andrew-Chen-Wang:
Fixes #2733 
## Description

Add RTD support for websites.

* Test repository: https://github.com/Andrew-Chen-Wang/test-cookiecutter-django-rtd
* Sample URL (old version looking): https://test-cookiecutter-django-rtd.readthedocs.io/en/v1/_source/index.html
* Sample URL (modern version looking): https://test-cookiecutter-django-rtd.readthedocs.io/en/latest/_source/index.html

## Rationale

Some people with websites may want to add documentation in their code, whether private for enterprise or public for [open source websites](https://donate-anything.org).

* Issue: you must go to /_source for this to work due to how the docs/* configuration is setup up.

To enable that modern looking RTD, check out https://github.com/readthedocs/sphinx_rtd_theme. Go to conf.py and add that to the extensions (make sure it's in the requirements file).

You can have something like this (below the extensions list):
```
if os.getenv("READTHEDOCS", default=False) == "True":
    import sphinx_rtd_theme
    extensions += ["sphinx_rtd_theme"]
    html_theme = "sphinx_rtd_theme"
else:
    html_theme = "alabaster"
```
Imo, more people would like that first part rather than the alabaster format.

### DIFF
--- a/{{cookiecutter.project_slug}}/docs/conf.py
+++ b/{{cookiecutter.project_slug}}/docs/conf.py
@@ -14,11 +14,20 @@ import os
 import sys
 import django
 
+if os.getenv("READTHEDOCS", default=False) == "True":
+    sys.path.insert(0, os.path.abspath(".."))
+    os.environ["DJANGO_READ_DOT_ENV_FILE"] = "True"
+    {% if cookiecutter.use_celery == 'y' -%}
+    os.environ["CELERY_BROKER_URL"] = os.getenv("REDIS_URL", "redis://redis:6379")
+    {%- endif %}
+    os.environ["USE_DOCKER"] = "no"
+    master_doc = "_source/index"
+else:
 {% if cookiecutter.use_docker == 'y' %}
-sys.path.insert(0, os.path.abspath("/app"))
-os.environ.setdefault("DATABASE_URL", "")
+    sys.path.insert(0, os.path.abspath("/app"))
+    os.environ.setdefault("DATABASE_URL", "")
 {% else %}
-sys.path.insert(0, os.path.abspath(".."))
+    sys.path.insert(0, os.path.abspath(".."))
 {%- endif %}
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")
 django.setup()
--- a/{{cookiecutter.project_slug}}/docs/conf.py
+++ b/{{cookiecutter.project_slug}}/docs/conf.py
@@ -17,16 +17,16 @@ import django
 if os.getenv("READTHEDOCS", default=False) == "True":
     sys.path.insert(0, os.path.abspath(".."))
     os.environ["DJANGO_READ_DOT_ENV_FILE"] = "True"
-    {% if cookiecutter.use_celery == 'y' -%}
+    {%- if cookiecutter.use_celery == 'y' %}
     os.environ["CELERY_BROKER_URL"] = os.getenv("REDIS_URL", "redis://redis:6379")
     {%- endif %}
     os.environ["USE_DOCKER"] = "no"
     master_doc = "_source/index"
 else:
-{% if cookiecutter.use_docker == 'y' %}
+{%- if cookiecutter.use_docker == 'y' %}
     sys.path.insert(0, os.path.abspath("/app"))
     os.environ.setdefault("DATABASE_URL", "")
-{% else %}
+{%- else %}
     sys.path.insert(0, os.path.abspath(".."))
 {%- endif %}
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")
--- a/{{cookiecutter.project_slug}}/.readthedocs.yml
+++ b/{{cookiecutter.project_slug}}/.readthedocs.yml
@@ -1,0 +1,13 @@
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+formats:
+  - pdf
+
+python:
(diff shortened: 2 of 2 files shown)


Andrew-Chen-Wang:
I'll fix the docs service as well to make sure it can run compose up. In this scenario, it follows @browniebroke 's comment on whether or not we should add the env vars to the .yml file or depend on Postgres.

In other words, for that PR, I'll just have SQLite DB be the DATABASE_URL.

Andrew-Chen-Wang:
It's possible to add the production.txt requirements as a second requirements file: https://docs.readthedocs.io/en/stable/config-file/v2.html#requirements-file I'm not sure if we need to add it. Let me know if I should.

Andrew-Chen-Wang:
One issue would be the modindex link. I don't know what it does, but it returns a 404

Andrew-Chen-Wang:
Once this is merged, I'll add the needed compose environment variables from #2674 once this is merged to solve two errors separately.

{{cookiecutter.project_slug}}/docs/conf.py
```diff
@@ -14,11 +14,20 @@
...
-{% else %}
-sys.path.insert(0, os.path.abspath(".."))
+if os.getenv("READTHEDOCS", default=False) == "True":
+    sys.path.insert(0, os.path.abspath(".."))
+    os.environ["DJANGO_READ_DOT_ENV_FILE"] = "True"
+    {%- if cookiecutter.use_celery == 'y' %}
+    os.environ["CELERY_BROKER_URL"] = os.getenv("REDIS_URL", "redis://redis:6379")
+    {%- endif %}
+    os.environ["USE_DOCKER"] = "no"
+    master_doc = "_source/index"
```
browniebroke, line 24: So if I follow, it's partly because of this line that we have this problem:

> you must go to /_source for this to work due to how the docs/* configuration is setup up.

But if we remove it the build fails because the source isn't in the right place relative to `conf.py`.

To fix it, we should move the `conf.py` in `_source`. We'll need to update a few other things to make it work, but I managed to make the RTD part work:

https://github.com/browniebroke/django-rtd
https://readthedocs.org/projects/django-rtd-tests/builds/

{{cookiecutter.project_slug}}/docs/conf.py
```diff
@@ -14,11 +14,20 @@
...
+else:
+{%- if cookiecutter.use_docker == 'y' %}
+    sys.path.insert(0, os.path.abspath("/app"))
+    os.environ.setdefault("DATABASE_URL", "")
```
browniebroke, line 28: This environbment variable needs to be set in RTD at the moment... I'm think, would it make sense to have the steps required documented?

These are the steps I had to take manually, which I don't think can be automated at the moment:

1. Go to readthedocs.org and login/create and account
2. Add the Github repository
3. Setup the environemnt variable `DATABASE_URL=sqlite:///database.db`
4. Trigger a build

We could then link to a few handy advanced settings from RTD like [pull request previews](https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html#autobuild-documentation-for-pull-requests) or versioned documentation.

browniebroke (changes requested):
Thanks for your patience in getting this reviewed, it's nearly there, but mostly uncovering existing issues...

> To enable that modern looking RTD

I think we can maybe enable it in a follow up pull request, not critical here.

> One issue would be the modindex link

That's pre-existing I think, due to this line:

https://github.com/pydanny/cookiecutter-django/blob/3152bdaeb37c863a4e7942d8a243af1d7453c9ce/%7B%7Bcookiecutter.project_slug%7D%7D/docs/_source/index.rst#L23

{{cookiecutter.project_slug}}/docs/conf.py
Andrew-Chen-Wang: Yes, we should definitely document the manual steps in the docs. I can add that in. Although I'm not sure if the `DATABASE_URL=sqlite:///database.db` is necessary... 

I think referring to our previous conversation to how docs are built, we should automatically create that database and even perhaps rename it to `.readthedocs.db` and leave the fact that we manually changed the DB in the docs. Which reminds me, did it create that sqlite db?

{{cookiecutter.project_slug}}/docs/conf.py
Andrew-Chen-Wang: That's awesome! I'll take a look and try to add your conf.py to this.

Andrew-Chen-Wang:
Thanks for the review! Kind of busy this week, so I'll follow up with a commit next week.

>> To enable that modern looking RTD

> I think we can maybe enable it in a follow up pull request, not critical here.

Sure, I can make a PR in the future for this. Perhaps as two options in `cookiecutter.json`.

{{cookiecutter.project_slug}}/docs/conf.py
browniebroke: Ideally, that step shouldn't be there, but for the moment we have this other issue that (I think) you were planning to fix in the other PR... As it's out of scope here, I was suggesting to document it until it's fixed.

PR #2775, opened 2020-08-28
Added visibility_timeout and retry_on_timeout to CELERY_BROKER_TRANSP…

foarsitter:
Using the ETA function of Celery has some caveats, which ~~the most of us learn~~ I learned the hard way. 

Adding `visibility_timeout` with the Celery default of `1 hour` may remind me in the future to customize this setting to the needs of the project before deploying to production.


{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -293,6 +293,12 @@
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#beat-scheduler
 CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"
 
+CELERY_BROKER_TRANSPORT_OPTIONS = {
+    # https://docs.celeryproject.org/en/stable/getting-started/brokers/redis.html#id1
+    # increase the visibility timeout to match the time of the longest ETA you’re planning to use.
+    "visibility_timeout": 60 * 60 * 1,  # default is 1 hour
+    "retry_on_timeout": True,
```
browniebroke, line 300: The visibility timeout makes sense, but I couldn't find the `retry_on_timeout` option in the Celery docs, is `True` the default as well? Should we set a max number of retries?

https://docs.celeryproject.org/en/stable/userguide/configuration.html#std:setting-broker_transport_options

foarsitter: I interpreted the sentence `See your transport user manual for supported options (if any).` in the way I cloud pass options to the `Redis` instance (https://redis-py.readthedocs.io/en/stable/#redis.Redis) but I have not tested It.

PR #2776, opened 2020-08-28
Reference production.txt in local.txt for installing production requi…

foarsitter:
## Description

My proposal is to install production.txt requirements in the dev/local environment. 

## Rationale

A lot of stuff for development is not necessary for production. The other way around is it no problem to have a few extra dependencies around in development. 

- When editing production.py, I face missing reference and cannot use autocomplete features of my IDE. 
- Developing a storage backend requires django-storages.

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/local.txt
+++ b/{{cookiecutter.project_slug}}/requirements/local.txt
@@ -1,12 +1,8 @@
 -r base.txt
+-r production.txt
 
 Werkzeug==1.0.1 # https://github.com/pallets/werkzeug
 ipdb==0.13.3  # https://github.com/gotcha/ipdb
-{%- if cookiecutter.use_docker == 'y' %}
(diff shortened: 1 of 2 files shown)


browniebroke:
I think this was raised some time ago when we considered pipenv/poetry, and an objection was that we have 3 sets of requirements, while these tools only allow 2. I don't remember why we kept it this way, but I vaguely recall some objections.

You're raising a good point with the difficulties to develop things for production only, the storage backend where IDE autocompletion are lacking is a good example.

I'm in favour of this change. However, I think we should go one step further and merge `base.txt` and `local.txt` so that we have only 2 files: one for developement, and one for production.

Thoughts?

foarsitter:
When we merge `base.txt` in to `production.txt` we can extent it in `local.txt`. 

In human speech it should sound like: we install dev dependencies on top of our production environment. 

It that what you propose?

browniebroke:
Yes, that what I meant, have 2 files: `local.txt` and `production.txt` with the local one starting with `-r production.txt`.

PR #2794, opened 2020-09-05
Updated Gitlab CI to use Python 3.8 instead of Python 3.7

arnav13081994:

[//]: # (Before you proceed:)

[//]: # (- Don't forget to update the `docs/` presuming others would benefit from a concise description of whatever that you're proposing)
[//]: # (- If you're adding a new option, please make sure that tests/test_cookiecutter_generation.py is updated accordingly)

## Description

Just updated the `.gitlab-ci.yml file` to use the same python version as is being used throughout the app.

## Rationale

So that the tests are run with the same python version as one is developing the project.

NA

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -265,6 +265,13 @@ Listed in alphabetical order.
     <td></td>
   </tr>
   <tr>
+    <td>Arnav Choudhury</td>
+    <td>
+      <a href="https://github.com/arnav13081994">arnav13081994</a>
+    </td>
+    <td></td>
+  </tr>
+  <tr>
     <td>AsheKR</td>
     <td>
       <a href="https://github.com/ashekr">ashekr</a>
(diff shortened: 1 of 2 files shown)


CONTRIBUTORS.md
```diff
@@ -264,6 +264,13 @@ Listed in alphabetical order.
     </td>
     <td></td>
   </tr>
+  <tr>
+    <td>Arnav Choudhury</td>
+    <td>
+      <a href="https://github.com/arnav13081994">arnav13081994</a>
+    </td>
+    <td></td>
+  </tr>
```
browniebroke, line 273: This file is now automatically generated, there was no need to update it. 

It's a bit new so this step might still be mentioned somewhere, let me know if this is the case, we should update it.

browniebroke (approved):

CONTRIBUTORS.md
arnav13081994: My bad. I thought this was supposed to be done manually. Should I remove that edit then and re-commit?

PR #2801, opened 2020-09-07
Traeffik and Django dockerfile changes

arnav13081994:

[//]: # (Before you proceed:)

[//]: # (- Don't forget to update the `docs/` presuming others would benefit from a concise description of whatever that you're proposing)
[//]: # (- If you're adding a new option, please make sure that tests/test_cookiecutter_generation.py is updated accordingly)

## Description

Reduce the number of image layers and speed up the build process slightly by combining varoius steps in Dockerfile into just one.  
## Rationale

In Django Dockerfile, I am proposing to combine ownership change step with file copying step. This would reduce the number of layers and potentially speed up the build process as well.

In Traefik Dockerfile, I am proposing to combine the 3 Run steps into 1 because if any one of them fails, the build should get aborted anyway and this would also reduce the number of layers and potentially speed up the build process as well.

NA


browniebroke (approved):
LGTM

browniebroke:
Thanks! 🎉

PR #2808, opened 2020-09-10
Add moto to local.txt

Andrew-Chen-Wang:
## Description

Checklist:

I've added Moto, a mock AWS service library, designed to for helping test your boto3 code. 

## Rationale

Although this repository doesn't use boto3 in development since it uses Django-storages, for others writing stuff like presigned posts or testing other methods of uploading media (e.g. ML checking if the image has at least one face), it'd be great to have something like this.

That's when moto comes in: instead of continuously sending requests to your S3 bucket especially during CI, incurring lots of costs, just use the mock service during testing.


browniebroke:
Thanks for suggesting this. 

It's true that `moto` can be quite convenient, but it's not a must have IMO. Not everyone will need it, and it comes at a cost, with a long list of dependencies. Depending on the services that one need to mock, it needs [some extra dependencies](https://github.com/spulec/moto/blob/9bc6bded6e6472f04a45f0bc429acc926c2cdc08/setup.py#L106-L122).

For the above reasons, I'm -1 on adding this.

Issue #1988, opened 2019-03-21
Pin nested Python dependencies 


browniebroke:
## Description

It's been a while that it's been good practice to pin all your dependencies. More and more, it's also becoming a good practice to also pin your nested dependencies. This is something we currently don't do.

## Rationale

Recently, we had a couple of issues where upgrades of nested dependencies broke the generated projects:

* #1949 
* #1953 
* #1954 

## Use case(s) / visualization(s)

It would be nice for `cookiecutter-django` to adopt this best practice, which would also avoid breaking too easily. The 3 tools trying to solve this problem that I know of are:

* [pip-tools](https://github.com/jazzband/pip-tools)
* [pipenv](https://github.com/pypa/pipenv)
* [poetry](https://github.com/sdispater/poetry)

Pipenv & Poetry are a bit new and imply a radically different workflow. They were ruled out in the past as they only provide 2 sets of dependencies (dev & prod: #1621 #1425). Moreover, I'm not sure how maintainable the files they produce would be in the template with all the if/else branches we have.

So my personal favourite is pip-tools, the steps to do that would be:

1. Make our current `requirements.txt` files into `requirements.in` files
2. Replace pinned versions with ranges wherever we care (e.g. Django)
3. Generate the pinned `requirements.txt`
4. Add the various if/else to the generated requirements.txt

Then, on Travis, when pyup sends us an update, we would need to check that the `requirements.in` don't produce any changes and is compatible. This is how they do it on [Wharehouse (a.k.a the new PyPI)](https://github.com/pypa/warehouse). Ideally, we would do that for all combinations of the template (#591 would help).

PS: Some [more reading](https://hynek.me/articles/python-app-deps-2018/) on this topic.

yunti:
pip-tools looks like it could work but I haven't tried it to be sure.  Pipenv has an unsafe by default api so that it's currently impossible to update only a single dependency. Poetry has a safe by default api, and seems to have recently fleshed out their overall api design, they are approaching version 1.0 but not quite there yet.

yunti:
We seem to be just using a local and production list of dependencies (with a base) which would be supported by Poetry so that could be an option.

carlmjohnson:
I have had a good experience with piptools so far. It's the least disruptive to current practices.

ollema:
piptools is really great!

demestav:
[...]

PR #2810, opened 2020-09-11
Merge base requirements into production and reference in local

foarsitter:
This pull request merges the requirements in base.txt in production.txt and references production.txt in local.txt. This way we have only two files of requirements and can we proceed with #1988.

There are two conflicts I cloud not resolve because they have another candidate for production than local: hiredis & pycopg2-binary. I moved them to production.txt and kept it as broad as possible.

closes #2776


mcsitter (approved):

foarsitter:
Updated this pull request to how I use it in production for over an year now. The `pycopg2-binary` requirement is dropped in favor of `pycopg2` so it matches production.

PR #2824, opened 2020-09-13
Making docs file smaller by around 40% using multi-stage builds

arnav13081994:
## Description

I am proposing to use multi-stage builds for building the python part of Docs.

This can be done by installing all required OS and python dependencies and then to just copy over those files to the run stage. This would cause a nearly 35-40% reduction in size!

Checklist:

## Rationale

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ All enhancements and patches to Cookiecutter Django will be documented in this f
 
 <!-- GENERATOR_PLACEHOLDER -->
 
+## [2020-09-14]
+### Fixed
+- Downgrade Celery to 4.4.6 ([#2829](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2829))
+### Updated
+- Update sentry-sdk to 0.17.5 ([#2828](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2828))
+- Update coverage to 5.3 ([#2826](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2826))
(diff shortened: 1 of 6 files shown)


browniebroke:
While this is a great improvement in #2815, especially for production image, I'm less convinced of the benefit of the docs image. This size improvement makes the `Dockerfile` a bit more complicated after all, but the `docs` image doesn't need to be deployed anywhere AFAIK, one can build it and just keep its output (which are a bunch of static files).

arnav13081994:
@browniebroke Agreed. I did this just to be consistent with the Django Dockerfile.

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@ All enhancements and patches to Cookiecutter Django will be documented in this f
 
 <!-- GENERATOR_PLACEHOLDER -->
 
+## [2020-09-14]
+### Fixed
+- Downgrade Celery to 4.4.6 ([#2829](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2829))
+### Updated
+- Update sentry-sdk to 0.17.5 ([#2828](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2828))
+- Update coverage to 5.3 ([#2826](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2826))
+- Update django-storages to 1.10.1 ([#2825](https://api.github.com/repos/pydanny/cookiecutter-django/pulls/2825))
```
browniebroke, line 12: Can you check how you've updated this PR? I think this shouldn't appear on it, it's already in master...

arnav13081994:
@browniebroke I forgot to merge upstream changes with my repo and that caused all of that. I have merged all changes and raised another PR. https://github.com/pydanny/cookiecutter-django/pull/2835

Closing this.

PR #2834, opened 2020-09-16
Centered the messages div in the base.html template.

PersonRP7:
## Description

The base.html alert div should be centered so as to render the message in the middle.

Checklist:

## Rationale

The default setting aligns the rendered message completely to the left. I believe this looks ungainly and should be rectified so as to maintain a symmetrical amount of padding on each side.

## Testing concerns

This is a small cosmetic change falling under the purview of front end testing, which to my understanding is not present within this project's testing structure.

## Separation of concerns

In the event of this pull request being accepted, concerns might be raised about implanting CSS within HTML tags. This change might also be implemented in the project's default CSS file, within the static directory.

## Demonstration

messages_before_after

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
@@ -80,7 +80,7 @@
 
       {% if messages %}
           {% for message in messages %}
-              <div class="alert {% if message.tags %}alert-{{ message.tags }}{% endif %}">{{ message }}<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>
+              <div align="center" class="alert {% if message.tags %}alert-{{ message.tags }}{% endif %}">{{ message }}<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>
           {% endfor %}
       {% endif %}
 


Andrew-Chen-Wang:
Instead of using align, use bootstrap's text-center class?

PersonRP7:
That's an elegant solution. Also, the close button on the right hand side of the message isn't centered. I've modified it using relative positioning and setting a margin-top in the project's css file. Should I open a pull request for this, or am I being too fussy with inconsequential details?

In any case, I thank you for your input.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
```diff
@@ -80,7 +80,7 @@
 
       {% if messages %}
           {% for message in messages %}
-              <div class="alert {% if message.tags %}alert-{{ message.tags }}{% endif %}">{{ message }}<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>
+              <div align="center" class="alert {% if message.tags %}alert-{{ message.tags }}{% endif %}">{{ message }}<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>
```
Andrew-Chen-Wang, line 83:
```suggestion
              <div class="text-center alert {% if message.tags %}alert-{{ message.tags }}{% endif %}">{{ message }}<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>
```

Andrew-Chen-Wang (changes requested):
I don't think centering the messages should really... happen? I think cosmetically, most websites have the text from the left so that, if a user is on mobile, then your paragraph of text from an error or success message doesn't look like a haiku.

Having the x on the right side is typical for most apps as a small standard that people have taken upon themselves to follow. This is all just personal opinion though!

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
PersonRP7: Thank you for your timely response. I find myself agreeing with you. Regarding the close button, I didn't mean changing its side, I meant that it's not aligned in the vertical plane, as in, it's closer to the bottom of the message.

Andrew-Chen-Wang: Ah I see. Sorry for not understanding. I'm no HTML expert, vehemently despising every moment of programming it, but I wanted to give some input lol

PersonRP7: No need for apologies. I understand the sentiment.

PR #2836, opened 2020-09-17
Making docs image 40% smaller and also making python version upgrades easier for multi-stage builds.

arnav13081994:
## Description

I am proposing to use multi-stage builds for building the python part of Docs.

This can be done by installing all required OS and python dependencies and then to just copy over those files to the run stage. This would cause a nearly 35-40% reduction in size!

Checklist:

## Rationale


browniebroke (approved):
I've merged the latest changes from master and tweaked a couple of things.

browniebroke:
Thank you for your contribution 🎉

arnav13081994:
Thanks @browniebroke 

Happy to help.

PR #2837, opened 2020-09-17
Adding GitHub-Action CI Option

arnav13081994:
## Description

Adding an option to let users choose GitHub action as a CI tool.

Checklist:

## Rationale

Since this code repo is on Github it would make more sense to take advantage of the deep CI integration that GitHub now provides out of the box like dependency caching to speed up future builds.


arnav13081994:
@browniebroke I am having trouble understanding why the checks are failing. Looks like `project_generation` is exiting with code 1. I have updated the tests etc but don't understand why this is happening. Any help would be appreciated.

{{cookiecutter.project_slug}}/.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,95 @@
+name: Github-CI
```
browniebroke, line 1: As this file lives in the `.github` folder, I think we could call it `ci.yml` and name it `CI` here.

{{cookiecutter.project_slug}}/.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Get pip cache dir
+        id: pip-cache-location
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+
+      - name: Cache pip Project Dependencies
+        uses: actions/cache@v2
+        with:
+          # Get the location of pip cache dir
+          path: ${{ steps.pip-cache-location.outputs.dir }}
```
browniebroke, line 80: These parts with double curly braces `${{ ... }}` will confuse the cookiecutter CLI, since it uses Jinja2 under the hood: it will try to interprete them as Jinja2 variables, but will fail.

You need to temporarily disable Jinja2 rendering with `{% raw %}` and `{% endraw %}`:

```suggestion
          path: {% raw %}${{ steps.pip-cache-location.outputs.dir }}{% endraw %}
```

Same for the others Github actions variables below.

docs/project-generation-options.rst
```diff
@@ -119,6 +119,7 @@ ci_tool:
     1. None
     2. `Travis CI`_
     3. `Gitlab CI`_
+    4. `Github CI`_
```
browniebroke, line 122: Should we label this as Github Actions instead of Github CI?

browniebroke (commented):
I think the issues might have to do with Jinja2 trying to interpolate GH actions variables (see below).

{{cookiecutter.project_slug}}/.github/workflows/github-ci.yml
arnav13081994: Ok. I'll do that.

docs/project-generation-options.rst
arnav13081994: Sure, but I think having `CI` should make it immediately obvious in case someone hasn't heard of them before. 
But `Github CI` sounds fine too.

{{cookiecutter.project_slug}}/.github/workflows/github-ci.yml
arnav13081994: Ok, I'll give it a go.

browniebroke (approved):

PR #2838, opened 2020-09-17
Made Traefik conf much easier to understand and improved redirect res…

arnav13081994:
## Description

Redirection of incoming requests from `http` to `https`  should happen at traefik `Entrypoint` as opposed to the `Middleware` layer.

Checklist:

## Rationale

1. Redirecting requests at `Entrypoint` will be faster as it's happening much earlier in the traefik request lifecycle. The typical flow in this scenrio would be:

**Current**:
 `Entrypoint (web)` --> `Router` --> `Middleware` -->  `Entrypoint (web-secure)`  --> `Router` --> `Middleware` --> .....

**Proposed**:
 `Entrypoint (web)` -->  `Entrypoint (web-secure)` --> `Router` --> `Middleware` --> .....

2. The resulting `traefik.yml` file is much easier to understand.
3. By redirecting to another entrypoint, one will have more control over where the request would go and which service will handle it. For ex: In case one wants to run `traefik Dashboard` with https but on port 8080 then the current setup will just redirect to https and in the process will get rid of `:8080` of the url. This can result in url name clashing issues because paths like `api` and `dashboard` are commonly used in APIs etc.


{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
```diff
@@ -5,6 +5,11 @@ entryPoints:
   web:
     # http
     address: ":80"
+    http:
+      redirections:
+        entryPoint:
+          to: web-secure
+          scheme: https
```
browniebroke, line 12: Could we add a comment linking to the relevant piece of Traefik documentation?

browniebroke (commented):
Really nice, looks much simpler indeed (I haven't tested it yet).

{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
arnav13081994: Made and pushed the change. Also I removed [`scheme: https`](https://docs.traefik.io/routing/entrypoints/#entrypoint) as it is the default value anyway

browniebroke (approved):
Made some testing today, and seems to be working 👍

browniebroke:
Thanks for the contribution 🎉

arnav13081994:
Happy to help.

Issue #2841, opened 2020-09-19
Traefik uses weak security settings by default


jameswilliams1:
## tl;dr
Traefik is configured using its default settings leaving it more open to attacks - a few tweaks could bring it up to production standard and lower the attack surface. See [this report](https://www.ssllabs.com/ssltest/analyze.html?d=spoter.co.uk&hideResults=on) of a site made using cookiecutter-django.

## Issues
- There is support for legacy tls versions that use weak security
- The full range of ciphers are used, including older and weaker ones that have been broken already
- There is no cipher preference: ideally websites should prefer ciphers in order of strength to ensure the weaker ones are used as an absolute last resort (or just unsupport them entirely)
- There are server name indication (SNI) issues with the site - browsers that don't support SNI are served the TRAEFIK_DEFAULT_CERT which is not trusted as it is self-signed
- Forward secrecy is not enabled properly - this means an attacker who breaks through any tls ciphers can also decode previous messages made to the site

## Fixes needed
- Remove support for TLS <1.2 (this affects only very, very old browsers which probably arent supported by bootstrap/other libs anyway)
- Reduce the range of ciphers used to only the strongest ones
- Set preference for the strongest ciphers
- Require SNI for browsers that access the site (this is all modern browsers, and even quite old ones)
- Ensure forward secrecy is present always, or at a minimum ensure ciphers that dont gurantee forward secrecy are only used if there is no better option in the client browser

Also slightly related, not big enough for its own issue: copying the traefik.yaml as part of the docker build is not really ideal as it prevents hot-reloading the config (a nice feature of traefik) - you need to completely rebuild to update config. Simply mounting the file read-only in production.yml would fix this.

I can make a PR if someone seconds these changes.

browniebroke:
Thanks for raising this issue. We're open to pull request(s) to fix these security issues. Some are definitely non-obvious and less experienced people would benefit a lot from someone experienced with Traefik tuning these.

While you're on the subject of tuning Traefik, you might also be interested in #1992. A fix was started in #2217 but it somewhat stalled now, and it would be great to fix it.

jameswilliams1:
@browniebroke I have an updated config for this that gets an A+ on ssllabs now, I will make a PR shortly. Worth also mentioning a major improvement to security would be to add a `Content-Security-Policy` header, basically only allowing scripts/styles/ajax calls from certain domains (majorly reduces the risk of XSS attacks, [more details here](https://content-security-policy.com/).

[...]

Issue #1992, opened 2019-03-25
SEC: Traefik should run as non-root


westurner:
Traefik should not run as root. When traefik runs as nonroot, it cannot bind to 80 or 443; but you can map 80 and 443 to e.g. 8080 and 8443 with Docker (e.g. in the docker-compose.yml production.yml). 

In order to run Traefik as non-root on other ports, I had to write a redirect rule in my traefik.toml; "You should not run as root" https://github.com/containous/traefik-library-image/issues/38#issuecomment-476142425

You can specify the user as `user: uid:gid` in the docker-compose.yml; or in the Dockerfile. AFAIU, the user does not have to be added first.
https://forums.docker.com/t/how-can-we-add-uid-and-gid-in-stack-file-or-compose-file/68261/3

- https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/production.yml

```yml
# ...
   user: 1005:1005
```

https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/Dockerfile

```
EXPOSE 8080/tcp 8443/tcp
USER 1005:1005  # or:  user: uid:gid
```

- https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/traefik.toml

```toml
[entryPoints]
  [entryPoints.http]
  address = ":8080"
    [entryPoints.http.redirect]
    #entryPoint = "https"
    regex = "^http://(.*):8080/(.*)"
    replacement = "https://$1/$2"

  [entryPoints.https]
  address = ":8443"
    [entryPoints.https.tls]
      #[[entryPoints.https.tls.certificates]]
      #certFile = "/certs/website.crt"
      #keyFile  = "/certs/website.key"
```

Without these default certs, when you access traefik by IP (instead of by the configured hostname); there's a default traefik cert that says 'traefik' in teh CN.

PR #2854, opened 2020-09-23
Improve Traefik secuity settings, add security documentation

jameswilliams1:
## Description

Update Traefik settings to increase TLS strength, hide server software and enforce the same security Django uses for Flower (when it is used). Also add new documentation on security and explain any caveats in compatibilty caused by security settings.

Checklist:

## Rationale

Fixes: #2841 #1992

The security settings used by Traefik were previously as default, this PR forces Traefik to use better TLS ciphers plus various other small security tweaks. When using docker, some HTTP secutity headers set by django are now set by Traefik so that they are also used by Flower (if celery is used) and any other parts of the site not served by Django.

Traefik has also been configured to run as non-root and to mount the `traefik.yml` rather than copy in the Dockerfile so that live reloading can be used when making changes.


jameswilliams1:
@browniebroke I think given there is already a fair bit of diff, CSP makes sense as its own issue. That would also need a lot of manual testing to ensure it doesn't block needed site features.

browniebroke:
I haven't looked into details yet, but from a quick glance, it looks like some very solid work here @jameswilliams1!

Since this is touching with security settings, I don't want to rush the review and give it a good look. 

One question I have (which might be a stupid one) is to know whether it makes sense to have some security settings both in Django and Traefik? You mention the Django deploy checks will raise false positives for example...

> I think given there is already a fair bit of diff, CSP makes sense as its own issue.

Agree

jameswilliams1:
@browniebroke yeah thats something I was deliberating with. On one hand I guess it's good to have a fallback, say if someone messes with the traefik.yml and removes the security and then django still catches it. But on the other config in 2 places gets confusing (as traefik will completely overwrite the django settings regardless of changes to production.py). I guess an option is to add the django settings back with a NOTE saying these are a fallback and won't affect anything when traefik is in use?

docs/security.rst
```diff
@@ -0,0 +1,52 @@
...
+.. _`SNI supported browsers`: https://caniuse.com/sni
+
+Useful Links
+------------
+- `Django Deployment Checklist`_
+- `SSL and TLS Deployment Best Practices`_
+- `Content Security Policy (CSP) Quick Reference Guide`_
+- `Mozilla Web Security Cheat Sheet`_
+
+.. _`Django Deployment Checklist`: https://docs.djangoproject.com/en/3.1/howto/deployment/checklist/
```
Andrew-Chen-Wang, line 49:
```suggestion
.. _`Django Deployment Checklist`: https://docs.djangoproject.com/en/dev/howto/deployment/checklist/
```

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -42,17 +42,14 @@
 
 # SECURITY
 # ------------------------------------------------------------------------------
-# https://docs.djangoproject.com/en/dev/ref/settings/#secure-proxy-ssl-header
-SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+{% if cookiecutter.use_docker -%}
+# NOTE headers are managed by the security-headers middleware in traefik.yml
```
Andrew-Chen-Wang, line 46: I wonder if we should comment the code out instead of removing these entirely...? For example, I deploy on AWS using Docker, so I would like to have these headers.

Andrew-Chen-Wang (changes requested):
Looks fine to me. Just a single concern.

{{cookiecutter.project_slug}}/config/settings/production.py
jameswilliams1: That seems like a good compromise to me, maybe a note to say these settings won't take effect if using traefik

Andrew-Chen-Wang: Cool. Something like "Uncomment if not using Traefik"

jameswilliams1:
@Andrew-Chen-Wang completely forgot about this PR, have just rebased and ci has passed now, could you review again?

Andrew-Chen-Wang (approved):
This LGTM now! I haven't tested this code due to limited resources. @browniebroke opinion?

foxwoods369:
Shouldn't production.yml have the ports remapped for traefik?
e.g. "0.0.0.0:80:8080"

alexanderbernath:
Sorry for bringing this old issue up, but is it still relevant? The topic seems important. Thanks.

PR #2865, opened 2020-09-30
Fix missing Django Debug toolbar with node container

lcd1232:
## Description

Show debug panel at localhost:3000

Checklist:


{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -69,6 +69,8 @@
 
     hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
     INTERNAL_IPS += [".".join(ip.split(".")[:-1] + ["1"]) for ip in ips]
+    _, _, ips = socket.gethostbyname_ex("node")
+    INTERNAL_IPS.extend(ips)
```
browniebroke, line 73: The build failed on Docker because it doesn't know about `node` service. I guess it should be conditional to using a JS task runner:

```suggestion
    {%- if cookiecutter.js_task_runner == 'Gulp' %}
    _, _, ips = socket.gethostbyname_ex("node")
    INTERNAL_IPS.extend(ips)
    {%- endif %}
```

Could we add this task runner to the extended docker test in Travis? What I mean by that is changing the `script` part below:

https://github.com/pydanny/cookiecutter-django/blob/219af571b30d4533631844333bcfad6ad9faee4d/.travis.yml#L20-L21

to:

```
script: sh tests/test_docker.sh use_celery=y use_drf=y 
```

lcd1232:
It looks like there is a problem when container with Django starts before node container.

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -69,6 +69,10 @@
 
     hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
     INTERNAL_IPS += [".".join(ip.split(".")[:-1] + ["1"]) for ip in ips]
+    {%- if cookiecutter.js_task_runner == 'Gulp' %}
+    _, _, ips = socket.gethostbyname_ex("node")
+    INTERNAL_IPS.extend(ips)
```
browniebroke, line 74: So... This doesn't work, because you need the node container to be started, but the node container depends on th Django one:

https://github.com/pydanny/cookiecutter-django/blob/b48067c03494993933bd2eb0aa2e5afdf367f12f/%7B%7Bcookiecutter.project_slug%7D%7D/local.yml#L114-L115

I thought this dependency was needed for browsersync to work, but actually we might not need it, and we might be able to flip it (have django depends on node).

Anyway, we should probably catch [the error](https://github.com/pydanny/cookiecutter-django/pull/2865/checks?check_run_id=1758500147#step:4:1493) to not fail entirely in case `node` has an issue or is explicitely not started (like in tests):

```suggestion
    try:
        _, _, ips = socket.gethostbyname_ex("node")
        INTERNAL_IPS.extend(ips)
    except socket.gaierror:
        # The node container isn't started (yet?)
        pass
```

browniebroke (approved):

Issue #2827, opened 2020-09-14
Celery breaks upgrading from 4.4.6 to 4.4.7


jskitz:
## What happened?

The requirements in Django Cookie Cutter changed Celery from 4.4.6 to 4.4.7 recently. I upgraded my projects that were based off of this project, and then I could no longer get my tasks to be pulled in to production. There is more on this issue in [Celery issues](https://github.com/celery/celery/issues/6285)

## What should've happened instead?

Production tasks should have been properly registered as tasks. Instead, when you invoke tasks in production, it says that there are no registered tasks. This can be verified by running celery from an init.d script, opening a shell on production, and executing the following tasks:

```
from celery.task.control import  inspect
i = inspect()
i.registered_tasks()
```

## Additional details

* Host system configuration:
    * Celery is run on ubuntu via an init script generating a command like the following: `celery -c /home/ubuntu/venv/bin/celery multi start w1 --workdir=/home/ubuntu/project --pidfile=/var/run/celery/%n.pid --logfile=/var/log/celery/%n%I.log --loglevel=INFO --app=config.celery_app --time-limit=300 --concurrency=4`. This will not pass the `--app=config.celery_app` to the workers.
    * Version of cookiecutter CLI (get it with `cookiecutter --version`): 1.7.2
    * OS: Problem happens using production settings but not local settings.
    * Python version: 3.8
    * Docker versions (if using Docker): N/A

* Options selected and/or [replay file](https://cookiecutter.readthedocs.io/en/latest/advanced/replay.html): Choosing Celery with the latest build of django-cookiecutter will use 4.4.7 instead 4.4.6 which exhibits the problem.

The solution is simply to downgrade to Celery 4.4.6 until Celery pushes out a new version. Posting here to save other people a lot of time and headaches trying to debug what was happening with this recent upgrade.

browniebroke:
I saw this issue in Celery a few days ago, but didn't get around to fix it here. We shoudl definitely downgrade if it's affecting us.

Andrew-Chen-Wang:
Actually this is pretty interesting. The fix is using CELERY_APP as an environment variable, according to that repo, for now (which happens to be in the entrypoint file) since I didn't realize that my celery 4.4.7 on Docker was working fine. This may just be a local setup issue.

arnav13081994:
[...]

PR #2868, opened 2020-10-01
Auto merge Dependabot updates

### LINKED ISSUE #2852: Add Github action to auto merge dependabot and pyup security updates if all tests pass.
## Description

I propose to add a new GitHub action or job that will auto merge pull requests from the `pyup bot` and the `dependabot` should the PR so created has no merge conflicts with master and all tests also pass.

This will be very similar to how [pre-commit tool versions](https://github.com/pydanny/cookiecutter-django/blob/master/.github/workflows/pre-commit-autoupdate.yml#L28) in the repo root are kept up to date.

## Rationale

This would make updating requirements easier and more automated and would likely also make the project more secure. The tests are used to make the decision to merge updates from dependabot and pyup right now anyway.

Andrew-Chen-Wang:
## Description

Fixes #2852 Adds auto-merging pip updates to this repository using dependabot to alleviate @browniebroke 's workload.

## Rationale

Make updating this repository less stressful. You can view the test repository here: https://github.com/Andrew-Chen-Wang/test-dependabot-automerge You can see that, when one update happens, the rest of the PRs close and are superseded with a new PR, as shown here: https://github.com/Andrew-Chen-Wang/test-dependabot-automerge/pull/3#issuecomment-701741950

This also means we don't need pyup anymore, especially since dependabot covers everything pyup does and auto merges and rebases in case of new updates to the PR.

---

Note: when GitHub acquired dependabot, they disabled the auto merge feature [natively for security reasons](https://github.com/dependabot/dependabot-core/issues/1973#issuecomment-640918321), so that's why I had to create the GitHub action. I took the GitHub action from a Medium post, but I just can't find the blog anymore......

AH! Here it is: https://medium.com/@toufik.airane/automerge-github-dependabot-alerts-with-github-actions-7cd6f5763750

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,11 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: daily
+  - package-ecosystem: pip
+    directory: "/{{ cookiecutter.project_slug }}/requirements"
+    schedule:
+      interval: daily
--- a/.github/workflows/dependabot-merge.yml
+++ b/.github/workflows/dependabot-merge.yml
@@ -1,0 +1,27 @@
+name: "Dependabot Automerge - Action"
+
+on:
+  pull_request:
+
+jobs:
+  worker:
+    runs-on: ubuntu-latest
+
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: automerge
+        uses: actions/github-script@0.2.0
+        with:
+          script: |
+            github.pullRequests.createReview({
+              owner: context.payload.repository.owner.login,
+              repo: context.payload.repository.name,
+              pull_number: context.payload.pull_request.number,
+              event: 'APPROVE'
+            })
+            github.pullRequests.merge({
+              owner: context.payload.repository.owner.login,
+              repo: context.payload.repository.name,
+              pull_number: context.payload.pull_request.number
+            })
+          github-token: ${{github.token}}


browniebroke:
Very nice! I think we should be careful with this one. A few questions I have:

- What happens if the build doesn't pass? From a quick look, it doesn't seem to check the build status...
- Did you check that Dependabot supports having Jinja markup in requirements files?
  In latest version of your example repo, Django is updated [in production](https://github.com/Andrew-Chen-Wang/test-dependabot-automerge/blob/a249161f233f403737b6b6ee2c636375df6c2386/%7B%7B%20cookiecutter.project_slug%20%7D%7D/requirements/production.txt#L1), but [not in local](https://github.com/Andrew-Chen-Wang/test-dependabot-automerge/blob/a249161f233f403737b6b6ee2c636375df6c2386/%7B%7B%20cookiecutter.project_slug%20%7D%7D/requirements/local.txt#L1).
  The difference between the 2 files is that local has some Jinja markup for a conditional. I think that's of the main thing preventing up from replacing pyup by Dependabot. It's well supported by pyup.
- Another neat feature from pyup is the ability to "filter" updates, that we use often, and [currently for Celery](https://github.com/pydanny/cookiecutter-django/blob/1a701ed70f3275b282d8fc3a1d98fde36e4e35f5/%7B%7Bcookiecutter.project_slug%7D%7D/requirements/base.txt#L20), we can say `!= 4.4.7` for example. [Celery 4.4.7 is a bit broken](https://github.com/pydanny/cookiecutter-django/issues/2827) currently, and we don't want to update to it, but the Celery bug isn't something that breaks our build, so an upgrade would be automerged, and if we downgrade, it would be re-updated again...

I think we can't switch to Dependanbot yet, but if we can get the build status checked before merging, that would work with pyup changes.

Andrew-Chen-Wang:
Interesting, didn't notice it failed with the Jinja2 templating; I probably got too tired, sorry about that.

I can rework the GitHub to auto merge pyup instead by changing the actor. If you don't mind, for the CI check, I'd like to use: https://github.com/WyriHaximus/github-action-wait-for-status for checking the status. The merge step in this case will use an `if` the status is successful and if not, the merge will not happen.

browniebroke:
Yea, I think it makes sense to delegate this check 👍 

> sorry about that.

Don't apologize for a small mistake! This is some great work! 💪

Andrew-Chen-Wang:
Superseded by #2872

PR #2881, opened 2020-10-07
Add dedicated websockets package

Andrew-Chen-Wang:
## Description

Ref: https://github.com/encode/uvicorn/issues/797

There used to be a package automatically installed; not anymore I guess. This came from a separate package and not cookiecutter-django, but the installation was similar.

## Rationale

For websockets to work, you need a package like wsproto or websockets (which is not as maintained). Uvicorn at some point might have dropped the auto installation of one of these packages that made websockets unavailable to work with.

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -25,6 +25,7 @@ flower==0.9.5  # https://github.com/mher/flower
 {%- endif %}
 {%- if cookiecutter.use_async == 'y' %}
 uvicorn==0.12.1  # https://github.com/encode/uvicorn
+wsproto==0.15.0  # https://github.com/python-hyper/wsproto/
 {%- endif %}
 
 # Django


{{cookiecutter.project_slug}}/requirements/base.txt
```diff
@@ -25,6 +25,7 @@ flower==0.9.5  # https://github.com/mher/flower
 {%- endif %}
 {%- if cookiecutter.use_async == 'y' %}
 uvicorn==0.12.1  # https://github.com/encode/uvicorn
+wsproto==0.15.0  # https://github.com/python-hyper/wsproto/
```
browniebroke, line 28: Good catch, I believe it's because of https://github.com/encode/uvicorn/pull/666. A bit of a shame that there is no `extra_require` for it from `uvicorn`. I guess we'll have to pay attention to `uvicorn` updates in case they remove the integration or provide an extra for it.

Didn't know about this library, looks good.

browniebroke (approved):

Andrew-Chen-Wang:
@browniebroke It may be necessary to add these packages as well: https://github.com/encode/uvicorn/pull/666/files#diff-2eeaed663bd0d25b7e608891384b7298R53-R54

Those used to come with the standard installation of uvicorn. Although, without them, uvicorn seems to be running properly, I somewhat have my doubts for production mode.

PR #2888, opened 2020-10-10
Fix .env file reading

PraneethKarnena:
In order to properly read `.env` file, the below line has to be added:
`environ.Env.read_env()`

## Description

Checklist:

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -9,6 +9,7 @@ ROOT_DIR = Path(__file__).resolve(strict=True).parent.parent.parent
 # {{ cookiecutter.project_slug }}/
 APPS_DIR = ROOT_DIR / "{{ cookiecutter.project_slug }}"
 env = environ.Env()
+environ.Env.read_env()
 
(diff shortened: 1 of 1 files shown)


browniebroke:
This is done line 16 below if `DJANGO_READ_DOT_ENV_FILE` is True... This [is documented](https://cookiecutter-django.readthedocs.io/en/latest/developing-locally.html) and the expected behaviour. Unless I'm missing something, there is nothing to fix.

PraneethKarnena:
> This is done line 16 below if `DJANGO_READ_DOT_ENV_FILE` is True... This [is documented](https://cookiecutter-django.readthedocs.io/en/latest/developing-locally.html) and the expected behaviour. Unless I'm missing something, there is nothing to fix.

@browniebroke I'm using Django 2.2.16 with Cookiecutter. I'm unable to read the `.env` file in the first place until I have added `environ.Env.read_env()`. Since the environment variable, `DJANGO_READ_DOT_ENV_FILE` is added in the `.env` file, How could I read the variable if I'm not able to read the `.env` file itself?

browniebroke:
That's the only variable that you should set outside of the `.env` file (see #985).

browniebroke:
Closing as per above explanations.

PR #2927, opened 2020-11-04
Use defer for script tags (Fix #2922)

### LINKED ISSUE #2922: Use defer for script tags
https://github.com/pydanny/cookiecutter-django/blob/4b97071cfe83c6ef48b5c383b14b8bbbe15c0cb4/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/templates/base.html#L98-L119

I've been optimizing my project lately, and one thing I came across was the attributes async and defer for `<script>` tags. They both load asynchronously, but the defer executes after the content is loaded. Graphics in this website make it pretty clear why performance is improved:

https://flaviocopes.com/javascript-async-defer/

I was thinking, would it make sense to load those jquery and bootstrap and `project.js` scripts in the head with the defer flag? The other thing about defer that may not be mentioned in the website is that defer executes the scripts in order (so long as they all have defer in them), which can help some devs better organize their script loading.

Andrew-Chen-Wang:
Fixes #2922 

## Description

* This way, scripts from external URLs are loaded asynchronously. By putting it at the top of the file, the browser parses it first, downloads it while continuing to parse the HTML, and then executes on parsing finish
* Additionally, developers will not need to use $(window).ready() or the like in their files anymore.
* Added inline_javascript tag in case anyone wants to use the bottom of the HTML page to execute some Javascript. Using defer here has no effect as inline scripts defer by default

Checklist:

## Rationale

Improve performance by pre-loading (think of it like HTTP/2) the script tags but executing after parsing is done. Lets go of the need to wait for HTML parsing when you can just let defer do that for you. So instead of doing:

```
let isAuthenticated = false;
$(window).ready(function() {
    isAuthenticated = !document.getElementById("sign-up-link");
})
```
You can instead simply this to use without jQuery:
```
const isAuthenticated = !document.getElementById("sign-up-link");
```
I had this conversation on Discord with some other JS developers and they said, even though the HTML may be fully parsed, doing the latter would still not work 5-10% of the time due to a browser being too slow to parse, even though the script tag is at the bottom.


Andrew-Chen-Wang:
I'd like to also improve email.html's inline javascript. Hopefully encouraging people to move away from jQuery as lots of features from jQuery are now in ECMAScript 6 thus removing the performance degradation of using jQuery.

So remove the `$(".form-class").removeClass("row")` in favor of:
```javascript
document.querySelectorAll('.form-class').forEach(el => el.classList.remove('row'));
```

Additionally, both variables should use ECMAScript 6's terms, specifically in this case `const` (`let` means allowing reassignment, but I don't see any of the variables being reassigned).

But I can create a new PR for that after this one is merged (or closed).

Andrew-Chen-Wang:
I see Travis CI does not like me one bit. None of these PRs are running........

browniebroke (approved):

PR #2931, opened 2020-11-09
Migrate CI to Github Actions

### LINKED ISSUE #2928: Use GitHub Actions for this instead of Travis's new pricing
## Description

To test cookiecutter-django itself, we should use GitHub actions which is free and unlimited for open source projects.

## Rationale

Travis recently announced a new pricing model by using a credits system. https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing Basically, we're gonna run out of builds REALLY quickly (if not already based on my two latest PRs and the current master branch commit).

GitHub actions [is completely free and unlimited for open source projects](https://github.com/features/actions). I don't see actions becoming some billing system for open source projects in the next few years.

browniebroke:
## Description

Migrate CI to Github Actions cc @Andrew-Chen-Wang @arnav13081994 

Checklist:

This doesn't aim at adding any sort of caching but just to keep feature parity with Travis for now.

## Rationale

Travis recently changed their pricing with hard limits to open source projects, Github actions have a higher allowace at the moment.

Fixes #2928


Andrew-Chen-Wang:
It LGTM. Just regarding 46a0b60a29d34d3b8654382408541a0c52b890e6, does pydanny commit directly to master? There might be a way to remove `branches: [master]` and add `if github.actor == pydanny` ([example?](https://github.com/SimpleJWT/drf-SimpleJWT-Vue/blob/376bb50defb47d46286b6a8360f916fe0feedc86/.github/workflows/automerge.yml#L10)) to prevent pyup from triggering but still allow pydanny to commit wherever.

browniebroke:
I -and all core contributors- can commit directly to master as well.

luzfcb:
@browniebroke Why do we need the 46a0b60a29d34d3b8654382408541a0c52b890e6 change? 
From what I understand, this will disable ci on all pull-requests...
at least it was the conclusion I got testing this pr with (https://github.com/luzfcb/cookiecutter-django/pull/4) and without (https://github.com/luzfcb/cookiecutter-django/pull/5) the 46a0b60a29d34d3b8654382408541a0c52b890e6 commit

.github/workflows/ci.yml
```diff
@@ -0,0 +1,85 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
```
sfdye, line 5:
```suggestion
    branches: master
```

sfdye:
> @browniebroke Why do we need the [46a0b60](https://github.com/pydanny/cookiecutter-django/commit/46a0b60a29d34d3b8654382408541a0c52b890e6) change?
> From what I understand, this will disable ci on all pull-requests...

I think that is fine. The pull requests CI will be just fine. Plus with `pull_request` it listens to reopen event as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,85 @@
...
+  pull_request:
+
+jobs:
+  tox:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
sfdye, line 12: when don't we need fail fast?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,85 @@
...
+        with:
+          python-version: 3.8
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          python -m pip install -U tox
+      - name: Tox ${{ matrix.tox-env }}
+        run: tox -e ${{ matrix.tox-env }}
+
+  docker:
```
sfdye, line 30: why not just directly translate from travis by using matrix for docker and bare metal?

sfdye:
I've recently moved 20+ repos from Travis CI to Github Action for my organization, let me know if you need help with this PR

Andrew-Chen-Wang:
@luzfcb Pyup uses several branches itself in this repository. That would mean pyup is running the CI in a branch here and in a PR.

browniebroke:
> @browniebroke Why do we need the 46a0b60a29d34d3b8654382408541a0c52b890e6 change? 
From what I understand, this will disable ci on all pull-requests...
at least it was the conclusion I got testing this pr with (https://github.com/luzfcb/cookiecutter-django/pull/4) and without (https://github.com/luzfcb/cookiecutter-django/pull/5) the 46a0b60a29d34d3b8654382408541a0c52b890e6 commit

Hum, that's really odd. I was trying to avoid the duplicated status that you have in https://github.com/luzfcb/cookiecutter-django/pull/5 where each job run twice, once for the `push` event and once for the `pull_request` event. I mean, it's fine in this pull request as far as I can tell. I don't mind removing this last commit... 

> but I suspect it's because actions aren't run on forks

I suspect too, but again, it's strange that both run fine in the other case...

.github/workflows/ci.yml
browniebroke: With `fail-fast` the whole job would be interrupted as soon as one of the subjobs fails. In my experience it's nicer to get both to run entirely to get all failures reported as opposed to only the first failure.

.github/workflows/ci.yml
browniebroke: Right now, I could probably merge the `docker` and `tox` jobs, but I'm hoping that we could speed up each job leter by adding some specialised caching. 

The bare metal job requires some extra services that the other jobs don't use.

.github/workflows/ci.yml
browniebroke: Does this work? I though this part is required to be an array. I couldn't find an example with a string [in the docs](https://docs.github.com/actions/reference/events-that-trigger-workflows#configuring-workflow-events)... Do you have a working example?

sfdye: I have many working examples, but they are all private repos. Anyways this is /nit

sfdye (approved):

luzfcb (approved):
LGTM

PR #2932, opened 2020-11-12
Upgrade factory-boy to 3.1.0

umgelurgel:
## Description

I've upgraded `factory-boy` to version `3.1.0` and updated the `tests.factories.UserFactory` to work with the new version. `factory.Faker.generate` signature has been modified in version 3.1.0; it no longer accepts an `extra_kwargs` argument, instead requires a `params` argument ([associated PR])(https://github.com/FactoryBoy/factory_boy/commit/f0a4ef008f07f8d42221565d8c33b88083f0be6d). The signature of the generate method is private ([discussed here](https://github.com/FactoryBoy/factory_boy/issues/790)), so looks like this should be fixed in downstream packages.

Checklist:


browniebroke (approved):
Nice one!

PR #2939, opened 2020-11-19
Fix after uvicorn 0.12.0 - Ship extra dependencies

### LINKED ISSUE #2915: Deploying on Heroku. ModuleNotFoundError: No module named 'uvloop'
Received a  ModuleNotFoundError: No module named uvloop when deploying my app to Heroku..

Not sure if this is a breaking change.

**My solution**
I did bypass this by rolling back --> [https://github.com/pydanny/cookiecutter-django/pull/2866](https://github.com/pydanny/cookiecutter-django/pull/2866)

In requirements/base.txt 

- Change this line --> `uvicorn==0.12.1  # https://github.com/encode/uvicorn`
- To this --> `uvicorn==0.11.8  # https://github.com/encode/uvicorn`
- Redeploy

Will this have any further implicatons for my project down the line?

eqqe:
Uvicorn no longer ships extra dependencies uvloop, websockets and httptools as default. To install these dependencies use uvicorn[standard].
cf https://github.com/encode/uvicorn/blob/master/CHANGELOG.md

## Description

Checklist:

## Rationale

Fixes #2915

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -24,7 +24,7 @@ flower==0.9.5  # https://github.com/mher/flower
 {%- endif %}
 {%- endif %}
 {%- if cookiecutter.use_async == 'y' %}
-uvicorn==0.12.2  # https://github.com/encode/uvicorn
+uvicorn[standard]==0.12.2  # https://github.com/encode/uvicorn
 wsproto==0.15.0  # https://github.com/python-hyper/wsproto/
 {%- endif %}
 
--- a/{{cookiecutter.project_slug}}/requirements/base.txt
+++ b/{{cookiecutter.project_slug}}/requirements/base.txt
@@ -25,7 +25,6 @@ flower==0.9.5  # https://github.com/mher/flower
 {%- endif %}
 {%- if cookiecutter.use_async == 'y' %}
 uvicorn[standard]==0.12.2  # https://github.com/encode/uvicorn
-wsproto==0.15.0  # https://github.com/python-hyper/wsproto/
 {%- endif %}
 
 # Django


{{cookiecutter.project_slug}}/requirements/base.txt
```diff
@@ -24,7 +24,7 @@ flower==0.9.5  # https://github.com/mher/flower
 {%- endif %}
 {%- endif %}
 {%- if cookiecutter.use_async == 'y' %}
-uvicorn==0.12.2  # https://github.com/encode/uvicorn
+uvicorn[standard]==0.12.2  # https://github.com/encode/uvicorn
 wsproto==0.15.0  # https://github.com/python-hyper/wsproto/
```
browniebroke, line 28: Do we still need `wsproto`? I think it's just [a fallback to `websocket`](https://github.com/encode/uvicorn/blob/a839da2564e5168c6e99c3b22819950f26f7d178/uvicorn/protocols/websockets/auto.py#L1-L5) which gets installed with the `standard` extra. So it's now effectively unused...

```suggestion
uvicorn[standard]==0.12.2  # https://github.com/encode/uvicorn
```

browniebroke (commented):
Nice one!

{{cookiecutter.project_slug}}/requirements/base.txt
Andrew-Chen-Wang: tl;dr It might just be best to remove wsproto.

If you want to block initial connection without accepting (e.g. because user is not authenticated), you'll have no choice but to use websockets. If you accept then immediately close the connection (making that two messages sent from the server rather than one), then you can use wsproto until they fix the issue I raised in their repo.

Andrew-Chen-Wang, line 28:
```suggestion
```

Remove wsproto until this issue is fixed: https://github.com/encode/uvicorn/issues/811

Andrew-Chen-Wang (changes requested):

browniebroke (approved):

browniebroke:
Thanks for the fix!

PR #2949, opened 2020-11-26
Add test for UserUpdateView.form_valid()

thorrak:
## Description

This implements a simple test for UserUpdateView.form_valid() in the `users` app to ensure that the message this method seeks to add to the request gets added.

This also updates the message itself to correct a typo.

Checklist:

- [X] ~I've made sure that `tests/test_cookiecutter_generation.py` is updated accordingly (especially if adding or updating a template option)~ Not applicable for this change

## Rationale

There was exactly one missing test case per coverage coming from the stock install of cookiecutter-django in my app. This tests that one missing case.


browniebroke:
Thanks!

PR #2963, opened 2020-12-08
Fix CVE-2018-1281

jackton1:
References:
- https://nvd.nist.gov/vuln/detail/CVE-2018-1281
- https://bandit.readthedocs.io/en/latest/plugins/b104_hardcoded_bind_all_interfaces.html

```
Test results:
>> Issue: [B104:hardcoded_bind_all_interfaces] Possible binding to all interfaces.
   Severity: Medium   Confidence: Medium
   Location: ../local.py:4
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b104_hardcoded_bind_all_interfaces.html
3	# https://docs.djangoproject.com/en/3.1/ref/settings/#allowed-hosts
4	ALLOWED_HOSTS = ["localhost", "0.0.0.0", "127.0.0.1"]
5
```

## Description

Checklist:

- [x] ~~I've made sure that `tests/test_cookiecutter_generation.py` is updated accordingly (especially if adding or updating a template option)~~ Not applicable
- [x] ~~I've updated the documentation or confirm that my change doesn't require any updates~~ Not applicable

## Rationale

Ensure that allowed host doesn't include a non-routable meta-address.

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -208,17 +208,17 @@ Increasingly it is becoming necessary to develop software in a secure environmen
 On order to create a secure environment, we need to have a trusted SSL certficate installed in our Docker application.
 
 #.  **Let's Encrypt**
-    
-    The official line from Let’s Encrypt is: 
 
-    [For local development section] ... The best option: Generate your own certificate, either self-signed or signed by a local root, and trust it in your operating system’s trust store. Then use that certificate in your local web server. See below for details. 
+    The official line from Let’s Encrypt is:
+
+    [For local development section] ... The best option: Generate your own certificate, either self-signed or signed by a local root, and trust it in your operating system’s trust store. Then use that certificate in your local web server. See below for details.
 
     See `letsencrypt.org - certificates-for-localhost`_
 
     .. _`letsencrypt.org - certificates-for-localhost`: https://letsencrypt.org/docs/certificates-for-localhost/
 
 #.  **mkcert: Valid Https Certificates For Localhost**
-    
+
     `mkcert`_ is a simple by design tool that hides all the arcane knowledge required to generate valid TLS certificates. It works for any hostname or IP, including localhost. It supports macOS, Linux, and Windows, and Firefox, Chrome and Java. It even works on mobile devices with a couple manual steps.
 
     See https://blog.filippo.io/mkcert-valid-https-certificates-for-localhost/
@@ -253,11 +253,11 @@ local.yml
       restart: always
       depends_on:
         - django
-    
+
     ...
 
 #. Link the ``nginx-proxy`` to ``django`` through environmental variables.
-   
+
    ``django`` already has an ``.env`` file connected to it. Add the following variables. You should do this especially if you are working with a team and you want to keep your local environment details to yourself.
 
    ::
@@ -274,7 +274,7 @@ config/settings/local.py
 
 You should allow the new hostname. ::
 
-  ALLOWED_HOSTS = ["localhost", "0.0.0.0", "127.0.0.1", "my-dev-env.local"]
+  ALLOWED_HOSTS = ["localhost", "127.0.0.1", "my-dev-env.local"]
 
 Rebuild your ``docker`` application. ::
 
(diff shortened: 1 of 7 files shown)


browniebroke:
Thanks for the report, I've looking into why this value was added, I would assume there is a good reason for it, but I could only trace it back to https://github.com/pydanny/cookiecutter-django/commit/64f25417436e3adceecacb0761d52077782d0a3a which doesn't give much context.

In the current form, this is breaking our Docker setup, we use this IP address by default when starting the django server:

https://github.com/pydanny/cookiecutter-django/blob/c54c4f5ef7ebe4ee85e96dd9e72df4a49a7242d1/%7B%7Bcookiecutter.project_slug%7D%7D/compose/local/django/start#L10-L12 

I'd be curious to understand whether if we also need to update this part:

https://github.com/pydanny/cookiecutter-django/blob/624dd94dd3738d50f91655d350ca9bf22b6af68f/%7B%7Bcookiecutter.project_slug%7D%7D/production.yml#L43-L46

Also, for non-docker setup, the documentation needs to be updated here:

https://github.com/pydanny/cookiecutter-django/blob/8cd3a796dabeba41c850684566b09a7764d784bc/docs/developing-locally.rst#L77-L83

As well as here:

https://github.com/pydanny/cookiecutter-django/blob/4f7204744667184bc2fc48a2e039bf44282a1a27/docs/developing-locally-docker.rst#L277

Finally, to prevent this from happenning again, I think we could run bandit on our CI (this last part can be done separately though).

jackton1:
> Thanks for the report, I've looking into why this value was added, I would assume there is a good reason for it, but I could only trace it back to [64f2541](https://github.com/pydanny/cookiecutter-django/commit/64f25417436e3adceecacb0761d52077782d0a3a) which doesn't give much context.
> 
> In the current form, this is breaking our Docker setup, we use this IP address by default when starting the django server:
> 
> https://github.com/pydanny/cookiecutter-django/blob/c54c4f5ef7ebe4ee85e96dd9e72df4a49a7242d1/%7B%7Bcookiecutter.project_slug%7D%7D/compose/local/django/start#L10-L12
> 
> I'd be curious to understand whether if we also need to update this part:
> 
> https://github.com/pydanny/cookiecutter-django/blob/624dd94dd3738d50f91655d350ca9bf22b6af68f/%7B%7Bcookiecutter.project_slug%7D%7D/production.yml#L43-L46
> 
> Also, for non-docker setup, the documentation needs to be updated here:
> 
> https://github.com/pydanny/cookiecutter-django/blob/8cd3a796dabeba41c850684566b09a7764d784bc/docs/developing-locally.rst#L77-L83
> 
> As well as here:
> 
> https://github.com/pydanny/cookiecutter-django/blob/4f7204744667184bc2fc48a2e039bf44282a1a27/docs/developing-locally-docker.rst#L277
> 
> Finally, to prevent this from happenning again, I think we could run bandit on our CI (this last part can be done separately though).

👍  Done

jackton1:
For setting up bandit. I currently use https://fussyfox.github.io/ but if it needs to be part of tox / github actions that's also an alternative.

browniebroke:
I think fussyfox isn't suitable for us as it would run on the templated code, while we need it to run bandit on the result of the template.

jackton1:
> I think fussyfox isn't suitable for us as it would run on the templated code, while we need it to run bandit on the result of the template.

Yeah that's true, I'll setup a github action instead.

Andrew-Chen-Wang:
The Django ticket that could've been the reason for running `runserver` with that arg of `0.0.0.0:8000`: https://code.djangoproject.com/ticket/27537

browniebroke:
I tried to generate a fresh project with Docker and it doesn't work at the moment: I can't access the homepage locally. The console shows that we can access it at the IP `127.0.0.1` but when opening in the browserm it's not reachable. I also tried `localhost`, same issue.

Andrew-Chen-Wang:
Err isn't the point of having 0.0.0.0 supposed to be exposing the port 8000 of the Docker network? Because it's in a Docker container, the localhost you type in is not the same as the localhost of the Docker network. So even though you exposed port 8000 from the Docker network, Django is only accepting localhost from within the network.

jackton1:
@browniebroke I'm gonna pull this again and test. In the meantime I added bandit here #2974

PR #2975, opened 2020-12-17
`pip install cookiecutter` was missing

guettli:
## Description

Checklist:

### DIFF
--- a/docs/developing-locally.rst
+++ b/docs/developing-locally.rst
@@ -24,7 +24,7 @@ First things first.
 
     $ source <virtual env path>/bin/activate
 
-#. Install cookiecutter and cookiecutter--django: ::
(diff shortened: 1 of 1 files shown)


browniebroke:
Feels a bit redundant with what is written a few lines above, no?

https://github.com/pydanny/cookiecutter-django/blob/202e8d40a9f33a8254c62fb8cec2149dcea99e75/docs/developing-locally.rst#L10-L15

I think the general advice is that the cookiecutter CLI should be installed globally rather than in the project virtualenv, there is [a dedicated page in the cookiecutter docs](https://cookiecutter.readthedocs.io/en/latest/installation.html).

guettli:
Above I read "PostgreSQL" and "Redis", then my eyes skipped to the code.

I never install Python stuff globally. No problem, just ignore this. I close it.

PR #2989, opened 2020-12-31
Sentry Redis integration enabled by default in production.

vascop:
## Description
Enables Redis Sentry integration by default.

Checklist:

## Rationale

Redis is the default so Sentry can have its Redis integration enabled by default in the production settings.


browniebroke (approved):

browniebroke:
Thanks!

PR #2992, opened 2021-01-02
Use exception var in status code pages if available

Andrew-Chen-Wang:
## Description

Checklist:

## Rationale

When you raise PermissionDenied or Http404, you can pass an exception string. Having that output would be nice: https://docs.djangoproject.com/en/dev/ref/views/#the-403-http-forbidden-view

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
@@ -5,5 +5,5 @@
 {% block content %}
 <h1>Forbidden (403)</h1>
 
-<p>{% if exception %}{{ exception }}{% else %}CSRF verification failed. Request aborted.{% endif %}</p>
+<p>{% if exception %}{{ exception }}{% else %}You're not allowed to access this page.{% endif %}</p>
 {% endblock content %}{% endraw %}
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
@@ -5,5 +5,5 @@
 {% block content %}
 <h1>Forbidden (403)</h1>
 
-<p>CSRF verification failed. Request aborted.</p>
+<p>{% if exception %}{{ exception }}{% else %}CSRF verification failed. Request aborted.{% endif %}</p>
 {% endblock content %}{% endraw %}
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/404.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/404.html
@@ -5,5 +5,5 @@
 {% block content %}
 <h1>Page not found</h1>
 
-<p>This is not the page you were looking for.</p>
+<p>{% if exception %}{{ exception }}{% else %}This is not the page you were looking for.{% endif %}</p>
 {% endblock content %}{% endraw %}


browniebroke:
Is this used in production? If so, I'm -1. Showing it in debug mode might acceptable though.

Andrew-Chen-Wang:
Hm can you elaborate why not? Never tried in production as of yet. If you were to raise `PermissionDenied(_("You cannot view this"))`, that would show instead of some internal server error (which isn't a 500.html template anyways).

browniebroke:
Hum, I misread the cases were it's used, it's true that it's not in the 500 case.

In the example of permission denied you give, it would be fine. I was more thinking if one were to raise an exception with more raw details in it, they could probably end up exposing secrets. I see this as a similar problem as the settings context processor we used to have (#2670 #2614).

Maybe I'm being paranoid, I was truing to consider possible misuses...

Andrew-Chen-Wang:
Oh as in people would confuse the purpose of the message in Response Exceptions? It is definitely reasonable to believe. I know many places don't actually give a detailed response in the first place like if you do https://github.com/asdfhjgadskjhfgakjsdhfgaehahdlrkjghslkjd. But I originally opened this PR due to the 403 page's message being poor UX. I'm also not sure how many people would confuse the purpose of the inner message, and surveys aren't really a thing on GitHub.

I think the inner API dev of me got the better of this PR, so that's why I inputted the exception variable in the first place. I think your concern is valid, but changing the 403 page's message from CSRF verification failed to something like Permission Denied for better UX would be helpful since most people aren't going to understand what CSRF is.

I'm ok whether this PR is closed; I'm mostly just concerned about the 403 page.

Andrew-Chen-Wang:
Like here's something I currently do. Since I'm using regular templates instead of an SPA, I sometimes show some anchor tag that the user can't actually access (this was to decrease query counts since I didn't want to constantly check if a user had permission to open that anchor tag in the first place). By having specific messages, it's cleaner on the UX, but this is also not necessary if I had just done the permission checking in the first place:

Screen Shot 2021-01-24 at 11 02 08 AM

browniebroke:
I agree that 403 showing `CSRF validation` failed is wrong, I was myself confused by it when reviewing this. I agree it should be less specific, there are several possible reasons why one might see a 403 beyond CSRF error.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
```diff
@@ -5,5 +5,5 @@
 {% block content %}
 <h1>Forbidden (403)</h1>
 
-<p>CSRF verification failed. Request aborted.</p>
+<p>{% if exception %}{{ exception }}{% else %}CSRF verification failed. Request aborted.{% endif %}</p>
```
browniebroke, line 8: Let's fix this confusing message, 403 might happen in other cases than just CSRF error:

```suggestion
<p>{% if exception %}{{ exception }}{% else %}You're not allowed to access this page.{% endif %}</p>
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/404.html
```diff
@@ -5,5 +5,5 @@
 {% block content %}
 <h1>Page not found</h1>
 
-<p>This is not the page you were looking for.</p>
+<p>{% if exception %}{{ exception }}{% else %}This is not the page you were looking for.{% endif %}</p>
```
browniebroke, line 8: If your view raises `Http404()`, the default `This is not the page you were looking for` message is displayed, right?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
Andrew-Chen-Wang: Just double checking: if a user submits a form without the CSRF token, then the exception variable should evaluate true right?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/404.html
Andrew-Chen-Wang: Yup!

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403.html
Andrew-Chen-Wang: I believe it doesn't. I'm somewhat concerned that a dev wouldn't know if a form needed a CSRF token because this exception page has helped me in the past.

This is in the CSRFMiddleware:

```python
    def _reject(self, request, reason):
        response = _get_failure_view()(request, reason=reason)
        log_response(
            'Forbidden (%s): %s', reason, request.path,
            response=response,
            request=request,
            logger=logger,
        )
        return response
```
though via the `log_response` the dev can see in the console that they need a CSRF token.

But I think this is OK!

browniebroke (approved):

PR #2998, opened 2021-01-08
Omit first_name and last_name in User model

### LINKED ISSUE #1268: Custom user model still contains first_name and last_name
The custom user model provided by this template has a `name` field which works better with name patterns around the globe, but still inherits `first_name` and `last_name` from AbstractUser.

Is this intended behavior? If that's the case, it would be appropiate to document it.

Andrew-Chen-Wang:
## Description

Fixes #1268

Remove `first_name` and `last_name` from User model. This is just how I personally did it awhile ago.

Checklist:

## Rationale

We already have a name attribute, and the other attributes aren't used in any forms (as far as I can see in my current project).


browniebroke (approved):

browniebroke:
Thanks again @Andrew-Chen-Wang !

PR #3021, opened 2021-01-26
Using SuccessMessageMixin to send success message to django template

arnav13081994:
## Description

This is a super simple fix taking advantage of the Django built-in SuccessMessageMixin to pass the success message attribute to the template. 

Checklist:

## Rationale

 form_valid() at `users/views.py` currently is being overridden only to pass the success message to the template. Therefore, it's better and easier to simply use what Django already provides: [SuccessMessageMixin](https://docs.djangoproject.com/en/3.1/ref/contrib/messages/#django.contrib.messages.views.SuccessMessageMixin)

The code is also easier to understand.


browniebroke (approved):
LGTM, nice one

Issue #1810, opened 2018-09-19
Pin base Docker images up to bugfix releases


webyneter:
## Description

[//]: # (What's it you're proposing? How should it be implemented?)

`django` and other Docker Compose service images' base image versions **must be pinned to the corresponding bugfix releases** i.e. pinned down to the maximum attainable semver depth.

## Rationale

[//]: # (Why should this feature be implemented?)

I've just been racking my brains trying to figure out why one of my project's Docker Compose stack has been failing with weird erros while exactly the same git state was successfully reproduced on my fellow teammates' machine hassle-free. Eventually, I figured out the app's image was relying upon `python:3.6-alpine` as base image -- the one I already had pulled to my host about a year ago. By contrast, some other developers' machines had been wiped out clean beforehand, with all the Docker images pulled in fresh afterwards. Hence, their local setups had been using the now-latest `python:3.6.6-alpine3.8` while my pre-existing `python:3.6-alpine`-matching base image was 9 months old.

Anticipating the possible objection on pinning minors rather than patches due to automatic pulling of bugfix releases -- what I've stumbled upon here renders such concerns irrelevant; besides, there is a number of automatic dependency management solutions, like pyup or dependencies.io allowing to strategize upgrade policies in explicit rather than implicit manner.

browniebroke:
Sounds good. Another automation solution I came across worth considering is [renovate](https://github.com/apps/renovate).

webyneter:
@browniebroke thank you for the input, didn't know about that one -- definitely gotta check it out.

browniebroke:
Looks like Dependabot is getting integrated into Github: https://dependabot.com/blog/hello-github/

PR #3023, opened 2021-01-26
Update from Python 3.8 to Python 3.9

areski:
## Description

Upgrade from Python 3.8 to 3.9

Checklist:

## Rationale

Most of peoples starting a project with `cookiecutter-django` would want to use a more recent version of Python.


foarsitter:
Nice enhancement, trying it locally now. 

Is it a good idea to take #1810 into account?

areski:
@foarsitter django-phonenumber-field doesn't seem to be used here or I might have missed it. 
Could you link to where this might be?

No doubt about the usefulness of #1810, but this PR was not trying to address this.

foarsitter:
You are right about that `django-phonenumber-field` is not used in this project.

browniebroke:
Nice work @areski! Changes look good. We'll need to test a few prod deployment, as some libraries are not used in test.

For whoever is interested to see progress on this pull request, they can help by testing it with their favorite setup and reporting back...

> No doubt about the usefulness of #1810, but this PR was not trying to address this.

I agree that we should keep #1810 separate 👍

Thanks again for picking this up.

{{cookiecutter.project_slug}}/runtime.txt
```diff
@@ -1 +1 @@
-python-3.8.7
+python-3.9.1
```
Andrew-Chen-Wang, line 1: I believe we're on 3.9.2 now.

Andrew-Chen-Wang (commented):
@browniebroke Any updates on this PR?

browniebroke:
Thanks for the change! 🎉

areski:
Thanks @browniebroke

PR #3029, opened 2021-01-29
Refactored users.forms to make the code more readeable

arnav13081994:
## Description

The current code of [users.forms](https://github.com/pydanny/cookiecutter-django/compare/master...arnav13081994:update_forms?expand=1#diff-7d693034238506cd1674fce5df950dd1f9ab2b717756878640779aca3041789f) is pretty dense for no reason unless I am missing something. It seems the file was written to achieve 3 goals:

**1)** Extend the UserCreation and UserUpdate forms to use the custom User class.
**2)** To make sure every username is unique (and hence doesn't already exist in the db)
**3)** In case 2) fails, raise a custom validation error.

### Side-effects of the current code.

**1)**  One of the biggest issues with the current code base is that clean_username() is overridden without calling super().clean_username(). In the file, [config.settings.base](https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/base.py#L103), 

```
AUTHENTICATION_BACKENDS = [
    # Needed to login by username in Django admin, regardless of `allauth`
    "django.contrib.auth.backends.ModelBackend",

    # `allauth` specific authentication methods, such as login by e-mail
    "allauth.account.auth_backends.AuthenticationBackend",
]
```

which means that the [clean_username()](https://docs.djangoproject.com/en/3.0/ref/contrib/auth/#django.contrib.auth.backends.RemoteUserBackend.clean_username) method of ModelBackend authentication backend will automatically be called for all username related authentication, as mentioned in the [docs](https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#specifying-authentication-backends)

> When somebody calls django.contrib.auth.authenticate() – as described in How to log a user in – Django tries authenticating across all of its authentication backends. If the first authentication method fails, Django tries the second one, and so on, until all backends have been attempted.

And since the clean_username() method has been completely overridden and the super inheritance hierarchy broken, the functionality provided by that method will be lost. Not sure if this was deliberate or not.

**2)** Despite using ModelForms, username uniqueness check is done which is rather pointless since Django provides that functionality out of the box.

Checklist:

## Rationale

If the reasons mentioned above are indeed correct then my code would achieve the same and would be very understandable.


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_forms.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+from django.utils.translation import ugettext_lazy as _
```
Andrew-Chen-Wang, line 2:
```suggestion
from django.utils.translation import gettext_lazy as _
```
`ugettext_lazy` is deprecated right?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -12,20 +11,9 @@ class Meta(admin_forms.UserChangeForm.Meta):
...
-    def clean_username(self):
-        username = self.cleaned_data["username"]
-
-        try:
-            User.objects.get(username=username)
-        except User.DoesNotExist:
-            return username
-
-        raise ValidationError(self.error_messages["duplicate_username"])
+        error_messages = {
```
Andrew-Chen-Wang, line 17: Does this override the other error messages?

Andrew-Chen-Wang (changes requested):

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
arnav13081994: Yes @Andrew-Chen-Wang

browniebroke: That seems wrong, the base class has another error message that we should keep:

https://github.com/django/django/blob/d8dfff2ab0edf7a1ca5255eccf45c447b2f9d57e/django/contrib/auth/forms.py#L80-L82

Doesn't seem covered by tests btw, if we wrote one we'd get a `KeyError` here:

https://github.com/django/django/blob/d8dfff2ab0edf7a1ca5255eccf45c447b2f9d57e/django/contrib/auth/forms.py#L111

browniebroke (commented):
> One of the biggest issues with the current code base is that `clean_username()` is overridden without calling `super().clean_username()`.

There is no implementation for `clean_username` in [the base `UserCreationForm` form](https://github.com/django/django/blob/d8dfff2ab0edf7a1ca5255eccf45c447b2f9d57e/django/contrib/auth/forms.py#L75), it's not possible to call `super().clean_username()`.

> the clean_username() method of ModelBackend authentication backend will automatically be called for all username related authentication

Looking at [the source code for `auth.ModelBackend`](https://github.com/django/django/blob/5fcfe5361e5b8c9738b1ee4c1e9a6f293a7dda40/django/contrib/auth/backends.py#L31), I don't see any `clean_username()` either, the one you link to is in `RemoteUserBackend`.

> Despite using ModelForms, username uniqueness check is done which is rather pointless since Django provides that functionality out of the box.

You're right about this point though, `ModelForm` does the check for us.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
arnav13081994: @browniebroke My change only overrides the error message in case **unique username validation fails**. Other checks for passwords, emails etc (other fields in the User model would remain as it is). And that is why I override **[error_messages.username.unique](https://github.com/django/django/blob/d8dfff2ab0/django/contrib/auth/models.py#L336)**:

```
        error_messages = {
            "username": {"unique": _("This username has already been taken.")}
        }

```
In fact, in case one adds more checks for the username field (for example every username should start with "@" etc) they should simply get appended to the list of error messages if any. And that is the use-case where overriding the **clean_username() method**  would make more sense, in my opinion.

You are 100% right about the **super().clean_username()**, I am not even sure how I confused RemoteUserBackend with ModelBackend. My bad.

To Summarise the only thing this code would do will change the error message that Django will output by default 
```A user with that username already exists.``` to ```This username has already been taken.``` when unique username validation fails. Nothing else would get affected.

browniebroke: Ok I misread the code 🤦  

I thought you were setting this `error_messages` class attribute on the form directly, but I just realised this is on the form nested `Meta` class (as [per documentation](https://docs.djangoproject.com/en/3.0/topics/forms/modelforms/#overriding-the-default-fields)).

This is fine, it just overrides the error message for the specific validation rule, as we want.

arnav13081994:
@browniebroke @Andrew-Chen-Wang I pushed another commit with the following updates:

1) Added module, class, and method docstrings
2) Renamed ```test_clean_username``` to ``` test_username_validation_error_msg ```.  It's very verbose on purpose.
3) Modified ``` test_username_validation_error_msg ``` to only test the required attributes:
    - Only 1 error is raised by the UserCreation Form
    - The desired error message is raised by the UserCreation Form
    - A new user with an existing username cannot be added.
    
   The other checks are unnecessary since Django username model field does that for us.

Please let me know if you have any comments on how to improve this.

P.S: I will raise a PR soon that adds docstrings to most modules, class, and methods. That would make the project and its structure much more understandable as well.

browniebroke (approved):

browniebroke:
Thanks @arnav13081994 !

PR #3030, opened 2021-01-30
Updated test_user_count() test to also make it work in case db already has users.

arnav13081994:
## Description

Updated the test, **test_user_count()**, to check for the difference in the number of users before the test is run and after the users are created. This will handle the scenario of users being present in the local db.

Checklist:

## Rationale

The same database is used for both local development and for running tests locally. In such a scenario, it will almost certainly be the case that the local db has 1 or more users before the test is run (Example an admin user). In such a scenario, the test, **test_user_count()**, will fail since it assumes that the db has 0 users when it is run.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_tasks.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_tasks.py
@@ -9,8 +9,16 @@ pytestmark = pytest.mark.django_db
 
 def test_user_count(settings):
     """A basic test to execute the get_users_count Celery task."""
-    UserFactory.create_batch(3)
     settings.CELERY_TASK_ALWAYS_EAGER = True
-    task_result = get_users_count.delay()
-    assert isinstance(task_result, EagerResult)
-    assert task_result.result == 3
+
+    # Get all existing users in the DB
+    current_users = get_users_count.delay()
+    assert isinstance(current_users, EagerResult)
+
+    # Create and add 3 more users to the DB.
+    UserFactory.create_batch(3)
+
+    # Get number of newly added users in the DB
+    task_result = get_users_count.delay().result - current_users.result
+
+    assert task_result == 3
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_models.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_models.py
@@ -1,8 +1,8 @@
 import pytest
-
-from {{ cookiecutter.project_slug }}.users.models import User
+from django.contrib.auth import get_user_model
 
 pytestmark = pytest.mark.django_db
+User = get_user_model()
 
 
 def test_user_get_absolute_url(user: User):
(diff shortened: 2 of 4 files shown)


browniebroke:
> The same database is used for both local development and for running tests locally

Can you expand a bit on this? Django creates a test DB to run test. It uses the same server as the one used for local development, yes, but it's under a different database name in that server:

> The default test database names are created by prepending `test_` to the value of each `NAME` in `DATABASES`.

https://docs.djangoproject.com/en/3.1/topics/testing/overview/#the-test-database

Therefore it should be completely separated from the DB you use to browser the site when starting it locally with `runserver`.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_tasks.py
```diff
@@ -9,8 +9,16 @@
...
     """A basic test to execute the get_users_count Celery task."""
-    UserFactory.create_batch(3)
     settings.CELERY_TASK_ALWAYS_EAGER = True
-    task_result = get_users_count.delay()
-    assert isinstance(task_result, EagerResult)
-    assert task_result.result == 3
+
+    # Get all existing users in the DB
+    current_users = get_users_count.delay()
+    assert isinstance(current_users, EagerResult)
```
browniebroke, line 16: Any reason to use the task to get the previous cont instead of calling directly `User.objects.count()`?

Might be clearer if it was:

```python
previous_count = User.objects.count()
UserFactory.create_batch(3)
... (same as old code)
assert isinstance(task_result, EagerResult)
assert task_result.result == previous_count + 3
```

Andrew-Chen-Wang:
Yea I can't understand this change to the test case as well. The database is separate from your local development or production server database called `test_REAL_DB_NAME`. The only time I can remember this being necessary is if you use custom SQL without using the Django ORM entirely AND there was an error during teardown rather than during the test.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_tasks.py
arnav13081994: @browniebroke @Andrew-Chen-Wang Thank you for your comments. I certainly was confused. Reading the [link](https://docs.djangoproject.com/en/3.1/topics/testing/overview/#the-test-database) regarding how Django creates a test DB cleared it up.

> Any migrations will also be applied in order to keep it up to date.


The tests on my local dev were failing and I couldn't explain them and came to this incorrect conclusion that my DB is being re-used when in fact it was due to me creating a superuser using migrations! 

Thanks for the help. 

@browniebroke Regarding your code review comment, I just wanted to use celery to get the number of users. No other reason


This PR seems unnecessary now although if someone like me ends up using migrations to add users then they will get the same issue. Let me know if I should just close this issue.

arnav13081994:
@Andrew-Chen-Wang Thank you for your comment. The issue was that I was using migrations to create an admin user and migrations were being run for my **"test_db"** as well which was causing this particular test on my local dev env to fail.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_tasks.py
browniebroke: Using Celery to get the count of users before and after seems to overcomplicate the test. I'd rather keep it simple.

The point of this test is just to have something testing that Celery is configured correctly. It's mainly for the sake of Cookiecutter Django, more than the projects you generate.

I'll leave it up to you if you want to simplify it or close the PR.

Thanks.

PR #3038, opened 2021-02-02
Fix deprecated warning in middleware tests

arnav13081994:
## Description

One can no longer pass ``` None ``` to the middleware factory object since now they are nested functions (or closures) that take a get_response callable.  From the [documentation](https://docs.djangoproject.com/en/3.1/topics/http/middleware/#writing-your-own-middleware),

> A middleware factory is a callable that takes a get_response callable and returns a middleware.

In order to get around this issue, I created a ```dummy_get_response``` callable that takes a request as an input and return ```None``` as the response so still maintains the behaviour we want but avoid passing ```None``` into the [SessionMiddleware factory](https://docs.djangoproject.com/en/3.1/ref/middleware/#django.contrib.sessions.middleware.SessionMiddleware), for example.

Checklist:

## Rationale

Passing None for the middleware get_response argument is deprecated.
    MessageMiddleware().process_request(request)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -27,6 +28,9 @@ class TestUserUpdateView:
         https://github.com/pytest-dev/pytest-django/pull/258
     """
 
+    def dummy_get_response(self, request: HttpRequest):
+        return None
+
```
Andrew-Chen-Wang, line 33: Not gonna be DRY when people start making more views. Instead, create something in conf.py that allows devs to use parameters. So:

```python
def session_middleware():
    def dummy_get_response(request: HttpRequest):
        return
    return SessionMiddleware(dummy_get_response).process_request

class TestUserUpdateView
    def test_get_success_url(self, user: User, rf: RequestFactory, session_middleware):
        ...
        session_middleware(request)
        ...
```
And the same thing with `MessageMiddleware`

Andrew-Chen-Wang (changes requested):
DRYer way:

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
browniebroke: Are you suggesting to make `session_middleware` a `pytest` fixture? I was going to suggest make `get_response` a fixture instead. I think it would be more reusable to test any middleware, even if they don't need a session. Also, it could return a response instead of `None`.

I have this in `conftest.py`:

```python
@pytest.fixture(name='get_response')
def get_response_fixture():
	def get_response(_request):
    	"""A callable to pass to a middleware we test."""
    	return HttpResponse()
    return get_response
```

Andrew-Chen-Wang: Ah yea a fixture, whoops. And yes conf**test**.py

I like that being separate, but for the most part there should still be a fixture for processing a middleware. So maybe:

```python
@pytest.fixture
def get_response():
    # Renamed to prepend with _ to not interfere with other functions
    # I'm not sure if it'll make it private though, which would be an issue
    # (would it be private since there are two underscores?)
    def __get_response(_request):
    	"""A callable to pass to a middleware we test."""
    	return HttpResponse()
    return __get_response

# This could be renamed in case someone DID want to use the other
# methods of SessionMiddleware, but I kinda doubt it.
@pytest.fixture
def session_middleware(get_response):
    return SessionMiddleware(dummy_get_response).process_request

def test_blah(session_middleware):
    session_middleware(request)
```
i.e. a combination since one wouldn't want to keep importing `SessionMiddleware` or whatever middleware everywhere, especially as a project grows.

Andrew-Chen-Wang: In other words: 1) definitely agree but also 2) we should still add the `message_middleware` and `session_middleware` fixture so (newer?) devs at least know that they can just use these fixtures instead. It's a nice introduction to pytest too.

Andrew-Chen-Wang: Oh wait: `messages.get_messages(request)` you would still need an import to `django.contrib.messages` huh... I just dislike having to remember imports and also dislike importing in general :P

arnav13081994: @Andrew-Chen-Wang @browniebroke Thank you for your insightful comments.

I borrowed from both your comments and propose adding the following to conftest.py


```conftest.py```:

```
@pytest.fixture(name="get_response")
def get_response_fixture():
    """
    Returns the get_response_callable
    """

    def get_response_callable(*args, **kwargs):
        """A callable to pass to a middleware we test that returns
        an HttpResponse object with the status
        code and content passed to it for the given HttpRequest."""

        response = HttpResponse(*args, **kwargs)

        def _get_response(_request: HttpRequest) -> HttpResponse:
            return response

        return _get_response

    return get_response_callable

```

```test_views.py```:

```
    def test_form_valid(
        self, user: User, rf: RequestFactory, get_response
    ):  
       ....
        # Add the session/message middleware to the request
        SessionMiddleware(get_response()).process_request(request)
        MessageMiddleware(get_response()).process_request(request)
        .....
```


This way one can re-use the same get_response callable across different middlewares and pass in the status_codes they need. And they can also override the default reason_phrase etc as well.

For Example:

```
    def test_user_not_authorised(
        self, user: User, rf: RequestFactory, get_response
    ):  
      """Return Forbidden Response.""" 
        ...
        SomeMiddleware(get_response(status_code = 403)).process_request(request)
        ....
```

Comment (browniebroke) [reply {{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py:33]: Not sure what to think about the 3rd level `_get_response`. I get the idea to provide different status codes, but a fixture that returns a callabale that returns a function might be a bit too complicated? What if you need a different response class e.g. `JsonResponse`?

Not sure if we need to have that much customisation when testing middlewares...

Comment (Andrew-Chen-Wang) [reply {{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py:33]: > What if you need a different response class e.g. JsonResponse? Not sure if we need to have that much customisation when testing middlewares...

I agree. Perhaps just having that `get_response` fixture is good enough. I just like the idea of minimizing the number of times you have to import a core Django functionality. That or we could also add a class attribute to RequestFactory. Like:

```python
@pytest.fixture
def session_middleware():
    return SessionMiddleware

# OR:

@pytest.fixture
def rf(rf):
   # Overriding request factory
   rf.session_middleware = SessionMiddleware
   return rf
```

browniebroke: I don't see a problem with importing the `SessionMiddleware`. It's not something you need 100% of the time when testing middlewares, and when you do, I think it's nice to make this dependency explicit in tests.

arnav13081994: @browniebroke I was going for a `middleware fixture factory` pattern that could be modified (in the future) to include other checks and attributes etc depending on the specific middleware one is using. I agree that for the 2 middlewares we are talking about here returning a nested closure is overkill.

Andrew-Chen-Wang: Would just like to add-on to the idea of the session middleware fixture. For me, the biggest issue is still those two imports. I just like being lazy, but the MessageMiddleware requires the SessionMiddleware be processed first. Then return `django.contrib.messages` for the user to use. Again, that's personal preference, but getting this merged would be great!

browniebroke (approved):
This is fixing the problem at hand, so I'm going to merge it as is.

PR #3041, opened 2021-02-04
Removed Redundant test_case_sensitivity() and made test_not_authenticated() get the LOGIN_URL dynamically.

arnav13081994:
## Description

1) Remove the test, ```test_case_sensitivity()```, which checks if the User model field, username, is case sensitive. It is redundant because **Django User model username field is case-sensitive by default**.

2) The Login URL in the test, test_not_authenticated(), should be fetched dynamically. Currently, it is hard-coded to ```/accounts/login/``` which would break if ```settings.LOGIN_URL``` is modified in ```config.settings.base``` file. 

Checklist:

## Rationale

1) This is just redundant and doesn't provide any advantage and hence should be removed.
2) This change would make the test more resilient to different user-specific configurations as wel.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -1,10 +1,12 @@
 import pytest
+from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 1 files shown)


browniebroke (approved):
Seems sensible to me. The case sensitivity test was added after someone raised #1020, but I agree, it's redundant with Django's behaviour.

PR #2203, opened 2019-08-22
Add context processor for settings variables

danihodovic:
This allows for usage of settings variables in templates, e.g

```
{% if settings.USE_ANALYTICS %}
<!-- Google Tag Manager (noscript) -->
<noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ settings.GOOGLE_TAG_MANAGER_ID }}" style="display:none;visibility:hidden"></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
{% endif %}
```

[//]: # (Thank you for helping us out: your efforts mean great deal to the project and the community as a whole!)

[//]: # (Before you proceed:)

[//]: # (1. Make sure to add yourself to `CONTRIBUTORS.rst` through this PR provided you're contributing here for the first time)
[//]: # (2. Don't forget to update the `docs/` presuming others would benefit from a concise description of whatever that you're proposing)

## Description

[//]: # (What's it you're proposing?)

## Rationale

[//]: # (Why does the project need that?)

## Use case(s) / visualization(s)

[//]: # ("Better to see something once than to hear about it a thousand times.")

PR #3042, opened 2021-02-04
Removed unnecessary custom context processor exposing the DEBUG Template Context Variable

arnav13081994:
## Description

Remove the custom context processor, ``` {{ cookiecutter.project_slug }}.utils.context_processors.settings_context ```,  that was exposing a Template Context Variable, DEBUG. This is because the first context processor in the list, ```settings.TEMPLATES.context_processor``` enabled [django.template.context_processors.debug](https://docs.djangoproject.com/en/3.1/ref/templates/api/#django-template-context-processors-debug), which exposes the same variable, debug.

Checklist:

## Rationale

This is just redundant and doesn't provide any advantage and hence should be removed. The code will also become easier to understand.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -198,7 +198,6 @@ TEMPLATES = [
                 "django.template.context_processors.static",
                 "django.template.context_processors.tz",
                 "django.contrib.messages.context_processors.messages",
-                "{{ cookiecutter.project_slug }}.utils.context_processors.settings_context",
             ],
         },
     }
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/utils/context_processors.py
@@ -1,8 +1,0 @@
-from django.conf import settings
-
-
-def settings_context(_request):
-    """Settings available by default to the templates context."""
-    # Note: we intentionally do NOT expose the entire settings
-    # to prevent accidental leaking of sensitive information
-    return {"DEBUG": settings.DEBUG}


browniebroke:
This context processor is meant to be the place where one would expose settings they need in templates (#2203). In the stock installation, it just exposes DEBUG since exposing all settings was deemed dangerous (#2694), which is true makes it redundant. Maybe we should expose another setting than DEBUG, which would actually be useful? I don't have anything in mind, though...

On the other hand, I appreciate that it might not be used in all projects, and it's quite a simple thing to add when needed, having it might not be so valuable after all 🤔

arnav13081994:
Yes agreed. I also thought it was meant to be a way for others to get into it

arnav13081994:
Personally, I also use[ Sentry for JS](https://docs.sentry.io/platforms/javascript/) and send ```SENTRY_DSN``` and other Sentry env vars to ```templates/base.html``` using context_processors. Perhaps we should also add SENTRY for JS to this project and then update context_processors to send the required info to the templates?

Andrew-Chen-Wang:
I think having it as an option for people won't hurt, even if they never use it. Lots of features of Django are never really known until needed OR if someone has seen it before. In other words, we present a new option for a dev to remember for the future when needed. Good creativity wise.

browniebroke:
> Personally, I also use Sentry for JS and send SENTRY_DSN and other Sentry env vars to templates/base.html using context_processors.

When I add Sentry for JS, I usually setup 2 separate projects for JS and Django, since Sentry asks for the language when setting up the project. Isn't the Sentry dashboard messy when both are using the same DSN? I really never tried to mix them up...

arnav13081994:
I actually never thought of using it as a separate project! It does get a bit messy but usually, it's easy to filter JS errors from other errors.

browniebroke:
OK, I've come around that PR again, and I'm now +1 on removing this. Since it's not used in the stock installation, it's adding some unneeded code to projects not using it while not providing a lot of value to people who do use it, as creating a context processor is not very complex thing to create (it's less that 10 lines of code after all).

Anyone object to this change? If so, now is the time to speak up 😄

PR #3055, opened 2021-02-16
Remove Email from User API

Andrew-Chen-Wang:
## Description

Remove a user's email from User API

Checklist:

## Rationale

Privacy concern. Unless a website really wants you to know a user's email, then don't show it publicly. It's not really public anywhere else on the Django templates so I wouldn't want that leaked.


browniebroke (approved):
Good catch 👍

PR #3064, opened 2021-02-21
refactor: remove default cache settings in test.py

danihodovic:
It's LocMemCache by default, there is no need to specify this in test.py
```python
django.core.cache.backends.locmem.LocMemCache
```

https://docs.djangoproject.com/en/dev/ref/settings/#caches


browniebroke (approved):
LGTM

PR #3066, opened 2021-02-23
Lint with pre-commit on CI with Github actions

arnav13081994:
## Description

Updated GitHub `ci.yml` file to run pre-commit hooks on all files instead of manually installing and running flake8.

Checklist:

## Rationale
A more DRY way of running code-quality tools would be to add them all in the `.pre-commit-config.yaml` file and then invoke pre-commit in CICD. This has a couple of benefits:

```
1) 1 central place for all code quality tools
2) No discrepancy on code quality tools that run in CICD versus on one's local dev env.
```

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ on:
 
 
 jobs:
-  flake8:
+  linter:
     runs-on: ubuntu-latest
     steps:
 
@@ -28,13 +28,13 @@ jobs:
         with:
           python-version: 3.8
 
-      - name: Install flake8
+      - name: Install flake8, flake8-isort, and black
         run: |
           python -m pip install --upgrade pip
-          pip install flake8
+          pip install flake8 flake8-isort black
 
-      - name: Lint with flake8
-        run: flake8
+      - name: Install and Run Pre-commit 
+        uses: pre-commit/action@v2.0.0
 
 # With no caching at all the entire ci process takes 4m 30s to complete!
   pytest:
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/.github/workflows/ci.yml
```diff
@@ -28,13 +28,16 @@ jobs:
         with:
           python-version: 3.8
 
-      - name: Install flake8
+      - name: Install flake8, flake8-isort, and black
         run: |
           python -m pip install --upgrade pip
-          pip install flake8
+          pip install flake8 flake8-isort black
```
browniebroke, line 34: Do we need this step now? I think the action below won't be using these packages, they are installed again

browniebroke (commented):
Good idea. Should we also update the gitlab file as well?

{{cookiecutter.project_slug}}/.github/workflows/ci.yml
arnav13081994: I was actually planning on also adding local repo pre-commit hooks for `mypy` and `pylint` and they would require all project dependencies to be loaded since they both do some dynamic checking and imports etc.

Should I just add them to this PR? Or raise another one with those changes?

browniebroke:
> I personally find pylint and mypy to be very useful and very thorough albeit a bit annoying at times too.

I'm -1 on pylint, but we can discuss in a separate PR/issue if you feel like it worth exploring... We can definitely add mypy though, we are using it in CI, so should be in pre-commits as well. Doesn't have to be done in this PR.

> travis.yml and gitlab-ci.yml would also need to be updated but I haven't ever used either so I'm a bit apprehensive

That's fine, we can do it in a follow up PR

> Also one should set up caching for the pre-commit hooks

Sure, why not. Another potential incremental improvement.

browniebroke (approved):
I'm ok with the PR as it stands, and to do further improvements later on.

browniebroke:
Thanks 🎉

PR #3070, opened 2021-02-26
Updated test_urls.py and views.py to re-use User.get_absolute_url()

arnav13081994:
## Description

Re-use ` User.get_absolute_url()` method to fetch the detail view URL instead of doing `reverse("users:detail", kwargs={"username": user.username})`
Checklist:

## Rationale

This would make the project "DRYer"

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_urls.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_urls.py
@@ -7,10 +7,7 @@ pytestmark = pytest.mark.django_db
 
 
 def test_detail(user: User):
-    assert (
-        reverse("users:detail", kwargs={"username": user.username})
-        == f"/users/{user.username}/"
-    )
+    assert user.get_absolute_url() == f"/users/{user.username}/"
     assert resolve(f"/users/{user.username}/").view_name == "users:detail"
 
 
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_urls.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_urls.py
@@ -7,7 +7,10 @@ pytestmark = pytest.mark.django_db
 
 
 def test_detail(user: User):
-    assert user.get_absolute_url() == f"/users/{user.username}/"
+    assert (
+        reverse("users:detail", kwargs={"username": user.username})
+        == f"/users/{user.username}/"
+    )
     assert resolve(f"/users/{user.username}/").view_name == "users:detail"
 
 
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/views.py
@@ -25,7 +25,7 @@ class UserUpdateView(LoginRequiredMixin, SuccessMessageMixin, UpdateView):
     success_message = _("Information successfully updated")
 
     def get_success_url(self):
-        return reverse("users:detail", kwargs={"username": self.request.user.username})
+        return self.request.user.get_absolute_url()  # type: ignore [union-attr]
 
     def get_object(self):
         return self.request.user


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_urls.py
```diff
@@ -7,10 +7,7 @@
 
 
 def test_detail(user: User):
-    assert (
-        reverse("users:detail", kwargs={"username": user.username})
-        == f"/users/{user.username}/"
-    )
+    assert user.get_absolute_url() == f"/users/{user.username}/"
```
browniebroke, line 10: Not sure about this... The spirit of this test is to make sure the URL reverses properly. Now we're introducing a dependency on `get_absolute_url()`

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/views.py
```diff
@@ -25,7 +25,7 @@ class UserUpdateView(LoginRequiredMixin, SuccessMessageMixin, UpdateView):
     success_message = _("Information successfully updated")
 
     def get_success_url(self):
-        return reverse("users:detail", kwargs={"username": self.request.user.username})
+        return self.request.user.get_absolute_url()  # type: ignore [union-attr]
```
browniebroke, line 28: Can we avoid this `# type: ignore` by any chance? I would expect it to work if we add `-> str` annotation to the function definition at the line above. Is my expectation wrong?

arnav13081994: Unfortunately, there is no easy way to avoid this. The issue is that mypy doesn't know that the user is authenticated, which is something we know because `LoginRequiredMixin` runs first. One can do the following

> ```
> from django.http import HttpRequest
> from my_user_app.models import MyUser
> 
> class AuthenticatedHttpRequest(HttpRequest):
>     user: MyUser
> 
> ```
> And then use AuthenticatedHttpRequest instead of the standard HttpRequest for when you know that the user is authenticated. 

 as mentioned in the docs but that would make this view rather messy, in my opinion. Adding that `# type ...` seems like an easy fix for now.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_urls.py
arnav13081994: Yes, I understand that but the `detail view` would always use the `get_absolute_url()` method throughout the project to get the detail view of the concerned object so wouldn't it make more sense to directly invoke that method and test it? What if someone adds a `userPassesTest` Mixin etc to not allow everyone to get the detail view? In that scenario, one would either need to update this test or add another test for this.

browniebroke: How about this one? It's not using `get_absolute_url()` is it?

https://github.com/pydanny/cookiecutter-django/blob/9ebfd5102c008debaf8721f90262a6d741c26898/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/templates/base.html#L81

> What if someone adds a userPassesTest Mixin etc to not allow everyone to get the detail view? In that scenario, one would either need to update this test or add another test for this.

I'm not sure I understand what you mean. Requesting the page is a different thing than reversing the URL. I think this type of mixin is added to the view. If not satisfied, it prevents view access, it doesn't prevent reversing the URL location.

Again, I think this is testing detail _URL route_, not the detail _view_.

arnav13081994: I see what you mean now. I was definitely testing the view with this change than the URL. I will revert the change in `test_urls.py`
Regarding the line, I would change it to:
`<a class="nav-link" href="request.user.get_absolute_url">{% trans "My Profile" %}</a>`

because I think one should reduce as many moving parts as possible. I know doing this might be a bit unconventional as I have also never seen anyone use `get_absolute_url()` to get the `detail-view-url` as opposed to using the usual URL template tag and supplying the pk and the URL name. 

Out of curiosity, is there any potential issue I am missing? Because it seems to me that with this change, the template code would be a bit cleaner, way more stable especially if the `slug_field` were ever to change or some permission mixin was added to the `get_absolute_url()` method etc. Those cases do seem not super likely though so that could be a reason for this.

browniebroke: Regarding the template tag vs `get_absolute_url` in the template... I really don't mind either way. We use the `url` tag several other times in this file, seems consistent as it is. 

If things were to change, one would have to find usages and update them, yes, but I don't see that as a compelling reason to change the template.

arnav13081994: Ok makes sense. Thank you for your reply.

PR #3071, opened 2021-02-26
Refactored test_views.py to not use unnecessary fake_url and made tes…

arnav13081994:
## Description

Refactor `test_views.py` to remove using `fake-url` since that is unnecessary anyway. And to properly test views especially when `POST` requests are concerned.

Checklist:

## Rationale

In my opinion, ``test_views.py`` is not properly written. `rf.post` exists for sending POST requests and much of the same functionality is trying to be achieved using `fake-url` with what I think is a hacky way to write tests.


arnav13081994:
Closing this for https://github.com/pydanny/cookiecutter-django/pull/3151

PR #3072, opened 2021-02-27
Added all open source licenses (Fixes #2941)

### LINKED ISSUE #2941: add AGPL
Please add the AGPL license to the list of available licenses! Thanks

Andrew-Chen-Wang:
## Description

Use licenses that we see on GitHub when we create a new repo. Also included is automatic updating of the licenses from the aforementioned license repository created by GitHub. The cookiecutter.json file is also updated.

* Uses open source licenses (including the Jekyll format) from https://github.com/github/choosealicense.com/tree/gh-pages/_licenses
* Added license section in README.rst because some licenses want it to be explicit
* cookiecutter.json shows all licenses (including "Not open source") in lexicographical order except for "Not open source"

Checklist:

Docs haven't been updated because honestly... I don't know if we want to show ALL of those licenses...

## Rationale

Fixes #2941

Lots of open source licenses... too many to choose.

Makes it easier to maintain since we can just use GitHub actions to constantly update the files if necessary.


Andrew-Chen-Wang:
Oooo that's an... interesting error to deal with...

Andrew-Chen-Wang:
Something that can make sure commits decrease in the automatic workflow is to check if the license file already exists since licenses don't change due to their copyright reasons. That means we should only commit if there's a difference in cookiecutter.json (i.e. any new license).

Additionally, it might be better to also include the MIT license, Apache, etc. near the front like "Not open source" license is in cookiecutter.json.

Andrew-Chen-Wang:
Just realized we should instead make a pull request instead of committing directly in case of a new file since there would be square brackets.

Andrew-Chen-Wang (commented):
Perhaps I missed a couple. I think that should be it from my initial generation (this was before I actually programmed the autoupdater)

hooks/post_gen_project.py
```diff
@@ -324,6 +312,34 @@ def remove_storages_module():
     os.remove(os.path.join("{{cookiecutter.project_slug}}", "utils", "storages.py"))
 
 
+def handle_licenses():
```
browniebroke, line 315: Is this to remove the Jekyll front matter? I'm wondering if wouldn't make more sense to do this work in the `update_licenses.py` scripts maybe? 

We can afford to use more modern Python there, we control the Python version better.

scripts/update_licenses.py
```diff
@@ -0,0 +1,59 @@
...
+        content = compiled.sub(replace, content)
+    return content
+
+
+def update_cookiecutter(titles: list):
+    with open("cookiecutter.json") as f:
+        data = json.load(f)
+    data["open_source_license"] = titles
+    with open("cookiecutter.json", "wt") as f:
+        json.dump(data, f, indent=2)
```
browniebroke, line 55: You can do away without the context manager with Pathlib's `read_text()` and `write_text()` APIs. Not a big deal, I find them a bit higher level.

scripts/update_licenses.py
```diff
@@ -0,0 +1,59 @@
...
+    ]
+    titles = [x for x in sorted(titles) if x not in front_options]
+    update_cookiecutter(front_options + titles)
+
+
+year = (re.compile(r"\[year]"), "{% now 'utc', '%Y' %}")
+email = (re.compile(r"\[email]"), "{{ cookiecutter.email }}")
+fullname = (re.compile(r"\[fullname]"), "{{ cookiecutter.author_name }}")
+project = (re.compile(r"\[project]"), "{{ cookiecutter.project_name }}")
+projecturl = (re.compile(r"\[projecturl]"), "{{ cookiecutter.domain_name }}")
```
browniebroke, line 41: Do you think we could add a test to make sure we don't miss any new placeholder? Seems a bit tricky to make the difference between a placeholder and a genuine `[...]` though. 🤔

cookiecutter.json
```diff
@@ -6,13 +6,7 @@
...
...o What The F*ck You Want To Public License\"", "Academic Free License v3.0", "Artistic License 2.0", "BSD 2-Clause \"Simplified\" License", "BSD 3-Clause Clear License", "BSD 4-Clause \"Original\" or \"Old\" License", "BSD Zero Clause License", "Boost Software License 1.0", "CeCILL Free Software License Agreement v2.1", "Creative Commons Attribution 4.0 International", "Creative Commons Attribution Share Alike 4.0 International", "Creative Commons Zero v1.0 Universal", "Eclipse Public License 1.0", "Eclipse Public License 2.0", "Educational Community License v2.0", "European Union Public License 1.1", "European Union Public License 1.2", "GNU Affero General Public License v3.0", "GNU General Public License v2.0", "GNU Lesser General Public License v2.1", "GNU Lesser General Public License v3.0", "ISC License", "LaTeX Project Public License v1.3c", "Microsoft Public License", "Microsoft Reciprocal License", "Mozilla Public License 2.0", "ODC Open Database License v1.0", "Open Software License 3.0", "PostgreSQL License", "SIL Open Font License 1.1", "The Unlicense", "Universal Permissive License v1.0", "University of Illinois/NCSA Open Source License", "Vim License", "zlib License"],
```
browniebroke, line 9: I see the script has now indenting this properly, although this is not yet reflected here.

.github/workflows/update-licenses.yml
```diff
@@ -0,0 +1,34 @@
+name: Update Licenses
+
+on:
+  # Every day at 2am
+  schedule:
+    - cron: "0 2 * * *"
```
browniebroke, line 6: We might want to run this at another time that the chnagelog update.

browniebroke (commented):
> Just realized we should instead make a pull request instead of committing directly in case of a new file since there would be square brackets.

We can reuse what we do to keep pre-commit hooks config up to date:

https://github.com/pydanny/cookiecutter-django/blob/master/.github/workflows/pre-commit-autoupdate.yml

It doesn't do anything in case of no changes.

hooks/post_gen_project.py
Andrew-Chen-Wang: Sorta, we need the Jekyll formatting *at first*. Each file's name is not the same as the title. So in order to find the correct file, we loop through each file until we find one that has the Jekyll title of the exact license name from cookiecutter.json.

We could improve on this inefficient method though by having `update_licenses.py` update this Python file to have a dictionary that maps titles (the dict keys) to their files (the dict values). That way we can remove the need for the Jekyll headers.

browniebroke: Ah, I see...

scripts/update_licenses.py
Andrew-Chen-Wang: That's why I wanted the pull request initially... In the Jekyll headers for all these files, there are directions that specify what to replace though! The directions say something like "replace [year] with etc." if I'm not mistaken. So after replacing everything, if we still match a regex of `\[.*?\]` (which matches to anything with square brackets including if there are multiple square brackets like `[asdf] [asf]` will be two matches).

Side note: I think we don't really need to care about Appendix directions (like that found in Apache 2.0 where they say to add the boilerplate code with year and name. With us putting it in already, I think the developer will understand. Additionally, license years don't have to be updated so there's that...)

cookiecutter.json
Andrew-Chen-Wang: I'm slightly confused what you mean here? Are you saying that the GitHub workflow may not get the indentation correctly? Hopefully when the argument "indent=2" or 4 (can't remember) from `json.dump` works properly (preserving ordering as well). But I can't remember during my test run if **all** the arrays became... well aren't vertically spaced out like `postgresql_version`

browniebroke: With indent=2, I would expect this part to wrap vertically, a bit like it used to, but it doesn't. I was trying to ask if it's because you generated this with an earlier version where indent was missing or if my expectation is just wrong.

Andrew-Chen-Wang:
@browniebroke I'll be creating a new PR. Looks like I completely missed their (new?) API that includes the licenses: https://docs.github.com/en/rest/reference/licenses#get-a-license

Andrew-Chen-Wang:
#3982 would resolve this

Issue #3091, opened 2021-03-11
Replace Gulp with Webpack


michael-yin:
## Use Webpack as static build tool

Now this project is still using Gulp as frontend build tool, since Webpack is very popoular and the ecosystem is very mature. I think it is better to use Webpack as frontend build tool.

## Rationale

1. `Webpack` is the most popular bundle solution in the frontend community today, it has received 50k stars on Github.
1. It has a great ecosystem, many plugins, loaders. If we search `webpack` on npmjs.com, we can get 20k resulst.
1. If we do not need `React`, `Vue`, we can still use Webpack to help us compile `ES6`, `SCSS` and do many other things (Many people do not know that!)
1. With a proper config, Webpack can save time and let us build modern web application in quick way.

Actually, other community did quicker on this:

1. Laraval has [laravel-mix](https://github.com/JeffreyWay/laravel-mix)
1. Rails has [webpacker](https://github.com/rails/webpacker)

They all simplified the Webpack config job and make developer can quickly write ES6 and SCSS.

## How I solve the problem

I already created a project [python-webpack-boilerplate](https://github.com/AccordBox/python-webpack-boilerplate)

It contains:

1. A `Webpack project template` which has good structure, easy to use and customize. User can use it even they have no idea how to config Webpack.
1. Template tags which can help load Webpack bundle file in the templates, it is smart to `auto load` dependency files.

If some users want to use React or Vue, the `python-webpack-boilerplate` also supports them.

## Features

- **Supports Django and Flask** (will support more framework in the future)
- Automatic multiple entry points
- Automatic code splitting
- Hot Module Replacement (HMR)
- Easy to config and customize
- ES6 Support via [babel](https://babeljs.io/) (v7)
- JavaScript Linting via [eslint-loader](https://github.com/MoOx/eslint-loader)
- SCSS Support via [sass-loader](https://github.com/jtangelder/sass-loader)
- Autoprefixing of browserspecific CSS rules via [postcss](https://postcss.org/) and [autoprefixer](https://github.com/postcss/autoprefixer)
- Style Linting via [stylelint](https://stylelint.io/)

## Benefits

`python-webpack-boilerplate` is to only provide frontend solution, so it can be used with both `Django` and `Flask` projects. 

So Django developers and Flask developers can both contribute and help improve this project and make it better.

## Project link

https://github.com/AccordBox/python-webpack-boilerplate

[...]

PR #3106, opened 2021-03-26
Replace Gulp with Webpack

michael-yin:
Please check #3091 to see why this PR is created.

### Basic workflow

1. In `hooks/post_gen_project`, I added a function to cookiecutter `frontend` project from the [python-webpack-boilerplate
](https://github.com/AccordBox/python-webpack-boilerplate/)
1. If people have any questions about the frontend, Webpack, they can create issues there.
1. I will also create PR for `cookiecutter-flask`, no matter people use Django or Flask, the frontend project would have consistent style, which make the project easier to maintain.

### Gulp

I have removed the Gulp code and all work is done by [python-webpack-boilerplate](https://github.com/AccordBox/python-webpack-boilerplate/)

### Bootstrap

This project use Bootstrap theme by default.

The `python-webpack-boilerplate` also contains `bootstrap` by default (installed by using `npm install`), so the page theme can still work.

`custom_bootstrap_compilation` is removed from the `cookiecutter.json` because it is not useful anymore.

### Docker

I also updated some docker code, the workflow can work without problem.


README.rst
```diff
@@ -184,15 +184,14 @@ Answer the prompts with your own desired options_. For example::
     5 - 9.5
     Choose from 1, 2, 3, 4, 5 [1]: 1
     Select js_task_runner:
-    1 - None
-    2 - Gulp
+    1 - Webpack
+    2 - None
```
Andrew-Chen-Wang, line 188:
```suggestion
    1 - None
    2 - Webpack
```
The default should be None as not everybody has or wants a js_task_runner (additionally, None was the original default).

cookiecutter.json
```diff
@@ -25,8 +25,8 @@
     "9.5"
   ],
   "js_task_runner": [
-    "None",
-    "Gulp"
+    "Webpack",
+    "None"
```
Andrew-Chen-Wang, line 29: Same as before regarding defaults.

hooks/post_gen_project.py
```diff
@@ -324,6 +312,19 @@ def remove_storages_module():
...
 
 
+def create_webpack_project():
+    from cookiecutter.main import cookiecutter
+
+    cookiecutter(
+        "https://github.com/AccordBox/python-webpack-boilerplate",
+        directory="frontend_template",
+        no_input=True,
+    )
```
Andrew-Chen-Wang, line 322: This is slightly concerning. Although less maintenance for @browniebroke is great, coming from a Swift background (i.e. SPM), I have a really hard time trusting dependencies solely based on a GitHub URL. But I guess this is up to the maintainers.

Andrew-Chen-Wang (commented):
Nice addition!

hooks/post_gen_project.py
browniebroke: Yes, I share the concern regarding trust... On the maintenance side of things, it's not clear win either: on one hand it's sort of outsourcing this part to an external package, but on the other hand I'm wondering how we can safely manage breaking changes. If this other template changes something we don't expect, we would break, causing maintenance issues. 

We need to make sure `python-webpack-boilerplate` and this line are using the same version, and we need to update this line each time we receive an automated updated for `python-webpack-boilerplate` from PyUP.

Calling `cookiecutter` programmatically is a really interesting idea, but I'm not sure we can consider it unless we own both projects. Right now, this part is a black box to me, and if anything were to go wrong, I would have a hard time debugging it.

{{cookiecutter.project_slug}}/package.json
```diff
@@ -1,38 +0,0 @@
-{
-  "name": "{{cookiecutter.project_slug}}",
-  "version": "{{ cookiecutter.version }}",
-  "dependencies": {},
-  "devDependencies": {
-    {% if cookiecutter.js_task_runner == 'Gulp' -%}
-    {% if cookiecutter.custom_bootstrap_compilation == 'y' -%}
-    "bootstrap": "4.3.1",
```
browniebroke, line 8: The `frontend` project uses Bootstrap 5:

https://github.com/AccordBox/python-webpack-boilerplate/blob/e0ca53cab5301512e4bdd19cc3b217ec05fa8723/frontend_template/%7B%7Bcookiecutter.project_slug%7D%7D/package.json#L55

That's probably going to break things since Crispy forms is configured to use Bootstrap 4:

https://github.com/pydanny/cookiecutter-django/blob/ebc2f0446ac1cd22f5817f8bb63b0bda1d22ea51/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/base.py#L210-L211

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -69,14 +69,6 @@
...
     hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
     INTERNAL_IPS += [".".join(ip.split(".")[:-1] + ["1"]) for ip in ips]
-    {%- if cookiecutter.js_task_runner == 'Gulp' %}
-    try:
-        _, _, ips = socket.gethostbyname_ex("node")
-        INTERNAL_IPS.extend(ips)
-    except socket.gaierror:
-        # The node container isn't started (yet?)
-        pass
-    {%- endif %}
```
browniebroke, line 79: This was needed to show the Django Debug toolbar. Is it no longer a problem?

cookiecutter.json
```diff
@@ -46,7 +46,6 @@
   ],
   "use_async": "n",
   "use_drf": "n",
-  "custom_bootstrap_compilation": "n",
```
browniebroke, line 49: Is it that we can't easily provide a custom Bootstrap compliation anymore or is it always enabled?

docs/deployment-on-heroku.rst
```diff
@@ -94,23 +94,12 @@ Or add the DSN for your account, if you already have one:
 .. _Sentry add-on: https://elements.heroku.com/addons/sentry
 
 
-Gulp & Bootstrap compilation
+Webpack
 ++++++++++++++++++++++++++++
 
-If you've opted for a custom bootstrap build, you'll most likely need to setup
-your app to use `multiple buildpacks`_: one for Python & one for Node.js:
+If you use Webpack as frontend solution, it is recommended to use Docker to deploy your project to Heroku.
```
browniebroke, line 100: Why this change? The multi-buildpack solution doesn't work? In my opinion, the Docker experience on Heroku was not super standard. That's a reason we don't really support Heroku + Docker.

hooks/post_gen_project.py
michael-yin: @Andrew-Chen-Wang @browniebroke 

Well, I think this is the biggest issue in this PR.

My solution is like this:

1. In the `hooks/post_gen_project.py`, I extract version of `python-webpack-boilerplate` from the `requirements/base.txt`
1. If the version is `0.1.2`, then `cookiecutter` the frontend project from Github tag url https://github.com/AccordBox/python-webpack-boilerplate/tree/v2.12.3

This can make sure same version are always used and can also help developers troubleshoot some potential issues.

Do you think is this good to go?

michael-yin:
@browniebroke @Andrew-Chen-Wang 

Hi, I have pushed some commits the PR and now it is ready to review. 

## Workflow

In the python-webpack-boilerplate `0.0.5`, I changed the workflow to generate the frontend app from the package in `site-package` instead of the remote Github repo.

So all the config files are `freezed` and this can help people to troubleshoot in some cases.

And I see a little problem, user need to install `python-webpack-boilerplate` first, and this can not work in the `local docker-compose` command, **because `frontend` does not exist when we run `docker-compose -f local.yml build`**

So we might still need to use `cookiecutter` to create frontend app from the Github tag in `hooks/post_gen_project.py `

```
    cookiecutter(
        f"https://github.com/AccordBox/python-webpack-boilerplate/",
        checkout=f"v{pkg_version}",
        directory="webpack_boilerplate/frontend_template",
        no_input=True,
    )
```

## Doc

I also write a doc to help people who are new to frontend stuff

frontend_doc_screenshot

## Bootstrap 5

Now `python-webpack-boilerplate` is using Bootstrap 5.1.1 by default, and everything works fine with the latest version `cookiecutter-django`

1. People can write SCSS to customize Bootstrap, so I think `custom_bootstrap_compilation` can be deprecated.
2. Modern ES bundle solution
3. SCSS, CSS, JS code linting.

I wish this can help people to write better frontend stuff with Django

Please let me know if you have any questions. 

Thx.

Andrew-Chen-Wang:
> However, this can not work in the local docker-compose command, because frontend does not exist when we run docker-compose -f local.yml build, the node service will break

how about putting it in the entry point file?

michael-yin:
@Andrew-Chen-Wang 

I do not think creating `frontend` app in the `entry point file` is good option here.

I will think about it for a while.

Thx.

{{cookiecutter.project_slug}}/package.json
charleshan: Regarding the concern about breaking changes between an external package, we can add `checkout` argument for cookiecutter to choose a specific branch, tags, or commit.

https://cookiecutter.readthedocs.io/en/1.7.2/usage.html#works-directly-with-git-and-hg-mercurial-repos-too

PR #3109, opened 2021-03-28
Add watchman reloader support

### LINKED ISSUE #3092: Use pywatchman and watchman support for reload
## Description

Django has built-in support for watchman file watcher for auto reloading.

## Rationale

Polling based watcher adds a decent amount of CPU and IO load for large projects. Enough to be noticeable. In my case CPU spikes went from 20% to close to 0.

[Watchman](https://facebook.github.io/watchman/) is a widely used file watcher and [Django runserver has built-in integration](https://docs.djangoproject.com/en/3.1/ref/django-admin/#runserver).

I am happy to make PR if there is interest. I added watchman installation step in `compose/local/django/Dockerfile` and pywatchman in `requirements/local.txt`.

snmishra:
## Description

Add WatchmanReloader support by installing watchman and pywatchman in Docker. For local development without docker, add hints about installing watchman.

Checklist:

- [ X] I've made sure that `tests/test_cookiecutter_generation.py` is updated accordingly (especially if adding or updating a template option)
- [ X] I've updated the documentation or confirm that my change doesn't require any updates

## Rationale

The default polling based watcher (StatReloader) adds a decent amount of CPU and IO load for large projects, enough to be noticeable. In my case CPU spikes went from 20% to close to 0.

Watchman is a widely used file watcher and Django runserver has built-in integration.

Fix #3092

### DIFF
--- a/hooks/pre_gen_project.py
+++ b/hooks/pre_gen_project.py
@@ -95,7 +95,5 @@ if (
     "{{ cookiecutter.use_async }}".lower() == "y"
     and "{{ cookiecutter.use_watchman }}".lower() == "y"
 ):
-    print(
-        "Async and watchman don't work together "
-    )
+    print("Async and watchman don't work together ")
     sys.exit(1)
--- a/hooks/pre_gen_project.py
+++ b/hooks/pre_gen_project.py
@@ -86,6 +86,16 @@ if (
     and "{{ cookiecutter.use_watchman }}".lower() == "y"
 ):
     print(
-        "Automatic watchman installation is only setup in Docker for now. You can install "
-        "watchman from https://github.com/facebook/watchman/releases and pywatchman from pypi."
+        "Automatic watchman installation is only setup in Docker for now. Please see "
+        '"Getting Up and Running Locally" section of cookiecutter-django docs for details '
+        "on local installation"
     )
+
+if (
+    "{{ cookiecutter.use_async }}".lower() == "y"
+    and "{{ cookiecutter.use_watchman }}".lower() == "y"
+):
+    print(
+        "Async and watchman don't work together "
+    )
+    sys.exit(1)
--- a/hooks/pre_gen_project.py
+++ b/hooks/pre_gen_project.py
@@ -80,3 +80,12 @@ if (
         "You should either use AWS or select a different Mail Service for sending emails."
     )
     sys.exit(1)
+
+if (
+    "{{ cookiecutter.use_docker }}".lower() == "n"
+    and "{{ cookiecutter.use_watchman }}".lower() == "y"
+):
+    print(
+        "Automatic watchman installation is only setup in Docker for now. You can install "
+        "watchman from https://github.com/facebook/watchman/releases and pywatchman from pypi."
+    )
--- a/tests/test_cookiecutter_generation.py
+++ b/tests/test_cookiecutter_generation.py
@@ -74,7 +74,8 @@ SUPPORTED_COMBINATIONS = [
     {"cloud_provider": "GCP", "mail_service": "Other SMTP"},
     # Note: cloud_providers GCP and None with mail_service Amazon SES is not supported
     {"use_async": "y"},
-    {"use_async": "n"},
+    {"use_async": "n", "use_watchman": "y"},
+    {"use_async": "n", "use_watchman": "n"},
     {"use_drf": "y"},
     {"use_drf": "n"},
     {"js_task_runner": "None"},
@@ -93,8 +94,6 @@ SUPPORTED_COMBINATIONS = [
     {"use_whitenoise": "n"},
     {"use_heroku": "y"},
     {"use_heroku": "n"},
-    {"use_watchman": "y"},
-    {"use_watchman": "n"},
     {"ci_tool": "None"},
     {"ci_tool": "Travis"},
     {"ci_tool": "Gitlab"},
--- a/tests/test_cookiecutter_generation.py
+++ b/tests/test_cookiecutter_generation.py
@@ -93,6 +93,8 @@ SUPPORTED_COMBINATIONS = [
     {"use_whitenoise": "n"},
     {"use_heroku": "y"},
     {"use_heroku": "n"},
+    {"use_watchman": "y"},
+    {"use_watchman": "n"},
     {"ci_tool": "None"},
     {"ci_tool": "Travis"},
     {"ci_tool": "Gitlab"},
--- a/{{cookiecutter.project_slug}}/.watchmanconfig
+++ b/{{cookiecutter.project_slug}}/.watchmanconfig
@@ -1,0 +1,13 @@
+{
+  "ignore_dirs": [
+    "node_modules",
+    ".envs",
+    "bin",
+    "lib",
+    "include",
+    ".git",
+    ".idea",
+    "templates",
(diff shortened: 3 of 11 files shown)


tests/test_cookiecutter_generation.py
```diff
@@ -93,6 +93,8 @@ def context():
     {"use_whitenoise": "n"},
     {"use_heroku": "y"},
     {"use_heroku": "n"},
+    {"use_watchman": "y"},
+    {"use_watchman": "n"},
```
Andrew-Chen-Wang, line 97: Watchman should not be used with `use_async="y"` option as uvicorn does not support watchman, only watchgod.

{{cookiecutter.project_slug}}/.watchmanconfig
```diff
@@ -0,0 +1,13 @@
...
+    "node_modules",
+    ".envs",
+    "bin",
+    "lib",
+    "include",
+    ".git",
+    ".idea",
+    "templates",
+    "static_assets"
+  ]
```
Andrew-Chen-Wang, line 12: Check if using Docker. If the dev is, you should add "compose".

{{cookiecutter.project_slug}}/.watchmanconfig
```diff
@@ -0,0 +1,13 @@
+{
+  "ignore_dirs": [
+    "node_modules",
+    ".envs",
+    "bin",
+    "lib",
+    "include",
+    ".git",
+    ".idea",
+    "templates",
```
Andrew-Chen-Wang, line 10:
```suggestion
    "templates",
    "static",
```
I believe static files are also not included. The browser when testing caches them, sure as you can invalidate the browser cache of static files, but Django doesn't care, just like the templates dir.

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
```diff
@@ -10,10 +10,31 @@ RUN apt-get update \
...
+WORKDIR /watchman-build
+RUN curl -sSLO https://github.com/facebook/watchman/releases/download/${WATCHMAN_VERSION}/watchman-${WATCHMAN_VERSION}-linux.zip && \
+    unzip watchman-${WATCHMAN_VERSION}-linux.zip && \
+    chmod +x watchman-${WATCHMAN_VERSION}-linux/bin/* \
+      watchman-${WATCHMAN_VERSION}-linux/lib/* && \
+    cp -rp watchman-${WATCHMAN_VERSION}-linux/bin/* /usr/local/bin/ && \
+    cp -rp watchman-${WATCHMAN_VERSION}-linux/lib/* /usr/local/lib/ && \
+    mkdir -p /usr/local/var/run/watchman && chown 2777 /usr/local/var/run/watchman
+# The pywatchman version on pypi is really outdated, build from source
+RUN svn export https://github.com/facebook/watchman/tags/${WATCHMAN_VERSION}/python
```
Andrew-Chen-Wang, line 32: Never used svn before. Try and use git since we're on "GitHub"

{{cookiecutter.project_slug}}/local.yml
```diff
@@ -9,6 +9,10 @@ services:
     build:
       context: .
       dockerfile: ./compose/local/django/Dockerfile
+      {%- if cookiecutter.use_watchman == 'y' %}
+      args:
+        WATCHMAN_VERSION: v2021.03.01.00
```
Andrew-Chen-Wang, line 14: I feel like this isn't necessary to be placed here and instead just be placed in the docker file itself.

hooks/pre_gen_project.py
```diff
@@ -80,3 +80,12 @@
...
     )
     sys.exit(1)
+
+if (
+    "{{ cookiecutter.use_docker }}".lower() == "n"
+    and "{{ cookiecutter.use_watchman }}".lower() == "y"
+):
+    print(
+        "Automatic watchman installation is only setup in Docker for now. You can install "
+        "watchman from https://github.com/facebook/watchman/releases and pywatchman from pypi."
```
Andrew-Chen-Wang, line 90: Idk if you want to mention pywatchman from pypi. Instead, say something like "refer to the cookiecutter-django docs for Python installation"

Andrew-Chen-Wang (changes requested):
This is a great addition @snmishra!

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
snmishra: Yeah, I wasn't sure about that. `git` doesn't let you download just the python, you'd have to clone the whole thing. But then we're deleting the thing afterward, so it shouldn't cause image size to go up. I will switch to git

{{cookiecutter.project_slug}}/local.yml
snmishra: Ah yes, I had missed that ARG can take a default value. Makes sense to put it in the Dockerfile

hooks/pre_gen_project.py
snmishra: True

snmishra:
Thanks for the encouragement and quick review @Andrew-Chen-Wang I'll get to these tomorrow, hopefully

snmishra:
I just discovered that cookiecutter-django is using Werkzeug via `runserver_plus` by default. Werkzeug doesn't support Watchman, but does [support watchdog](https://werkzeug.palletsprojects.com/en/1.0.x/serving/#reloader). I don't remember why I switched back to runserver.

So I don't think this PR is needed. What do you think @Andrew-Chen-Wang

{{cookiecutter.project_slug}}/.watchmanconfig
snmishra: Done

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
snmishra: Switched to downloading zip. Why didn't I think of that before, duh?

Andrew-Chen-Wang:
@snmishra Yea this PR might not be needed if people use runserver_plus. But:

> I don't remember why I switched back to runserver.

Interestingly enough, I also switched back to runserver but it was mostly because of my habit of using my run configuration in PyCharm IDE, maybe also because of the UI but I'm not entirely sure.

However, you mentioned the watchdog package. I don't think that's actually included in this package. We currently use watchgod for the async packages to decrease uvicorn CPU usage. I think if you create a new PR with watchdog, that'll be great!

snmishra:
@Andrew-Chen-Wang Unfortunately, watchdog doesn't reduce CPU usage for me. Not sure why. 

I guess it's worth adding in the requirements anyways. Should `use_celery` option also use watchdog or continue using watchgod? I don't know if watchgod for celery and watchgod for uvicorn share information so that there's just one set of stat happening.

snmishra:
> Interestingly enough, I also switched back to runserver but it was mostly because of my habit of using my run configuration in PyCharm IDE, maybe also because I dislike the debug UI but I'm not entirely sure.

I think that's what did it for me as well. I have been using `ptvsd` with VS Code and runserver_plus doesn't play well with that, IIRC.

snmishra:
Closing request based on above discussion.

PR #2664, opened 2020-07-06
ADD instructions on how to set up a secure development environment.

Afrowave:
## Description

Facebook have moved their users to [secure by default](https://www.facebook.com/notes/facebook-engineering/secure-browsing-by-default/10151590414803920/#:~:text=We've%20focused%20on%20making,long%20used%20https%20as%20well.). 

## Rationale

If a developer chooses to build *Facebook* login capability with the `users` app, then setting `https` in the developer environment will be necessary.

If required, I can add this capability through a new PR so that it becomes part of the project configuration.

PR #3111, opened 2021-03-29
Docs: Adds details to on how to set up a secure development environment

### LINKED ISSUE #2840: https://my-dev-env.local serving pages from both django and docs containers
## What happened?

After implementing the steps described under the "Developing locally with HTTPS" section of the docs, https://my-dev-env.local served pages randomly from both the django and docs containers, resulting in returning successful pages (200) at times and returning 404 page not found errors at other times.

https://my-dev-env.local should only serve pages from the django container.

## Additional details

From my observations, this is due to both django and docs containers being included within the "upstream" section of the generated nginx-proxy '/etc/nginx/conf.d/default.conf' file.

This is in turn caused by both django: and docs: within local.yml pointing to the same .env file (/.envs/.local/.django). Hence when VIRTUAL_HOST and VIRTUAL_PORT are added to this .env file (as instructed in the docs), they ended up linking nginx-proxy to both django and docs containers.

As a workaround, I simply created a /.envs/.local/.docs, pointed to this new env file in local.yml (under docs:), rebuilt the containers, and https://my-dev-env.local is serving everything properly now.

I'm wondering if there's a more elegant way around this or is there something I may have missed out in my set up?

* Host system configuration: 
    * Version of cookiecutter CLI (get it with `cookiecutter --version`): 1.7.2
    * OS: MacOS 10.15.6 
    * Python version: 3.8.5
    * Docker versions (if using Docker): 19.03.12

Afrowave:
## Description

This PR adds details and corrects some erros that were occuring as one set out to install TLS certificates in the developer environment.

This adds the need to edit `/etc/hosts`.
This adds instructions on using `django-sslserver` for non-docker dev environments.

Checklist:

- [x] I've updated the documentation

Fix #2664
Fix #2840

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -97,8 +97,8 @@ The most important thing for us here now is ``env_file`` section enlisting ``./.
 
     .envs
     ├── .local
-    │   ├── .django
-    │   └── .postgres
+         ├── .django
+         └── .postgres
     └── .production
         ├── .django
         └── .postgres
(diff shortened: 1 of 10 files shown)


docs/developing-locally-docker.rst
```diff
@@ -255,32 +298,39 @@ local.yml
...
 
-  ALLOWED_HOSTS = ["localhost", "0.0.0.0", "127.0.0.1", "my-dev-env.local"]
+  ALLOWED_HOSTS = ["localhost", "0.0.0.0", "127.0.0.1", "my-dev-env"]
 
 Rebuild your ``docker`` application. ::
 
   $ docker-compose -f local.yml up -d --build
 
-Go to your browser and type in your URL bar ``https://my-dev-env.local``
+Go to your browser and type in your URL bar ``https://my-dev-env:8800``.
```
linusthe3rd, line 333: How does the 8800 port apply to the Docker container approach?

I think this is a typo and meant for the Django SSL Server approach.

docs/developing-locally-docker.rst
```diff
@@ -255,32 +298,39 @@ local.yml
...
     
     ...
+#. On the django service, confirm that you are connected to the ``envs`` files and remove/comment out the ``ports``. ::
+
+    ...
+    env_file:
+      - ./.envs/.local/.django
+    ...
+    # ports:
+    #  - 8000:8000
```
linusthe3rd, line 308: I followed these instructions and needed to keep this configuration in order to get local SSL to work.

browniebroke:
Closing in favour of #5252

PR #3112, opened 2021-03-30
Fix gitlab-ci using duplicate key name for image

Tamerz:
## Description

When choosing to use Docker and GitLab, there is a duplicate key value for the image name for pytest.

```
pytest:
  stage: test
  image: python:3.8
  image: docker/compose:latest
```

Checklist:

- [ ] I've made sure that `tests/test_cookiecutter_generation.py` is updated accordingly (especially if adding or updating a template option) **I don't see any options in the tests for this.**

## Rationale
The duplicate key didn't seem to cause any problems, but it is incorrect.

### DIFF
--- a/{{cookiecutter.project_slug}}/.gitlab-ci.yml
+++ b/{{cookiecutter.project_slug}}/.gitlab-ci.yml
@@ -21,7 +21,6 @@ flake8:
 
 pytest:
   stage: test
-  image: python:3.8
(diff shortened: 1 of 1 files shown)


Tamerz:
I tried adding a test, but it seems `yaml.safe_load()` simply ignores duplicate keys and takes the second value.

browniebroke (approved):

browniebroke:
Thanks!

PR #3150, opened 2021-05-02
Add an assertion to fix mypy type error

arnav13081994:
## Description

Added a simple assertion statement that helps mypy figure out that the user it is dealing with is authenticated

Checklist:

## Rationale

Just gets rid of having to explicitly ignore the warning.


browniebroke (approved):

browniebroke:
Thanks for the contribution 🎉

PR #3151, opened 2021-05-02
Updated the tests_views module by replacing RequestFactory with Djang…

arnav13081994:
## Description

Refactored `test_views.py` to use `Client` instead of `RequestFactory`. 

Checklist:

## Rationale

Doing this has the following benefits:

1) Code is more readable and easier to understand.
1) No need to route the request through various middlewares. `Client` does that for us.
1) No need to define a dummy middleware class.
1) No need to add `# type: ignore` to get rid of mypy warnings
1) Provides a better end-to-end test especially for `POST` requests.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
@@ -1,101 +1,132 @@
 import pytest
 from django.conf import settings
 from django.contrib import messages
-from django.contrib.auth.models import AnonymousUser
-from django.contrib.messages.middleware import MessageMiddleware
-from django.contrib.sessions.middleware import SessionMiddleware
-from django.http import HttpRequest
-from django.test import RequestFactory
+from django.test import Client
 from django.urls import reverse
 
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
Client is really really really slow due to middleware and template loads... especially once you write hundreds of tests, it's not fun.

arnav13081994:
I agree, but this project doesn't have many tests so overall time for test execution would not change by that much. And I personally feel the trade off with time for easier to understand code is worth it.

Andrew-Chen-Wang:
Personally, I'm glad this project taught me to use request factory, especially rf for class based views. Client is great when you need to work with several components at a time -- and client is really easy to use -- for overall testing purposes. But unit testing is just that: in units.

Yes, time here would not cost much as an open source project with limited tests, but again, scaling on your private repositories using costly CI minutes, if you can't afford your own runner, is a big ouch. I think this repo using rf is a good practice that should be adopted more by users, so replacing these tests with clients is a regression to me.

browniebroke:
Ok, so I'm replying to this old PR again, I agree with @Andrew-Chen-Wang here (thanks for the great explanation btw). I suggest we close this and leave the existing tests as is.

Thanks for the suggestion @arnav13081994!

PR #3152, opened 2021-05-02
Added mypy pre-commit hook

arnav13081994:
## Description

Added `mypy` pre-commit hook

Checklist:

## Rationale

Adding `mypy` as another pre-commit hook will have the following benefits:

1) Compile time errors can be discovered while developing locally.
1) Compile time errors can be discovered faster


foarsitter:
I'm against this change since `mypy` cannot fix the problems it encounters. Some typing problems are hard to solve and this would block the user from commiting which would disadvantage beginners.

{{cookiecutter.project_slug}}/mypy.sh
browniebroke, line 1: That seems a bit too complicated to me to maintain... It's a bit too beyond my bash skills 😄 

How should folks run this hook? Should it run inside of the Docker container? If it doesn't work outside of Docker than it's a blocker.

{{cookiecutter.project_slug}}/.github/workflows/ci.yml
```diff
@@ -28,6 +28,11 @@ jobs:
         with:
           python-version: 3.9
 
+      - name: Install All Project Dependencies for pylint and mypy to check
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements/local.txt
```
browniebroke, line 34: That seems like a big overhead

browniebroke (commented):
I'm also against this change. I personally think that it could be nice to run mypy on pre-commit (in terms of UX).

Unfortunately, the fact that it needs the project dependencies in a blocker in my opinion.

PR #3168, opened 2021-05-11
Async don't reload templates + static

### LINKED ISSUE #3093: Move template and static to root
## Description

Move the template directory and the static directory to the root level

## Rationale

It's mostly a uvicorn reloading issue. I haven't tested this yet, but the current configuration always reloads if HTML / static files are edited, but on Django they do not reload. Will look a little more into this.

Andrew-Chen-Wang:
## Description

Make async reloading much more efficient by only reloading Python files. No template/static file should be reloaded. Fixes #3093 (doesn't implement what was proposed but the general idea is done)

Checklist:

## Rationale

Watchgod doesn't need to check HTML, CSS, JS etc. files. Like the Django statreloader, we only need to reload the Python files. These other files are automatically changed because Django is just loading files. This makes fixing HTML, CSS, and JS files much faster to fix since you don't have to wait for the server to finish reloading.

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -115,6 +115,12 @@ def remove_async_files():
         os.remove(file_name)
 
 
+def remove_async_docker_files():
+    file_names = [os.path.join("compose", "local", "django", "run.py")]
+    for file_name in file_names:
+        os.remove(file_name)
+
+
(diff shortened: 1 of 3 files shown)


Andrew-Chen-Wang:
Latest uvicorn has a CLI option for excluding. Closing.

PR #3174, opened 2021-05-12
Added Django's current language to the lang attribute of the html tag

douglascdev:
## Description

Add Django's current language to the lang attribute of the html tag.

Checklist:

## Rationale
Allows browsers to know the correct language in use.
In the current state, the user's browser will always interpret the text as english, causing translation issues if the browser is set to translate english while the text is another language. This could also cause issues with screen readers.
I don't think it's necessary to update the `tests/test_cookiecutter_generation.py` script or the docs.


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/base.html
```diff
@@ -1,5 +1,6 @@
 {% raw %}{% load static i18n {% endraw %}{% if cookiecutter.use_compressor == "y" %}compress{% endif %}{% raw %}%}<!DOCTYPE html>
-<html lang="en">
+{% get_current_language as LANGUAGE_CODE %}
+<html lang="{{ LANGUAGE_CODE }}">
```
browniebroke, line 3: With this, I'm getting a different value of `en-gb`. This seems to go against [the W3C recommendations](https://www.w3.org/International/articles/language-tags/):

> The golden rule when creating language tags is to keep the tag as short as possible. Avoid region, script or other subtags except where they add useful distinguishing information.

browniebroke: I did a bit more reserch, this is fine. It's certainly more correct that always being "en".

browniebroke:
Thank you for the fix!

PR #3230, opened 2021-06-20
Update docs/howto.rst

Andrew-Chen-Wang:
## Description

Update docs/howto.rst after the PR that fixed docs building. I also wonder if we should add this "howto" portion to the README. Not sure. (if you need to check this, visit https://docs.hearye.us)

Checklist:

## Rationale

The _source directory doesn't exist

### DIFF
--- a/{{cookiecutter.project_slug}}/docs/howto.rst
+++ b/{{cookiecutter.project_slug}}/docs/howto.rst
@@ -4,20 +4,19 @@ How To - Project Documentation
 Get Started
 ----------------------------------------------------------------------
 
-Documentation can be written as rst files in the `{{cookiecutter.project_slug}}/docs/_source`.
+Documentation can be written as rst files in `{{cookiecutter.project_slug}}/docs`.
 
 {% if cookiecutter.use_docker == 'n' %}
-To build and serve docs, use the command:
-    ::
+To build and serve docs, use the command::
     
-        make livehtml 
+    make livehtml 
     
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/docs/howto.rst
```diff
@@ -4,20 +4,19 @@ How To - Project Documentation
 Get Started
 ----------------------------------------------------------------------
 
-Documentation can be written as rst files in the `{{cookiecutter.project_slug}}/docs/_source`.
+Documentation can be written as rst files in `{{cookiecutter.project_slug}}/docs`.
```
browniebroke, line 7: I guess the intend of these `` ` `` is to have path formatted as `preformatted` not _italic_, which is done with double backtick ``` `` ``` in RST, single backtick work in Markdown... 

WDYT? Would you mind fixing them in this file?

browniebroke (approved):
LGTM

{{cookiecutter.project_slug}}/docs/howto.rst
Andrew-Chen-Wang: ah sorry I missed this comment! Yes, I agree. I can fix those in a separate pr.

browniebroke: Yes, wan't directly related, so I decided to go ahead with it.

PR #3239, opened 2021-06-25
Wrap jQuery call in `DOMContentLoaded` event listener on account email page

wwwtony5488:
## Description

Fix error occurred while loading jQuery with defer and call it but didn't wait for the whole package downloaded.

Checklist:

## Rationale
The line `$('.form-group').removeClass('row');` is not executable, because at that moment `$` is not defined,  unless wrap it in DOMContentLoaded


browniebroke (approved):

browniebroke:
Thanks for the fix!

PR #3255, opened 2021-07-05
Upgrade to Redis 6

### LINKED ISSUE #2758: Use Redis 6.0
https://github.com/pydanny/cookiecutter-django/blob/6f02a737bdf7b6bfaca94511587e30e7e6cfa344/%7B%7Bcookiecutter.project_slug%7D%7D/local.yml#L67

Redis 6.0 had some nice features added. I'm not sure about stability of those features, but for the most part, for web devs, Redis shouldn't have changed to much between major versions to be unstable. So should cookiecutter-django upgrade Redis?

Andrew-Chen-Wang:
## Description

Use Redis 6 instead of 5. I'm not sure if we want to specify such a specific number though. Last I checked, Redis automatically updates itself every time you restart Redis, so maybe we can specify an image like `redis:6`? Thoughts?

Checklist:

## Rationale

Fixes #2758

### DIFF
--- a/{{cookiecutter.project_slug}}/local.yml
+++ b/{{cookiecutter.project_slug}}/local.yml
@@ -64,7 +64,7 @@ services:
   {%- if cookiecutter.use_celery == 'y' %}
 
   redis:
-    image: redis:5.0
+    image: redis:6.2.4
     container_name: redis
 
   celeryworker:
--- a/{{cookiecutter.project_slug}}/local.yml
+++ b/{{cookiecutter.project_slug}}/local.yml
@@ -64,7 +64,7 @@ services:
   {%- if cookiecutter.use_celery == 'y' %}
 
   redis:
-    image: redis:6.2.4
+    image: redis:6
     container_name: redis
 
   celeryworker:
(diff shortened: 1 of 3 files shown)


browniebroke:
> I'm not sure if we want to specify such a specific number though. Last I checked, Redis automatically updates itself every time you restart Redis, so maybe we can specify an image like redis:6? Thoughts?

I'd be ok with that, yes.

{{cookiecutter.project_slug}}/production.yml
```diff
@@ -47,7 +47,7 @@ services:
       {%- endif %}
 
   redis:
-    image: redis:5.0
+    image: redis:6.2.4
```
browniebroke, line 50: Did you leave the specific version in `local.yml` and `production.yml` on purpose? I don't think we need to be too specific here either.

Andrew-Chen-Wang: Fixed

{{cookiecutter.project_slug}}/local.yml
```diff
@@ -64,7 +64,7 @@ services:
   {%- if cookiecutter.use_celery == 'y' %}
 
   redis:
-    image: redis:5.0
+    image: redis:6.2.4
```
browniebroke, line 67:
```suggestion
    image: redis:6
```

PR #3265, opened 2021-07-11
Make `django` depend on `redis` in local Docker 

PoDuck:
## Description

The django container needs to depend on redis in both local and production docker-compose files.  This caused me some confusion while implementing services that depended on redis, so I decided to fix it.

Checklist:

I ran the tests, and it broke nothing.  No new tests needed.
This should not need any changes in the documentation.


{{cookiecutter.project_slug}}/local.yml
```diff
@@ -13,6 +13,7 @@ services:
     container_name: django
     depends_on:
       - postgres
+      - redis
```
Andrew-Chen-Wang, line 16: This is only necessary if celery is installed

Andrew-Chen-Wang (changes requested):

{{cookiecutter.project_slug}}/local.yml
browniebroke: Agree, there is no redis container if Celery isn't used.

browniebroke (approved):
Should be good to go now

browniebroke:
Thanks 🎉

PR #3268, opened 2021-07-14
Add documentation for automating backups

Mike97M:
## Description

Backup documentation to show how to use "docker-compose -f local.yml ps -q postgres" to get the container ID instead of hard coding it.

Checklist:

## Rationale

For automating backups, updating the container ID every time we start the database isn't really double, this is why we can use the above command to get the ID using docker-compose.


browniebroke (approved):

browniebroke:
Thanks!

PR #3277, opened 2021-07-27
Convert trans to translate in templates

Andrew-Chen-Wang:
## Description

Convert all django tags of "trans" to "translate"

Checklist:

## Rationale

Django might one day just drop "trans" as a tag, so I thought just update all our tags rn.


browniebroke (approved):
I didn't realise they've renamed this tag in 3.1. Good catch :+1:

PR #3280, opened 2021-07-27
Update Celery to v5

browniebroke:
## Description

Checklist:


browniebroke:
I think this is now ready to go. Would be nice if someone else could take another look before I merge. The main update is around the command line options, in v5, the glocal option (app, broker, etc...) must be passed before the subcommand. I had to update a few places.

Tried locally as well as deploying to prod with Heroku and Docker, everything seems ok.

steveputman:
Working for me (local Docker and production Docker).

PR #3284, opened 2021-07-30
Update project-generation-options.rst

tanushyara:
## Description

Checklist:


luzfcb:
@tanushyara Thanks for wanting to contribute to this project, but Migomail and Mailcot are not an option of the current [cookiecutter.json](https://github.com/pydanny/cookiecutter-django/blob/fd41f399d85007cf4c74d0b41e6aa77cda7069d1/cookiecutter.json#L36-L44) configuration and we have no plans to add it in the future.

PR #3287, opened 2021-08-06
How to pre-commit in Docker Development

### LINKED ISSUE #3086: Improve docs on docker use
## Description

After setting up a new project, everything works, but moving forward, there are many moving pieces that the documentation covers and a lot that it doesn't. I'm still trying to piece it together, but probably due to this not being my beach, I end up with a lot of lacking information (I ended finding it, so I compile this issue). 

I propose to add the following to the docs, section by section. If this makes sense, I can add this to the docs and do a PR.

### [Getting Up and Running Locally With Docker](https://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html)

This section is in a tutorial fashion, and I think that is very good.

However, I would love to see a section on operations with docker-compose, explaining that ``docker-compose -f local.yml run --rm django`` (referenced in [Execute Management Commands](https://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html#execute-management-commands)) can be used to run things the way you would run locally. 

It would also make sense, being this a tutorial section, to explain or point to documentation explaining how to run the docker CLI (I think running docker-compose always is a bit of overkill, principally when you have the stack running for dev already).

I would also add: 

> The ```local.yml``` maps docker app with your project directory, so you are working "locally" (every change will reload the server, and every command that changes files will result in a local change).

#### PyCharm

A simple note explaining that the docs explain how to set up would be enough to save time for first-timmers. I did a lot of stuff in PyCharm before setting the docs.

### [Linters](https://cookiecutter-django.readthedocs.io/en/latest/linters.html)

This section explains linters, but what about specifics with docker?

Some problems and solutions

1. Python environment is in docker. 
    -  Used Docker CLI or ``docker-compose -f local.yml run --rm django`` to run, and all works as expected!
    
2. The container used doesn't have GIT, and [pre-commit](https://pre-commit.com/) (that, by the way, should be in the docs because it gives failure in CI) requires git.
    - Installed a local virtual environment in the dev folder where git is set up and run pre-commit from there.
    - Didn't try it, but it makes sense that installing git on the container is a solution. Should it be there for local? (I can create another issue for this)

### [Testing](https://cookiecutter-django.readthedocs.io/en/latest/testing.html)

Perfect! I had no doubts.

## Rationale

This correction on the docs would improve project start-up time for first-timmers.

Qoyyuum:
## Description

It is clear that pre-commit cannot work inside Docker containers and it can only work with git stored on the local machine since its not in a shared volume with the containers. Therefore, it should be noted that anyone who uses this cookicutter with pre-commit already generated by default in every project, should start using pre-commit into their general workflow.

Checklist:

## Rationale

Fixes #3086

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -18,10 +18,11 @@ Prerequisites
 
 * Docker; if you don't have it yet, follow the `installation instructions`_;
 * Docker Compose; refer to the official documentation for the `installation guide`_.
+* Pre-commit; refer to the official documentation for the `installation guide`_.
 
 .. _`installation instructions`: https://docs.docker.com/install/#supported-platforms
 .. _`installation guide`: https://docs.docker.com/compose/install/
-
+.. _`pre-commit`: https://pre-commit.com/#install
 
 Build the Stack
 ---------------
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang (changes requested):
You shouldn't install everything from local.txt. Simply doing pre-commit install is good enough as an environment with all the needed hooks are stored elsewhere, not inside a virtual environment. Instead, the docs should say "to use pre-commit, you should have pre-commit installed globally"

Qoyyuum:
> 
> 
> You shouldn't install everything from local.txt. Simply doing pre-commit install is good enough as an environment with all the needed hooks are stored elsewhere, not inside a virtual environment. Instead, the docs should say "to use pre-commit, you should have pre-commit installed globally"

Good idea. I've reworded it so it should be installed as part of a prerequisite.

Andrew-Chen-Wang (approved):

browniebroke (approved):

browniebroke:
Thanks for the clarification!

PR #3290, opened 2021-08-09
Moved isort config from `.editorconfig` to `setup.cfg`

### LINKED ISSUE #3201: Move isort config from .editorconfig to setup.cfg
## Description

Move the config settings for isort from .editorconfig to setup.cfg (or some other file, like isort.cfg or pyproject.toml). It would go from this in .editorconfig
```
[*.py]
line_length = 88
known_first_party = {{cookiecutter.project_slug}},config
...
```
to this in setup.cfg
```
[*.py]
line_length = 88
known_first_party = {{cookiecutter.project_slug}},config
...
```

## Rationale

Atom's (and I suspect others') editorconfig package does not have a builtin integration with isort, so when I open the editorconfig file it reads the isort configuration as an error. Putting it under the [isort] block instead of under [*.py] also makes it clear what exactly is being configured by that block (it took me a while to figure out why those settings were there in the first place since this is my first time using isort). Per their [documentation](https://pycqa.github.io/isort/docs/configuration/config_files/), .editorconfig is also the last place isort looks for settings, so moving up that list is likely an improvement.

mfosterw:
## Description

Checklist:

## Rationale

Per #3201, moves config options for isort to be more sensibly labeled and alongside other package configurations, as well as to follow the recommendations in the isort docs. Also removes the "recursive" option as per [this comment](https://github.com/pydanny/cookiecutter-django/issues/3201#issuecomment-878664767) on the issue and [the docs](https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html).

Fix #3201

### DIFF
--- a/{{cookiecutter.project_slug}}/.editorconfig
+++ b/{{cookiecutter.project_slug}}/.editorconfig
@@ -12,18 +12,6 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
-[*.py]
-line_length = 88
-known_first_party = {{cookiecutter.project_slug}},config
-multi_line_output = 3
-default_section = THIRDPARTY
-recursive = true
-skip = venv/
-skip_glob = **/migrations/*.py
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-
 [*.{html,css,scss,json,yml}]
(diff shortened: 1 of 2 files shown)


browniebroke:
Thanks!

PR #3293, opened 2021-08-12
Fix RTD build image to support Python 3.9

Andrew-Chen-Wang:
## Description

* Python 3.9 is technically supported via the testing image.

Checklist:

## Rationale

Python 3.9 version is not actually available as a version, only via the image, which is the officially supported place to get Python 3.9. This is probably because most open source projects using RTD uses Python 3.6

### DIFF
--- a/{{cookiecutter.project_slug}}/.readthedocs.yml
+++ b/{{cookiecutter.project_slug}}/.readthedocs.yml
@@ -3,7 +3,10 @@ version: 2
 sphinx:
   configuration: docs/conf.py
 
-python:
+build:
   image: testing
+
+python:
+  version: 3.9
   install:
     - requirements: requirements/local.txt
--- a/{{cookiecutter.project_slug}}/.readthedocs.yml
+++ b/{{cookiecutter.project_slug}}/.readthedocs.yml
@@ -4,6 +4,6 @@ sphinx:
   configuration: docs/conf.py
 
 python:
-  version: 3.9
+  image: testing
   install:
     - requirements: requirements/local.txt


{{cookiecutter.project_slug}}/.readthedocs.yml
```diff
@@ -4,6 +4,6 @@ sphinx:
   configuration: docs/conf.py
 
 python:
-  version: 3.9
+  image: testing
```
browniebroke, line 7: That doesn't seem right... That's not how one specifies the build image according to [the docs](https://docs.readthedocs.io/en/stable/config-file/v2.html#build):

```
version: 2

build:
  image: latest

python:
  version: "3.7"
```

However, even [in the testing image](https://hub.docker.com/r/readthedocs/build/), I don't think 3.9 is supported:

> readthedocs/build:7.0 testing Ubuntu 18.04 supporting Python 2.7, 3.5, 3.6, 3.7, 3.8 and PyPy3.5-7.0.0.

Also, they recommend not using the testing image:

> For internal development testing only, not available for public usage yet.

I think we should stick to Python 3.8, the `latest` image is used by default, so there is no need to specify it.

browniebroke (commented):
Good catch on this problem, didn't realised that 3.9 wasn't supported!

{{cookiecutter.project_slug}}/.readthedocs.yml
Andrew-Chen-Wang: Ah you're right about the `build:` needed, not `python:`. But I disagree downgrading. If you were to have simple typing like:

```python
def generate_authors() -> list[str]:
    return []
```

then readthedocs will fail because you're typing is using Python 3.9+ syntax.

browniebroke: Hum, ok, I see.

PR #3331, opened 2021-09-29
Update deployment-on-heroku.rst

raybesiga:
This sets the Heroku buildpack for the app you have created with Cookiecutter without creating a new Heroku app on the command line.

## Description

Checklist:


browniebroke:
At this point of the deployment, the user shouldn't have an app on Heroku yet, hence why we need to call `heroku create`. if you have already an app, then you can adjust the guide, but I think we should assume that nothing exists and hence the command is correct.

Let me know if I missed something...

PR #3333, opened 2021-09-29
Modified views.py and models.py

raghav2404:
## Description

Checklist:


luzfcb:
@raghav2404 Can you please explain why we need this change? What is the problem this solves?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/views.py
```diff
@@ -21,7 +21,7 @@ class UserDetailView(LoginRequiredMixin, DetailView):
 class UserUpdateView(LoginRequiredMixin, SuccessMessageMixin, UpdateView):
 
     model = User
-    fields = ["name"]
+    fields = ['first_name','last_name']
```
luzfcb, line 24: We purposely do not use `first_name` and `last_name` because they are not flexible enough to support the many possible ways that names are defined in various countries and cultures.

PR #3337, opened 2021-10-01
Use reload-excludes in compose start

Andrew-Chen-Wang:
## Description

Exclude certain dirs for uvicorn reload which Django already does

Checklist:

## Rationale

It's annoying having to reload every time I changed an HTML or JS file


Andrew-Chen-Wang:
Good catch @cachitas It was just added in 0.15.0

PR #3358, opened 2021-10-13
Add django-admin startapp template

### LINKED ISSUE #881: Canonical location for local apps
Cookiecutter-django defines apps inside project folder, i.e.: red_dwarf/red_dwarf/users. I tried following the convention, but Django complains that it can't find my module, e.g.:
```
md red_dwarf/posts
python manage.py startapp red_dwarf/posts
```
config/settings/common.py
```
LOCAL_APPS = (
    'ref_dwarf.posts.apps.PostsConfig',
)
```
**ImportError: No module named 'posts'**

I'd have to adjust `sys.path` to include the APPS_DIR, but then Django complains, burns, and crashes with `Model class [...] doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS`

So my question is, what is the recommended location for local apps? I've red the docs back to back, gh issues, and the internets but could not find a meaningful answer.

### LINKED ISSUE #1505: How to properly generate startapp in cookiecutter
I've just successfully cloned this repo and everything works fine. My only question is how to generate the properly the startapp.

Normal way:
./manage.py startapp <APP NAME>  

Is it the same way when you use cookiecutter as the boilerplate?

### LINKED ISSUE #1725: New apps created are located in the project root, not inside the project slug
## What happened?
When creating a new app using `python manage.py startapp <name-of-the-app>`, the application is created within the project root

The application should be created within the project slug, as indicated [here](https://github.com/pydanny/cookiecutter-django/blob/6865f5b21fb8bf522bc12237ba17323999f803f0/%7B%7Bcookiecutter.project_slug%7D%7D/manage.py#L26)

## Steps to reproduce
- Create a bare new application using _cookiecutter-django_
- Install all requirements into your virtualenvironment using `pip install -r requirements/local.txt`
- Try to create a new app (with the name foo, for instance) using `python manage.py startapp foo`

### LINKED ISSUE #1876: Typical place to create app in cookiecutter-django
I have a question.
Startapp creates the app in the first level, the same directory where manage.py is. 
Why not create the app in the <project_name>/ level, where also the users app is?

dalrrard:
## Description

django-admin has a --template flag that lets you specify a directory to
base new apps off of. I've added a `startapp_template` directory to the
top level directory, updated the post gen hook to remove any files that
may not be needed, and updated the documentation for this new process.

Checklist:

## Rationale

This lets people semi-automatically create new apps without doing a lot of manual file moving and editing. 

Fix #1876
Fix #1725 
Fix #1505
Fix #881


Manjit2003:
Please approve this :(

browniebroke:
I'm a bit on the edge with this: while it's solving the problem, I find it a bit odd to have an new app template in the repo. After all, the startapp command accepts an URL as template. But maybe having it in the repo means that it can evolve with each person taste...

cjauvin:
I use cookiecutter-django with enough time passing between my generation runs that every single time I find myself struggling with the issue of app creation. Very recently it happened again, and even though you added some extra instructions in the manual to deal with this issue:

https://cookiecutter-django.readthedocs.io/en/latest/2-local-development/developing-locally.html#creating-your-first-django-app

the thing is that you only added it in the local dev section:

https://cookiecutter-django.readthedocs.io/en/latest/2-local-development/developing-locally.html

but not in the local dev with docker section:

https://cookiecutter-django.readthedocs.io/en/latest/2-local-development/developing-locally-docker.html 

and in my mind at least, these two sections are optional, not sequential (so I only read the docker one, and missed those instructions, and was confused again).

So I started to think how I could try to solve this by proposing a PR with some automated post gen hook with the `startapp` command, but I came here first and saw this still open PR.

So as an experienced but casual Django developer, I would say this is definitely a pain point that should be addressed in this otherwise excellent and tasty cookiecutter recipe.

luzfcb:
There is a discussion about possible changes to django startapp and maybe this is a good place to contribute and maybe make the django app creation API easier to extend https://github.com/django/deps/pull/98

PR #3366, opened 2021-10-17
Update README.rst

0sahilahmed:
Punctuation Changes

## Description

Checklist:


browniebroke:
Thanks for the suggestion, I've updated the case in the entire file

PR #3379, opened 2021-11-03
change path in docs Makefile to use APP variable

d57montes:
## Description

Checklist:

## Rationale

Changes the docs Makefile to use the APP variable for the apidocs target (for best practice).

### DIFF
--- a/{{cookiecutter.project_slug}}/docs/Makefile
+++ b/{{cookiecutter.project_slug}}/docs/Makefile
@@ -29,9 +29,9 @@ livehtml:
 # Outputs rst files from django application code
 apidocs:
 	{%- if cookiecutter.use_docker == 'y' %}
-	sphinx-apidoc -o $(SOURCEDIR)/api /app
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/docs/Makefile
```diff
@@ -29,9 +29,9 @@ livehtml:
 # Outputs rst files from django application code
 apidocs:
 	{%- if cookiecutter.use_docker == 'y' %}
-	sphinx-apidoc -o $(SOURCEDIR)/api /app
+	sphinx-apidoc -o $(SOURCEDIR)/api $(APP)
 	{%- else %}
-	sphinx-apidoc -o $(SOURCEDIR)/api ../{{cookiecutter.project_slug}}
+	sphinx-apidoc -o $(SOURCEDIR)/api $(APP)
 	{%- endif %}
```
browniebroke, line 35: Looks like these 2 lines are the same now and we no longer need the if/else?

browniebroke (approved):

PR #3385, opened 2021-11-08
refactor: remove user API methods parameter

danihodovic:
`methods` defaults to ["GET"] if not specified. There is no need to
explicitly set this in the action decorator.

https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/decorators.py#L145

```
    :param methods: A list of HTTP method names this action responds to.
                    Defaults to GET only.
```


browniebroke (approved):

browniebroke:
Thanks 🎉

PR #3330, opened 2021-09-28
fix: Switched to node:10 base image.

RobRoseKnows:
## Description

- Fixes #3202
- Does make builds larger, there might be a better solution.

Checklist:

- [ X ] I've made sure that `tests/test_cookiecutter_generation.py` is updated accordingly (especially if adding or updating a template option)
- [ X ] I've updated the documentation or confirm that my change doesn't require any updates

## Rationale

Currently a fresh local build of the cookiecutter using the node builder and the docker-compose will fail. This should put it in working order. Though there may be a better solution long term in order to make build sizes and times smaller.

Issue #3202, opened 2021-05-27
Build failed for node service on Docker + mac with M1 chip


EVENFATE:
```
[+] Building 59.1s (9/9) FINISHED                                                                                                                         
 => [internal] load build definition from Dockerfile                                                                                                 0.0s
 => => transferring dockerfile: 191B                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                    0.0s
 => => transferring context: 34B                                                                                                                     0.0s
 => [internal] load metadata for docker.io/library/node:10-stretch-slim                                                                              2.7s
 => [auth] library/node:pull token for registry-1.docker.io                                                                                          0.0s
 => [1/4] FROM docker.io/library/node:10-stretch-slim@sha256:88932859e3d022d79161b99628c4c2c50e836437455e2d1b1a008d98367b10d6                        0.0s
 => [internal] load build context                                                                                                                    0.1s
 => => transferring context: 34B                                                                                                                     0.0s
 => CACHED [2/4] WORKDIR /app                                                                                                                        0.0s
 => CACHED [3/4] COPY ./package.json /app                                                                                                            0.0s
 => ERROR [4/4] RUN npm install && npm cache clean --force                                                                                          56.3s
------                                                                                                                                                    
 > [4/4] RUN npm install && npm cache clean --force:                                                                                                      
#7 9.972 npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.                        
#7 13.51 npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142                              
#7 20.89 npm WARN deprecated debug@4.1.1: Debug versions >=3.2.0 <3.2.7 || >=4 <4.3.1 have a low-severity ReDos regression when used in a Node.js environment. It is recommended you upgrade to 3.2.7 or 4.3.1. (https://github.com/visionmedia/debug/issues/797)                                                   
#7 21.45 npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
#7 26.42 npm WARN deprecated har-validator@5.1.5: this library is no longer supported
#7 41.85 npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
[...]

PR #3400, opened 2021-11-13
Upgrade JS dependencies and upgrade to node 16

browniebroke:
## Description

Upgrade our JS dependencies to more recent versions and upgrade node version as many of them dropped support for node 10.

Checklist:

## Rationale

Our JS deps haven't been updated in a while and we're node 10 that we're using is no longer supported.

Closes #3336 
Closes #3330 
Ref #3202


browniebroke:
I've done a good amount of testing and all seems fine as far as I can tell.

PR #3401, opened 2021-11-13
Add docker, pip and npm to GitHub's Dependabot

mpoli:
## Description

Add GitHub's Dependabot coverage of Dockerfiles, NPM and Pip dependencies.

Checklist:

## Rationale

When using GitHub as the CI tool, automatic access to [Dependabot](https://github.com/dependabot) is available.

cookiecutter-django already has the .github/dependabot.yml in place, but it is currently limited to updating GitHub Actions.

Incrementing its scope, adding coverage of all Dockerfiles, of NPM's packages.json and of Pip's [runtime.txt,requirements.txt,requirements/*.txt] is a nice addition to whoever is already using GitHub as the CI tool of choice.

Dependabot generates quite some action with only cookiecutter-django's installed dependencies. A choice might be asking for it in the template, in case Dependabot requires too much attention.


{{cookiecutter.project_slug}}/.github/dependabot.yml
```diff
@@ -1,7 +1,90 @@
...
     directory: "/"
+    # Check for updates to GitHub Actions every weekday
+    schedule:
+      interval: "daily"
+
+{%- if cookiecutter.use_docker == 'y' %}
+
+  # Enable version updates for Docker
+  # As there is no way to use wildcards, each dockerfile has its own entry
+  # Check https://github.com/dependabot/dependabot-core/issues/2178 for updates to that...
```
browniebroke, line 17: I think this comment might need a bit more clarification: one might understand it as missing support for this `compose/*/*/` and might change the `directory` to `compose/`, thinking that recursive search would work - it appears  to be working fine for `pip` - although it doesn't in the case of Docker. A bit annoying that we have to repeat all this, but it's sadly out of our control...

I think we should mention that it's not only about wildcards, but also about not being recursive. 

```suggestion
  # We need to specify each Dockerfile in a separate entry because Dependabot doesn't 
  # support widlcards or recursively checking subdirectories. Check this issue for updates:
  # https://github.com/dependabot/dependabot-core/issues/2178
```

browniebroke (approved):
Looking good, I made a small suggestion to clarify a comment as I had to try it myself to be sure.

mpoli:
An immediate concern is that Dependabot will PR "Bump node from 10-stretch-slim to 17-stretch-slim in /compose/production/django" and "Bump node from 10-stretch-slim to 17-stretch-slim in /compose/local/node" which isn't ideal. We might want to pull https://github.com/cookiecutter/cookiecutter-django/pull/3400 or even upgrade to 17-bullseye-slim to avoid that PRs right from the start.

mpoli:
It will also try to upgrade from the beginning gulp-rename, gulp-uglify-es, gulp-sass, gulp-postcss, cssnano and the postgres image to 14.

browniebroke:
Yeah, once #3400 is merged, we should have a lot less noise. I'll probably merge this after. I want to wait a couple of days to get a bit of feedback on #3400, but from my testing it seems to work fine.

browniebroke:
Thanks!

PR #3415, opened 2021-11-17
Switch template to calendar versioning & automate releases

### LINKED ISSUE #3014: Switch template to calendar versioning
## Description

The template is currently versioned after the version of Django it uses:

https://github.com/pydanny/cookiecutter-django/blob/b48067c03494993933bd2eb0aa2e5afdf367f12f/setup.py#L11-L13

I suggest changing this to calendar versioning instead e.g. 2021.1.24

## Rationale

With Github actions, we've been able to automate a few things around the project management:

1. [Drafting Github releases](https://github.com/pydanny/cookiecutter-django/releases) thanks to [release drafter
](https://github.com/pydanny/cookiecutter-django/blob/master/.github/workflows/draft-release.yml)
2. Generating Changelog [with a custom script](https://github.com/pydanny/cookiecutter-django/blob/master/scripts/update_changelog.py) that we [run every night](https://github.com/pydanny/cookiecutter-django/blob/master/.github/workflows/update-changelog.yml).

However, this approach has a few issues:

- Changelog file and release changes are not consistent, they 2 different views of the same thing.
- The Github release doesn't tell much about when changes were actually usable, one typically uses the latest in `master` when calling the cookiecutter CLI.
- The changelog script might need some maintenance down the road, would be nicer to rely on a community supported Github action.
- Changing the version template is not automated yet, and we often miss the step when the version of Django is updated.

## Solution

Ideally we would use off the shelf action(s) like:

- https://github.com/fregante/daily-version-action
- https://github.com/notlmn/release-with-changelog

These are providing some building pieces. Ideally, we also need one to keep the changelog file up to date.

Would be nice to find an all in one solution, an equivalent of https://github.com/relekang/python-semantic-release, but for calendar base releases.

This process would run like the current update changelog script, every night on schedule.

browniebroke:
## Description

As suggested in #3014, change versioning to calendar versioning and automate release process. To do so, this extends the script currently updating the changelog to keep the version from `setup.py` in line & publich a new GitHub release.

Checklist:

## Rationale

As explained in the linked issue, our changelog update is not consistent with our release & versioning. We update the changelog every night with what was merged during the previous day. On the other hand, we set the version of the template to the Django version used in the generated project and we update the GitHub releases accordingly. 

When the Django version is updated, a maintainer needs to update the version of the template, and publish the draft release, which is often overlooked.

Fix #3014 

## Example:

See it in action on my fork:

- Commit: https://github.com/browniebroke/cookiecutter-django/commit/3b6839ed0ca46040291daa3f2eb7fdef485063c6
- Release: https://github.com/browniebroke/cookiecutter-django/releases/tag/2021.11.16


luzfcb (approved):
LGTM. 

Note: The workflow creates a commit on the `main` branch, the current configuration ends up causing the `main` branch of a fork to diverge from our `main` branch, which can hinder the contribution flow.

I think all workflows that create commits should be disabled in forks.

We can achieve that by including in our config something like:


```yml
jobs:
  build:
    # Disables this workflow from running in a repository that is not part of the indicated organization/user
    if: github.repository_owner == 'cookiecutter'
```

scripts/update_changelog.py
```diff
@@ -29,26 +34,44 @@ def main() -> None:
...
+    release_changes_summary = generate_md(grouped_pulls)
+    print(f"Summary of changes: {release_changes_summary}")
 
     # Update CHANGELOG.md file
-    file_path = ROOT / "CHANGELOG.md"
-    old_content = file_path.read_text()
-    updated_content = old_content.replace(
-        "<!-- GENERATOR_PLACEHOLDER -->",
-        f"<!-- GENERATOR_PLACEHOLDER -->\n\n{rendered_content}",
+    release = f"{merged_date:%Y.%m.%d}"
```
browniebroke, line 41: That bit assumes that we do no more than 1 release a day.

scripts/update_changelog.py
```diff
@@ -77,12 +100,59 @@ def generate_md(grouped_pulls: dict[str, list[github.PullRequest.PullRequest]])
...
+
+    repo.git.commit(
+        m=message,
+        author=f"{user} <{email}>",
+    )
+    repo.git.tag("-a", release, m=message)
+    server = f"https://{GITHUB_TOKEN}@github.com/{GITHUB_REPO}.git"
+    print(f"Pushing changes to {GIT_BRANCH} branch of {GITHUB_REPO}")
+    repo.git.push(server, GIT_BRANCH)
+    repo.git.push("--tags", server, GIT_BRANCH)
```
browniebroke, line 146: That function has been heavily inspired by [how `python-semantic-release` does it](https://github.com/relekang/python-semantic-release/blob/master/semantic_release/vcs_helpers.py).

setup.py
```diff
@@ -1,21 +1,11 @@
...
-import sys
-
 try:
     from setuptools import setup
 except ImportError:
     from distutils.core import setup
 
-# Our version ALWAYS matches the version of Django we support
-# If Django has a new release, we branch, tag, then update this setting after the tag.
-version = "3.1.13"
```
browniebroke, line 13: We should probably make a last release with the old versioning before merging this

browniebroke:
> Note: The workflow creates a commit on the `main` branch, the current configuration ends up causing the `main` branch of a fork to diverge from our `main` branch, which can hinder the contribution flow.

That's an excellent point, I also noticed it and wanted to improve it

> We can achieve that by including in our config something like:

Excellent suggestion, I'll implement it

PR #3418, opened 2021-11-17
Add post-hook Bootstrap version upgrader

### LINKED ISSUE #3412: PR update Bootstrap in HTML
## Description

Update the links and version of bootstrap

## Rationale

So everytime I use cookiecutter, I don't get an older version and need to check if there's a newer one

Andrew-Chen-Wang:
Signed-off-by: Andrew-Chen-Wang <acwangpython@gmail.com>

## Description

Fixes #3412

TL;DR post-hook that gets latest bootstrap (not including pre-releases). Maybe use GitHub action instead?

Add a user-side Bootstrap version upgrader in the post-hook. However, after writing this, it might actual be better to make a PR here rather than make a post-hook (this method could slow down template generation; not much, but could).

This PR also doesn't resolve that, if you're in a place with no internet, you'll be stuck with an older Bootstrap version. Though, I suppose most wouldn't be offline (though it's accounted for).

However, by making a post-hook, regardless of the template (if you haven't downloaded the latest), then you'll be happy to have an automated bootstrap versioner.

Checklist:

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -9,10 +9,17 @@ TODO: ? restrict Cookiecutter Django project initialization to Python 3.x enviro
 """
 from __future__ import print_function
 
+import json
 import os
 import random
 import shutil
 import string
+
+try:
+    from urllib.error import HTTPError, URLError
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import HTTPError, URLError, urlopen
 
 try:
     # Inspired by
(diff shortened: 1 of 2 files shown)


hooks/post_gen_project.py
```diff
@@ -287,6 +298,67 @@ def set_flags_in_settings_files():
...
+def set_bootstrap_version():
+    add_css = "{{ cookiecutter.custom_bootstrap_compilation }}".lower() == "n"
+    add_js = not (
+        "{{ cookiecutter.custom_bootstrap_compilation }}".lower() == "y"
+        and "{{ cookiecutter.js_task_runner }}".lower() == "Gulp"
+    )
+    if not (add_css or add_js):
+        return
+    try:
+        r = urlopen("https://api.cdnjs.com/libraries/bootstrap?fields=versions")
```
browniebroke, line 330: I'm not sure how I feel about making HTTP requests in the post generation hook, where we don't control the environment very well... Especially since we're still supporting Python 2 (reminds me that it might be time to drop it maybe). Also this might upgrade to a Bootstrap version we don't support yet? e.g. what happens when Bootstrap 6 is out and we're not yet ready?

As a side note, I'm trying to get Dependabot setup on the repo to work for npm, it seems to be working on my fork. 

Once this is done, I'm thinking we could tweak the approach here to :

- have the parts we need from this HTTP call (version & SRI) saved in a JSON file in the repo
- update that JSON file using GH action if a PR is opened/updated and looks like a Dependabot PR for Bootstrap
- use that JSON file in the post gen hook to replace the bits of HTML like you do currently

Not sure if we actually need a JSON file, maybe we could stick to the constants you have now or even leave the script tags in the HTML and replace them. My main idea would be to keep the version from `package.json` and the HTML in line, and update them together in a single PR

browniebroke:
What do you think @Andrew-Chen-Wang of that above suggestion? I've set up Dependanbot in #3436 and here is an example of PR we should look out for: https://github.com/browniebroke/cookiecutter-django/pull/28

Andrew-Chen-Wang:
Yes using the bootstrap dependabot PR would be more suitable!

PR #3426, opened 2021-11-20
Bump pre-commit/action to 2.0.3

mpoli:
Bump pre-commit/action to 2.0.3 so Dependabot will not PR this right after first push to GitHub.

## Description

Simply update pre-commit/action to 2.0.3.

Checklist:

## Rationale

Dependabot is already creating a PR to Bump pre-commit/action from version 2.0.0 to version 2.0.3 as soon as a project created with cookiecutter-django is pushed to GitHub.

This should also be committed in preparation for #3401.


browniebroke:
Thanks 🎉

PR #3433, opened 2021-11-22
Setup pre-commit for the template files

browniebroke:
## Description

We are already adding pre-commit to the generated projects, but we're not using them in the template itself. We cannot run python tools on the files that contain Jinja templates, so for now I exclude them for simplicity. We're already running pre-commit when testing the generated projects, so that should be enough.

Later, I would like is move some of our checks to pre-commit CI. I've set [it up on my fork](https://results.pre-commit.ci/run/github/99044898/1637618757.PmUO3L55QFW71aDCt9c0vQ) and it's really fast.

Checklist:

## Rationale

I think this will make the experience of contributing changes to the template a bit easier by reducing the feedback loop. I often forget to run `black` or `flake8` before pushing and having them run on save isn't suitable for most of the project files.


Andrew-Chen-Wang:
for templated Python files, what do you think about simply removing all the Jinja template tags? Idk how pyupgrade does it, but we could try to make something similar? It'll just be difficult to keep track of where the jinja templates originally were.

browniebroke:
> for templated Python files, what do you think about simply removing all the Jinja template tags? 

I haven't looked in details but I think that it sounds like a rabbit hole 😄 I assume doing that might result in invalid Python code in some cases (although I don't have a clear example).

Even if that's not a problem, I'm really unsure how to hook into the the pre-commit lifcycle. Maybe I'm missing a feature from pre-commit. I imagine we would need to do something like:

1. remove jinja tags
2. run pre-commit hooks
3. revert step 1
4. actually commit

At first glance, it sounds complicated...

Andrew-Chen-Wang:
> I haven't looked in details but I think that it sounds like a rabbit hole 😄 I assume doing that might result in invalid Python code in some cases (although I don't have a clear example)

Here's one:

```python
{% if cookiecutter %}
if blah:
{% else %}
if blah + blah:
{% endif %}
    ...
```
Yea it might be a rabbit hole 😅 

Another thought was to chunk the code somehow, but I don't have any other thoughts. Pretty difficult issue.

PR #3434, opened 2021-11-24
Cache Python dependencies on our CI

browniebroke:
## Description

Add caching of pip dependencies using [the solution](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/) provided by the official GitHub action `setup-python`.

Let's try it on the template first. Once we get it right, do the same in the generated projects.

## Rationale

Faster builds on CI


luzfcb (approved):

luzfcb:
Awesome 🎉. Thank you @browniebroke

PR #3435, opened 2021-11-24
Fix pre-commit config

### LINKED ISSUE #3252: pre-commit config file excluding files that are not supposed to be excluded
## What happened?
In file: ` cookiecutter-django/{{cookiecutter.project_slug}}/.pre-commit-config.yaml ` line number 1:
`exclude: 'docs|node_modules|migrations|.git|.tox'` using this statement it excludes any file/directory with word 'docs' in it, for example it will exclude:
* ml_docs/
* docs_serializer.py
* docs/
* docs.py
instead it should only exclude docs/
## Additional details
I tried changing the exclude to: `exclude: '^docs$|node_modules|migrations|.git|.tox'`
but it didn't work either

* Host system configuration: 
    * Version of cookiecutter CLI (get it with `cookiecutter --version`):
    * OS name and version:

      On Linux, run 
      ```bash
      lsb_release -a 2> /dev/null || cat /etc/redhat-release 2> /dev/null || cat /etc/*-release 2> /dev/null || cat /etc/issue 2> /dev/null
      ```

      On MacOs, run
      ```bash
      sw_vers
      ```
      
      On Windows, via CMD, run
      ```
      systeminfo | findstr /B /C:"OS Name" /C:"OS Version"
      ```

      ```bash
      # Insert here the OS name and version
      
      ```

    * Python version, run `python3 -V`:
    * Docker version (if using Docker), run `docker --version`:
    * docker-compose version (if using Docker), run `docker-compose --version`:
    * ...
* Options selected and/or [replay file](https://cookiecutter.readthedocs.io/en/latest/advanced/replay.html):
  On Linux and MacOS: `cat ${HOME}/.cookiecutter_replay/cookiecutter-django.json`
  (Please, take care to remove sensitive information)
    ```json
    # Insert here the replay file content
    ```
<summary>
Logs:
<details>
<pre>
$ cookiecutter https://github.com/pydanny/cookiecutter-django
project_name [Project Name]: ...
</pre>
</details>
</summary>

browniebroke:
## Description

Adjust pre-commit config to avoid ignoring files that should be considered and fix a few issues that may appear from the vanilla project.

## Rationale

Our pre-commit config starts with an exclude pattern which is a bit too broad:

- We just want to exclude the tope level `docs/` folder, not any file which contains "docs" in its name
- We don't need to exclude files already ignored by git
- Remove `fail_fast` as it's better to get all issues in one go, the hooks aren't that slow to run.

Fix #3252


{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -1,6 +1,5 @@
-exclude: "docs|node_modules|migrations|.git|.tox"
```
luzfcb, line 1: pre-commit excludes `node_modules`, `.git` and `.tox` by default?

luzfcb: reading description again I saw that apparently pre-commit respects .gitignore

browniebroke: From my testing, it does. It should only check what's tracked in `.git` and these are already in our `.gitignore`.

browniebroke: Do you want to take another look at this? I'm planning to merge it in a next couple of days if there is no objection

PR #3436, opened 2021-11-24
Configure Dependabot for npm packages at the template level

browniebroke:
## Description

Changes to get Dependanbot updates for npm packages:

- Update Dependabot config
- Remove all the Jinja tags from the `package.json` in order to make it parsable by Dependanbot
- Remove  unwanted dependencies in post-generation hook, to replace what was done by templating

Checklist:

## Rationale

We can't keep npm packages up to date because they are not automatically updated.


browniebroke:
Let's give this one a try.

PR #3462, opened 2021-12-10
Add a PyCharm run configuration for docker-compose

browniebroke:
## Description

Add 2 run configurations for running Django and running the docs.

Checklist:

## Rationale

We already have a few run configurations but it would be nice to also provide one for docker-compose setups.

### DIFF
--- a/{{cookiecutter.project_slug}}/.idea/runConfigurations/docker_compose_up_docs.xml
+++ b/{{cookiecutter.project_slug}}/.idea/runConfigurations/docker_compose_up_docs.xml
@@ -1,12 +1,11 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration name="docker-compose up" type="docker-deploy" factoryName="docker-compose.yml" server-name="Docker">
+  <configuration name="docker-compose up django" type="docker-deploy" factoryName="docker-compose.yml" server-name="Docker">
     <deployment type="docker-compose.yml">
       <settings>
         <option name="envFilePath" value=""/>
         <option name="services">
           <list>
-            <option value="django"/>
-            <option value="node"/>
+            <option value="docs"/>
           </list>
         </option>
         <option name="sourceFilePath" value="local.yml"/>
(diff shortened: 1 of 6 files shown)


lancelote:
@browniebroke Out of interest - why separate `docker-compose up` configuration if one can use [runserver.xml](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.idea/runConfigurations/runserver.xml) with a docker-compose-based interpreter? It is not possible to run a Docker-type configuration under the Python debugger though it is possible to debug `runserver.xml` configuration with a docker-compose-based interpreter.

{{cookiecutter.project_slug}}/.idea/runConfigurations/docker_compose_up_docs.xml
```diff
@@ -0,0 +1,16 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration name="docker-compose up django" type="docker-deploy" factoryName="docker-compose.yml" server-name="Docker">
```
lancelote, line 2:
```suggestion
  <configuration name="docker-compose up docs" type="docker-deploy" factoryName="docker-compose.yml" server-name="Docker">
```

browniebroke:
> @browniebroke Out of interest - why separate `docker-compose up` configuration if one can use [runserver.xml](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.idea/runConfigurations/runserver.xml?rgh-link-date=2021-12-14T12%3A24%3A51Z) with a docker-compose-based interpreter? It is not possible to run a Docker-type configuration under the Python debugger though it is possible to debug `runserver.xml` configuration with a docker-compose-based interpreter.

Good question. The difference is that the existing `runserver` config would only be starting the `django` container, while this new config also starts other useful services, celery and node, if applicable. 

I see the 2 run configuration as complementary: if you want to do some Python debugging, you should use the runserver config, but if you want to start the whole stack without debugging, the docker-compose configuration would start everything with one button.

PR #3488, opened 2021-12-24
Simplify `TEMPLATES` settings with `APP_DIRS=True`

### LINKED ISSUE #3475: django-debug-toolbar 3.2.3 introduced APPS_DIR set to True warning
## What happened?

https://github.com/cookiecutter/cookiecutter-django/pull/3466 bumped django-debug-toolbar to version 3.2.3. This version introducted the following WARNING into Django, which persisted in 3.2.4:

````
django      | System check identified some issues:
django      |
django      | WARNINGS:
django      | ?: (debug_toolbar.W006) At least one DjangoTemplates TEMPLATES configuration needs to have APP_DIRS set to True.
django      |   HINT: Use APP_DIRS=True for at least one django.template.backends.django.DjangoTemplates backend configuration.
````

No warning should have been shown.

## Additional details

The APPS_DIR situation also surfaces when implementing some packages like django-sitemaps, that  also tells that it is necessary to have at least one APPS_DIR=True somewhere in the templating options ([check the docs](https://docs.djangoproject.com/en/4.0/ref/contrib/sitemaps/#installation)).

cookiecutter-django's current approach seems to accomplish similar results by using template-dir by template-dir approarch instead of the APPS_DIR=True catchall. That might be a good thing, but somehow at least in two instances (one part of cookiecutter-django and one that seems to be part of the Django Project itself) are complaining about this.

So the question is: is there a workaround that warning, and a way to circumvent django-sitemaps instructions and keep the current templating approach? And is there any downside to having it instead of the APPS_DIR=True solution?

* Host system configuration:
    * Version of cookiecutter: Cookiecutter 1.7.3
    * OS name and version: docker
    * Python version: 3.10
    * Docker version: 20.10.10
    * docker-compose version: 1.29.2
    * cookiecutter-django version: **2021.12.15**
    *  django-debug-toolbar: **3.2.4**

browniebroke:
## Description

- Update test script to fail if the `check` command flags any warnings
- Simplify the `TEMPLATES` settings as explained [here](https://github.com/cookiecutter/cookiecutter-django/issues/3475#issuecomment-996186254) 

Checklist:

## Rationale

Our template settings are configuring the `loaders` option explicitely, and this option is optimised in the production/test setttings to wrap the `cached.Loader`, which duplciates this list of loaders in 3 places.

However, it seems that our config can be replicated by setting `APP_DIRS=True`:

- It has the same list of loaders when `DEBUG=True`
- When `DEBUG=False`, the list is wrapped with `cached.Loader`

Fixes #3475

### DIFF
--- a/tests/test_bare.sh
+++ b/tests/test_bare.sh
@@ -29,6 +29,9 @@ pre-commit run --show-diff-on-failure -a
 # run the project's tests
 pytest
 
+# Make sure the check doesn't raise any warnings
+python manage.py check --fail-level WARNING
+
 if [ -f "package.json" ]
 then
     npm install
--- a/tests/test_docker.sh
+++ b/tests/test_docker.sh
@@ -35,3 +35,6 @@ docker-compose -f local.yml run django python manage.py makemigrations --dry-run
 
 # Test support for translations
 docker-compose -f local.yml run django python manage.py makemessages --all
+
+# Make sure the check doesn't raise any warnings
+docker-compose -f local.yml run django python manage.py check --fail-level WARNING
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -182,15 +182,11 @@ TEMPLATES = [
     {
         # https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-TEMPLATES-BACKEND
         "BACKEND": "django.template.backends.django.DjangoTemplates",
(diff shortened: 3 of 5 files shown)


cjauvin:
I tested your fix by simply commenting out the `loaders` item of the `TEMPLATES` dict of my already existing app, which was created using a recent version of this cookiecutter (about a week ago), and it seems that the toolbar warning is still present. Any idea what I might be doing wrong?

cjauvin:
Ohhh sorry about that, my bad! I forgot to add ""APP_DIRS: True"`.. now it works, many thanks!

PR #3515, opened 2022-01-05
Add UserSignupForm and UserSocialSignupForm

### LINKED ISSUE #450: Adding fields to the User model
Please excuse me if the answer to my question is painfully obvious but I'm new to Django and I'm not very familiar with custom user models.

If I wanted to add another field to the user model (for example, "department" - the users are employees), where would I add it?

I figured I could add a `department` variable to `models.py` but it doesn't seem to work. When I login to the `admin` site, I don't see a "department" field when I add a user. Similarly, I don't see a "name" field in the `admin` site - I only see `First Name`, `Last Name`, and `Email Address`. 

```
# models.py
# -*- coding: utf-8 -*-
from __future__ import unicode_literals, absolute_import

from django.contrib.auth.models import AbstractUser
from django.core.urlresolvers import reverse
from django.db import models
from django.utils.encoding import python_2_unicode_compatible


@python_2_unicode_compatible
class User(AbstractUser):

    # First Name and Last Name do not cover name patterns
    # around the globe.
    name = models.CharField(blank=True, max_length=255)
    department = models.CharField(blank=True, max_length=5)

    def __str__(self):
        return self.username

    def get_absolute_url(self):
        return reverse('users:detail', kwargs={'username': self.username})
```

### LINKED ISSUE #2708: allauth sign up doesn't use custom UserCreation form
## What happened?

I wanted to include some custom fields (django-recaptcha) to my UserCreation form. Turns out, the settings for allauth wasn't pointed to UserCreationForm.

Allauth should be configured to use the custom UserCreationForm.

## Steps to reproduce

1. Generate a new project.
2. Add a random field to the form.
3. Go to signup. It won't show up.
4. Add the setting:

```
ACCOUNT_FORMS = {"signup": "donate_anything.users.forms.UserCreationForm"}
```

Voila the form field is there. I think the creation of a user should be parallel regardless of where it happens (admin or user input). Taking a look at AbstractUser model, which the custom user model uses, the `email` attribute is set with `blank=True`, meaning it's not required. However, the allauth setting for email is set to `mandatory`. I think we should add that setting (and any other necessary configuration) for the custom user creation form.

Was this deliberate? If not, I'll make a PR.

mpoli:
Add UserSignupForm and UserSocialSignupForm
And docstrings clarifying how to use Custom User Model and Signup Forms.

## Description

The proposal is based on the discussion in issue #2708, and how hard it is to currently understand/use the Custom User Model shipper with cookiecutter-django. 

Creating the stub UserSignupForm and UserSocialSignupForm will certainly save hours (if not more) of research and trial-and-error for users wanting to expand the default User class.

Checklist:

## Rationale

The proposal is to ship the explicit Form Classes that can be used to expand the Signup Process and clarify the class that concerns the Admin User Creation process that is already present.

Fix #2708
Fix #450

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -306,8 +306,12 @@ ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_EMAIL_VERIFICATION = "mandatory"
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
 ACCOUNT_ADAPTER = "{{cookiecutter.project_slug}}.users.adapters.AccountAdapter"
+# https://django-allauth.readthedocs.io/en/latest/forms.html
+ACCOUNT_FORMS = {"signup": "{{cookiecutter.project_slug}}.users.forms.UserSignupForm"}
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
 SOCIALACCOUNT_ADAPTER = "{{cookiecutter.project_slug}}.users.adapters.SocialAccountAdapter"
+# https://django-allauth.readthedocs.io/en/latest/forms.html
+SOCIALACCOUNT_FORMS = {"signup": "{{cookiecutter.project_slug}}.users.forms.UserSocialSignupForm"}
 {% if cookiecutter.use_compressor == 'y' -%}
 # django-compressor
 # ------------------------------------------------------------------------------
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
@@ -17,6 +17,7 @@ class UserCreationForm(admin_forms.UserCreationForm):
     Form for User Creation in the Admin Area.
     To change user signup, see UserSignupForm and UserSocialSignupForm.
     """
+
     class Meta(admin_forms.UserCreationForm.Meta):
         model = User
 
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
@@ -1,5 +1,7 @@
 from django.contrib.auth import forms as admin_forms
 from django.contrib.auth import get_user_model
+from allauth.account.forms import SignupForm
+from allauth.socialaccount.forms import SignupForm as SocialSignupForm
 from django.utils.translation import gettext_lazy as _
 
 User = get_user_model()
@@ -11,9 +13,56 @@ class UserChangeForm(admin_forms.UserChangeForm):
 
 
 class UserCreationForm(admin_forms.UserCreationForm):
+    """
+    Form for User Creation in the Admin Area.  
+    To change user signup, see UserSignupForm and UserSocialSignupForm.
+    """
     class Meta(admin_forms.UserCreationForm.Meta):
         model = User
 
         error_messages = {
             "username": {"unique": _("This username has already been taken.")}
         }
+
+class UserSignupForm(SignupForm):
+    """
+    Form that will be rendered on a user sign up section/screen.
+    Default fields will be added automatically.
+    Check UserSocialSignupForm for accounts created from social.
+    """
+
+    def save(self, request):
+
+        # Ensure you call the parent class's save.
+        # .save() returns a User object.
+        user = super(UserSignupForm, self).save(request)
+
+        # Add your own processing here.
+
+        # You must return the original result.
+        return user
+
+
+    class Meta:
+        model = User
+    
+class UserSocialSignupForm(SocialSignupForm):
+    """
+    Renders the form when user has signed up using social accounts.
+    Default fields will be added automatically.
+    See UserSignupForm otherwise.
+    """
+
+    def save(self, request):
+
+        # Ensure you call the parent class's save.
+        # .save() returns a User object.
+        user = super(UserSocialSignupForm, self).save(request)
+
+        # Add your own processing here.
+
+        # You must return the original result.
+        return user
+
+    class Meta:
+        model = User
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
@@ -7,12 +7,12 @@ from django.utils.translation import gettext_lazy as _
 User = get_user_model()
 
 
-class UserChangeForm(admin_forms.UserChangeForm):
+class UserAdminChangeForm(admin_forms.UserChangeForm):
     class Meta(admin_forms.UserChangeForm.Meta):
         model = User
 
 
-class UserCreationForm(admin_forms.UserCreationForm):
+class UserAdminCreationForm(admin_forms.UserCreationForm):
     """
     Form for User Creation in the Admin Area.
     To change user signup, see UserSignupForm and UserSocialSignupForm.
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
@@ -1,7 +1,7 @@
-from django.contrib.auth import forms as admin_forms
-from django.contrib.auth import get_user_model
 from allauth.account.forms import SignupForm
 from allauth.socialaccount.forms import SignupForm as SocialSignupForm
+from django.contrib.auth import forms as admin_forms
+from django.contrib.auth import get_user_model
 from django.utils.translation import gettext_lazy as _
 
 User = get_user_model()
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
@@ -14,7 +14,7 @@ class UserChangeForm(admin_forms.UserChangeForm):
 
 class UserCreationForm(admin_forms.UserCreationForm):
     """
-    Form for User Creation in the Admin Area.  
+    Form for User Creation in the Admin Area.
     To change user signup, see UserSignupForm and UserSocialSignupForm.
     """
     class Meta(admin_forms.UserCreationForm.Meta):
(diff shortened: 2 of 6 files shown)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -1,5 +1,7 @@
 from django.contrib.auth import forms as admin_forms
 from django.contrib.auth import get_user_model
+from allauth.account.forms import SignupForm
+from allauth.socialaccount.forms import SignupForm as SocialSignupForm
 from django.utils.translation import gettext_lazy as _
```
browniebroke, line 5: isort fails because these imports aren't sorted:

```suggestion
from allauth.account.forms import SignupForm
from allauth.socialaccount.forms import SignupForm as SocialSignupForm
from django.contrib.auth import forms as admin_forms
from django.contrib.auth import get_user_model
from django.utils.translation import gettext_lazy as _
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,56 @@ class Meta(admin_forms.UserChangeForm.Meta):
 
 
 class UserCreationForm(admin_forms.UserCreationForm):
+    """
+    Form for User Creation in the Admin Area.
```
browniebroke, line 17:
```suggestion
    Form for User Creation in the Admin Area.
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,56 @@ class Meta(admin_forms.UserChangeForm.Meta):
...
+        user = super(UserSignupForm, self).save(request)
+
+        # Add your own processing here.
+
+        # You must return the original result.
+        return user
+
+
+    class Meta:
+        model = User
```
browniebroke, line 47: A few comments here:

- I'm not convinced we should to override the save method, it's explaining something which is quite standard Python/Django: call the super implementation and return the same thing
- Styling is complaining about too many blank lines before `Meta`

```suggestion
class UserSignupForm(SignupForm):
    """
    Form that will be rendered on a user sign up section/screen.
    Default fields will be added automatically.
    Check UserSocialSignupForm for accounts created from social.
    """

    class Meta:
        model = User
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,56 @@ class Meta(admin_forms.UserChangeForm.Meta):
...
+        # .save() returns a User object.
+        user = super(UserSocialSignupForm, self).save(request)
+
+        # Add your own processing here.
+
+        # You must return the original result.
+        return user
+
+    class Meta:
+        model = User
```
browniebroke, line 68: Similar comments here + missing blank line at the end of the file

```suggestion
class UserSocialSignupForm(SocialSignupForm):
    """
    Renders the form when user has signed up using social accounts.
    Default fields will be added automatically.
    See UserSignupForm otherwise.
    """

    class Meta:
        model = User

```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,56 @@ class Meta(admin_forms.UserChangeForm.Meta):
...
+
+        # Add your own processing here.
+
+        # You must return the original result.
+        return user
+
+
+    class Meta:
+        model = User
+
```
browniebroke, line 48: Whitespaces in blank line
```suggestion

```

browniebroke (commented):
Made a couple of comments that should fix the style errors... I agree with @Andrew-Chen-Wang regarding renaming `UserChangeForm` and `UserCreationForm` to have admin in their name, e.g. `UserAdminChangeForm` and `UserAdminCreationForm`

Andrew-Chen-Wang (changes requested):
I have nothing else to add beside the name change for admin forms.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
mpoli: Agreed. I was just folowing the example from allauth documentation here. But I don't think we need to get to that level of detail/training wheels.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,56 @@ class Meta(admin_forms.UserChangeForm.Meta):
 
 
 class UserCreationForm(admin_forms.UserCreationForm):
```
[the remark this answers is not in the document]
mpoli: There are some mentions of that class that need to be altered in tandem in admin.py etc and documentation.

I will take a look in a while and commit them all together.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,32 @@ class Meta(admin_forms.UserChangeForm.Meta):
...
+        model = User
+class UserSocialSignupForm(SocialSignupForm):
+    """
+    Renders the form when user has signed up using social accounts.
+    Default fields will be added automatically.
+    See UserSignupForm otherwise.
+    """
+
+    class Meta:
+        model = User
```
browniebroke, line 44:
```suggestion


class UserSignupForm(SignupForm):
    """
    Form that will be rendered on a user sign up section/screen.
    Default fields will be added automatically.
    Check UserSocialSignupForm for accounts created from social.
    """

    class Meta:
        model = User


class UserSocialSignupForm(SocialSignupForm):
    """
    Renders the form when user has signed up using social accounts.
    Default fields will be added automatically.
    See UserSignupForm otherwise.
    """

    class Meta:
        model = User
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -11,9 +13,35 @@ class Meta(admin_forms.UserChangeForm.Meta):
 
 
 class UserCreationForm(admin_forms.UserCreationForm):
+    """
+    Form for User Creation in the Admin Area.
+    To change user signup, see UserSignupForm and UserSocialSignupForm.
+    """
```
browniebroke, line 19:
```suggestion
    """

```

mpoli:
So I have changed UserCreateForm and UserChangeForm to UserAdminCreateForm and UserAdminChangeForm and everything seems to be working without errors in user addition and editing in the admin interface.

But I was unable to add a new field to neither the UserAdminCreateForm nor UserAdminChangeForm that would show up in the admin area of a newly generated project. :(

I will keep on researching on the reason (or on how to make something show up in the panel properly).

mpoli:
I will write some on how to use Custom User Models in the documentation as well.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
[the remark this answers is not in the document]
mpoli: Solved by https://github.com/cookiecutter/cookiecutter-django/pull/3515/commits/14cdb4ad2799ce618986f7f575d11952dc2af4b9

mpoli:
Ok, some further reading and researching, and it works.

Converting from draft to ready and I will dedicate some time on writing documentation about the Custom User Model for some extra improvement in another PR.

Thanks for all the help, @browniebroke and @Andrew-Chen-Wang!

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
```diff
@@ -1,19 +1,48 @@
...
+
+class UserSignupForm(SignupForm):
+    """
+    Form that will be rendered on a user sign up section/screen.
+    Default fields will be added automatically.
+    Check UserSocialSignupForm for accounts created from social.
+    """
+
+    class Meta:
+        model = User
```
browniebroke, line 37: I noticed that the base class isn't a `ModelForm`, it seems to be a simple `Form`, so we probably don't need the nested `Meta` class actually.

{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -306,8 +306,12 @@
 ACCOUNT_EMAIL_VERIFICATION = "mandatory"
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
 ACCOUNT_ADAPTER = "{{cookiecutter.project_slug}}.users.adapters.AccountAdapter"
+# https://django-allauth.readthedocs.io/en/latest/forms.html
+ACCOUNT_FORMS = {"signup": "{{cookiecutter.project_slug}}.users.forms.UserSignupForm"}
```
browniebroke, line 310: While looking through this, I ended up looking at the allauth code and found out about [the setting](https://django-allauth.readthedocs.io/en/latest/configuration.html?highlight=ACCOUNT_SIGNUP_FORM_CLASS) `ACCOUNT_SIGNUP_FORM_CLASS`. If this setting is configured, the class is used as base class for both `SignupForm` (social and not social). 

It seems like we could configure this single setting instead of these 2. What do you think?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/forms.py
mpoli: I see. Agreed.

mpoli: Done

{{cookiecutter.project_slug}}/config/settings/base.py
mpoli: Isn't it possible that some of the information that will fill up the profile of a new user will come from social information? 

I am not experienced in social logins/signups but as a user, some information might be requested by the app, like phone number, birth date, etc. that would make the journey for a user signing up from social quite different from one that signup using email/password.

If it is indeed the case, having separate classes for social and non-social would make more sense compared to a single unified form. That might also have been the reason why allauth has given somewhat more attention to separate forms than to a combined single signup form.

What do you think?

browniebroke: That's a good point, probably best to stick to the current approach 👍🏻

browniebroke (approved):

browniebroke:
Gave this a spin and it's all working as expected. I was able to add a new field to the `User` model and get it to show in the admin solely by customizing the `UserAdmin.fieldsets`.

PR #3519, opened 2022-01-07
Fix `open` option for `initBrowserSync` when using Docker

mateesville93:
## Description
Fix for initBrowserSync open option

Error:
```
'initBrowserSync' errored after 225 ms
Error: Cannot set unknown key "open" on Record
```

Checklist:

### DIFF
--- a/{{cookiecutter.project_slug}}/gulpfile.js
+++ b/{{cookiecutter.project_slug}}/gulpfile.js
@@ -139,9 +139,11 @@ function initBrowserSync() {
       `${paths.js}/*.js`,
       `${paths.templates}/*.html`
     ], {
+      {%- if cookiecutter.use_docker == 'y' %}
       // https://www.browsersync.io/docs/options/#option-open
       // Disable as it doesn't work from inside a container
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/gulpfile.js
```diff
@@ -139,14 +139,14 @@ function initBrowserSync() {
       `${paths.js}/*.js`,
       `${paths.templates}/*.html`
     ], {
+      // https://www.browsersync.io/docs/options/#option-open
+      // Disable as it doesn't work from inside a container
+      open: false,
```
browniebroke, line 144: Good catch, although we can let this option to true when not using Docker:

```suggestion
      {%- if cookiecutter.use_docker == 'y' %}
      // https://www.browsersync.io/docs/options/#option-open
      // Disable as it doesn't work from inside a container
      open: false,
      {%- endif %}
```

browniebroke (approved):

browniebroke:
Thank you for the fix!

PR #3535, opened 2022-01-14
Unify compressor, gulp and custom bootstrap options

browniebroke:
## Description

Replace the 3 separate options for django-compressor, Gulp task runner and custom Bootstrap compilation under a single one. 

Checklist:

## Rationale

Django compressor and Gulp are performing a very similar job, but our current questionnaire put them on 2 separate questions. That means users may opt to use both of them, which doesn't make a ton of sense in my understanding. It should be more a case of picking one out (I think). I've [asked about this](https://github.com/django-compressor/django-compressor/discussions/1095) on the Django Compressor, and it seemed shared by the maintainers.

Similarly the custom Bootstrap compilation only make sense if the user opted for Gulp task runner, otherwise the styling is broken.

This pain was especially visible in `base.html` where we have a lot of template tag interleaved.


cookiecutter.json
```diff
@@ -46,8 +42,12 @@
   ],
   "use_async": "n",
   "use_drf": "n",
-  "custom_bootstrap_compilation": "n",
-  "use_compressor": "n",
+  "frontend_pipeline": [
+    "None",
+    "Django Compressor",
+    "Gulp",
+    "Gulp with custom Bootstrap"
```
browniebroke, line 49: I'm very tempted to merge these 2 and make the custom Bootstrap the default. The Gulp setup without the Bootstrap compilation doesn't provide a lot of value IMO.

Issue #3346, opened 2021-10-06
MySQL support


mabdullahadeel:
## Description
`cookiecutter-django` at the moment only supports Postgresql setup. It would be great if MySQL is implemented as an option with Docker setup as MySQL is another highly used DB with Django

What are you proposing? How should it be implemented?
MySQL cookiecutter support with Docker setup

## Rationale

Why should this feature be implemented?
MySQL support.

PR #3539, opened 2022-01-17
mysql support

mabdullahadeel:
## Description

- Added full MySQL support
    - MySQL support in local development
    - MySQL support with docker (local & production)

Checklist:

## Rationale

Mysql is one of the heavily used databases with Django. Adding support for this will open `cookiecutter` to those developers who prefer using MySQL with django. Besides that, django also has full support for MySQL.


Afrowave:
Good work @mabdullahadeel.  
  
I was a moderate user of MySQL, so this will be helpful to those coming out MySQL-heavy frameworks. However PostgreSQL is an industry-leading database technology hiding in the OSS, which why Django fully supports it.

browniebroke:
Waw, that's a big one, I'm sure people who prefer MySQL would like that! 

As per #3346, we're not going to accept this but you're welcome to send us a PR to add a link to the README. To not include any of these changes in that other pull request, on your fork, you can create a branch from the latest common revision with cookiecutter/cookiecutter-django, and add the link there.

PR #3591, opened 2022-02-15
fix typo in servers settings for drf-spectacular

ghazi-git:
update the local dev server url to use http instead of https

## Description

Checklist:


luzfcb:
@ghazi-git Thank you. 🚢 :shipit:

PR #3592, opened 2022-02-15
Fix Swagger schema API endpoint & add a test for it

ghazi-git:
## Description

Checklist:

## Rationale

Currently, after a new project generation with `use_drf: y`, the API call to the schema endpoint fails with the error
```
AssertionError: Incompatible AutoSchema used on View <class 'drf_spectacular.views.SpectacularSwaggerView'>. Is DRF's DEFAULT_SCHEMA_CLASS pointing to "drf_spectacular.openapi.AutoSchema" or any other drf-spectacular compatible AutoSchema?
```
So, the change sets the default schema class as `drf_spectacular.openapi.AutoSchema` as per [drf-cpectacular docs](https://drf-spectacular.readthedocs.io/en/latest/readme.html#installation) and adds a test to make sure sure that schema generation works properly

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py
@@ -34,8 +34,7 @@ class TestUserViewSet:
         }
 
 
-class TestAPISchemaView:
-    def test_api_schema_generated_successfully(self, admin_client):
-        url = reverse("api-schema")
-        response = admin_client.get(url)
-        assert response.status_code == 200
+def test_api_schema_generated_successfully(admin_client):
+    url = reverse("api-schema")
+    response = admin_client.get(url)
+    assert response.status_code == 200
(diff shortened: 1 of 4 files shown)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_drf_views.py
```diff
@@ -31,3 +32,10 @@ def test_me(self, user: User, rf: RequestFactory):
             "name": user.name,
             "url": f"http://testserver/api/users/{user.username}/",
         }
+
+
+class TestAPISchemaView:
+    def test_api_schema_generated_successfully(self, admin_client):
+        url = reverse("api-schema")
+        response = admin_client.get(url)
+        assert response.status_code == 200
```
browniebroke, line 41: If we only have one method, we can keep it simple with a plain function:

```suggestion
def test_api_schema_generated_successfully(admin_client):
    url = reverse("api-schema")
    response = admin_client.get(url)
    assert response.status_code == 200
```

Also, it sounds like it would be better located [with the other swagger tests `users/tests/test_swagger_ui.py`](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/users/tests/test_swagger_ui.py)

Since this will test both the UI and the schema, we could rename it to `test_swagger.py`.

browniebroke (approved):
Thanks for the fix! I left a small comment regarding where the test is located, would you have time to change it please?

ghazi-git:
Changes done, thanks for the feedback @browniebroke

browniebroke:
Thanks again for the fix 🎉

PR #3609, opened 2022-02-20
set SERVERS in production settings instead of base settings

### LINKED ISSUE #3593: API calls from Swagger UI page fail while running the project at "localhost:8000" due to CORS
## What happened?
- New project generated with `use_drf: y`
- When making an API request from the swagger UI at `http://localhost:8000/api/docs/`, the API call fails due to CORS policy
cors

The API call should have worked with no issues

## Additional details
The issue is due to project running on localhost:8000 while the `SERVERS` option in drf-spectacular settings has the url `http://127.0.0.1:8000`.

It seems like there is more than one option to fix this:
- one way is to set the `CORS_ALLOWED_ORIGINS = ["http://localhost:8000", "http://127.0.0.1:8000"]` in local settings so that api calls from `localhost:8000` work fine ([django-cors-headers docs](https://github.com/adamchainz/django-cors-headers#cors_allowed_origins-sequencestr))
- the other option is to remove `SERVERS` from drf-spectacular

I'm leaning towards the second option and that's because I don't think it's a good idea to be able to make API call to a production server when working locally. Also, When the project is deployed to production, it doesn't make a lot of sense to see the option for making API calls to "127.0.01:8000" in the swagger UI page.

So, before creating a PR, I wanted to check if there are good reasons to keep `SERVERS` in drf-spectacular settings (or maybe there is a better third option)

ghazi-git:
it is removed from local settings since having the url as 127.0.0.1:8000 triggers a CORS issue when the project runs on localhost:8000. It is kept in production settings as that allows tools that generate code samples to point to the correct domain.

## Description

Checklist:

## Rationale

Based on the comments in the issue, I thought it would be good to keep `SERVERS` in production settings so the original goal of adding it is met, but also thought it might not be useful in local settings (plus adding it requires setting `CORS_ALLOWED_ORIGINS=["localhost:8000"]`)

fixes #3593

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -346,10 +346,6 @@ SPECTACULAR_SETTINGS = {
     "DESCRIPTION": "Documentation of API endpoints of {{ cookiecutter.project_name }}",
     "VERSION": "1.0.0",
     "SERVE_PERMISSIONS": ["rest_framework.permissions.IsAdminUser"],
-    "SERVERS": [
-        {"url": "http://127.0.0.1:8000", "description": "Local Development server"},
-        {"url": "https://{{ cookiecutter.domain_name }}", "description": "Production server"},
-    ],
 }
 {%- endif %}
 # Your stuff...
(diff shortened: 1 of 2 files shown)


browniebroke (approved):
Looks good to me. 

@luzfcb do you want to give this a look before we merge it?

PR #3680, opened 2022-04-21
Add the possibility to set a max django version on create_django_issue script

luzfcb:
## Description

- Add the possibility to set a max Django version
- Remove newline character from `version` and `compat_version` before generating the markdown

usage: `python scripts/create_django_issue.py MAX_VERSION`
example: `python scripts/create_django_issue.py 3.2`

## Rationale

- I found that the pull-request created by `create_django_issue` script can be very useful when I need to upgrade a major version of django. 
I was trying to make it more reusable and independent of cookiecutter-django.

A few things I'd like to add to this script in the future:
- include in the table the first version of the dependency that has official support for Django x.x.
- Link to the repository with the source code (if there is a link in the package metadata)

### DIFF
--- a/scripts/create_django_issue.py
+++ b/scripts/create_django_issue.py
@@ -47,6 +47,11 @@ class DjVersion(NamedTuple):
         major, minor, *_ = version_str.split(".")
         return cls(major=int(major), minor=int(minor))
 
+    @classmethod
+    def parse_to_tuple(cls, version_str: str):
+        version = cls.parse(version_str=version_str)
+        return version.major, version.minor
+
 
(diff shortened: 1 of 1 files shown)


browniebroke (approved):
> I was trying to make it more reusable and independent of cookiecutter-django.

I agree, this could be an standalone package on PyPI, with a simple CLI and maybe a GitHub action. However, if that was the case, it would probably start to receive feature requests (e.g. support for other package managers) and grow in scope, and I don't really want to have to support all these cases :)

Anyway, this PR looks good to me

PR #3698, opened 2022-05-06
Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions
- https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
- [Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/pre-commit-autoupdate.yml
+++ b/.github/workflows/pre-commit-autoupdate.yml
@@ -8,9 +8,15 @@ on:
     - cron: "15 2 * * *"
   workflow_dispatch: # to trigger manually
 
+permissions:
+  contents: read
+
 jobs:
   auto-update:
     # Disables this workflow from running in a repository that is not part of the indicated organization/user
+    permissions:
+      contents: write  # for peter-evans/create-pull-request to create branch
+      pull-requests: write  # for peter-evans/create-pull-request to create a PR
     if: github.repository_owner == 'cookiecutter'
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


browniebroke (commented):
By reading [the article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) in the last link, I'm under the impression that the problem is with certain kind of workflow events. However, it's not clear to me (yet) which ones are potentially insecure, or what makes them insecure... I probably need to read a bit more about these.

Do you think we could also apply a similar fix to the workflows in the generated projects?

Anyway thanks for bringing this up to the project attention.

naveensrinivasan:
> By reading [the article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) in the last link, I'm under the impression that the problem is with certain kind of workflow events. However, it's not clear to me (yet) which ones are potentially insecure, or what makes them insecure... I probably need to read a bit more about these.
> 
> Do you think we could also apply a similar fix to [the workflows in the generated projects](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.github/workflows/ci.yml)?
> 
> Anyway thanks for bringing this up to the project attention.

Yes! IMHO you should.

PR #3702, opened 2022-05-07
Add pyupgrade to pre-commit config

browniebroke:
## Description

Pyupgrade automatically upgrade Python syntax to more modern construct. This adds it as pre-commit hook.

Checklist:

## Rationale

I really like pyupgrade and use it on many of my projetcs. It enables to use better syntax as soon as possible when upgrading Python version.


luzfcb (approved):
LGTM.

PR #3711, opened 2022-05-13
Update PostgreSQL versions that fix some security issues.

Afrowave:
There are some crucial updates that have been carried out on the PostgreSQL database 

## Description

This pull request updates the database versions of choice.

Checklist:

## Rationale

It would be nice for those starting a new project to start with the last fixes to the "The World's Most Advanced Open Source Relational Database". 😁

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -18,11 +18,11 @@
   "use_pycharm": "n",
   "use_docker": "n",
   "postgresql_version": [
-    "14.1",
-    "13.5",
-    "12.9",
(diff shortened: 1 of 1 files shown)


browniebroke:
Thanks! These versions aren't on docker hub yet it seems... Do you think we could specify the major version only to avoid having to keep these up to date?

We should also update the sample output in the README...

Afrowave:
You @browniebroke have an excellent point. I was wondering the same. Let me sort that and make changes to the README.

PR #3772, opened 2022-07-08
[fix] runserver needs to use keep-meta-shutdown

areski:
When you will upgrade werkzeug to 2.1.x, an issue with occur when using runserver_plus:
`
    The 'environ['werkzeug.server.shutdown']' function is deprecated and will be removed in Werkzeug 2.1.
`

In werkzeug 2.1.0 deprecated code was removed, including `werkzeug.server.shutdown`

ref: https://github.com/django-extensions/django-extensions/issues/1715
ref: https://github.com/plotly/jupyter-dash/issues/63

There is a simple fix, it's to add `--keep-meta-shutdown` to `runserver`


browniebroke:
I think this was fixed in #3774, so I'm going to close this. Let me know if I'm missing anything.

Thanks!

PR #3773, opened 2022-07-08
[fix] upgrade docker image node:16-bullseye-slim to node:18-bullseye-slim

areski:
Upgrade docker image node:16-bullseye-slim to node:18-bullseye-slim,
it's necessary to add `apt-get install python3`


{{cookiecutter.project_slug}}/compose/local/node/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM node:16-bullseye-slim
+FROM node:18-bullseye-slim
+
+RUN apt-get update && apt-get install -y python3 g++ make
```
browniebroke, line 3: Why do we need to install Python in the node image?

PR #3782, opened 2022-07-11
Provide more context when wating for PostgreSQL takes too long

jugglinmike:
## Description

Advise the operator that they may be observing an unrecoverable problem if attempts to connect to the PostgreSQL database consistently fail. In this case, also display the relevant error message.

Checklist:

## Rationale

Currently, if a connection to the PostgreSQL database cannot be established, operators are instructed to wait indefinitely. They need to exercise their own judgement to determine that an unrecoverable error may have occurred, and they must then formulate their own plan to diagnose the problem. This change introduces a suggestion about the possibility of failure and some contextual information that will assist in debugging. It also causes the script to fail immediately for truly exceptional circumstances such as the absence of required modules.

I've arbitrarily chosen 30 seconds as the duration to wait before suggesting that something may not be working. Does that seem like an appropriate value to folks more familiar with this process?


Afrowave:
👍  
  
If there are no eyes on the logs generated by the connection process, this would be very helpful.

browniebroke (approved):

browniebroke:
Great stuff! Thanks a lot!

jugglinmike:
My pleasure!

PR #3786, opened 2022-07-18
Always use `const` instead of `var` in `gulpfile.js`

sobolevn:
## Description

Checklist:

## Rationale

Using `var` is discouraged in js.


browniebroke (approved):

browniebroke:
Thanks!

PR #3807, opened 2022-07-24
Simplify database access in tests

danihodovic:
Creating an instance of the user model requires database access. When
requesting the user fixture it fails unless we explicitly mark the test
with pytest.mark.django_db.

https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest-mark-django-db-request-database-access

Pytest-django solves this elegantly with the admin_user fixture since
it implicitly depends on the `db` fixture:

https://github.com/pytest-dev/pytest-django/blob/fba51531f067a985ec6b6be4aec9a2ed5766d69c/pytest_django/fixtures.py#L406-L411

We can do the same for the cookiecutter user fixture by depending on the
db fixture.

So instead of:
```python
import pytest

pytestmark = pytest.mark.django_db

def test_user(user):
    pass
```

We can simply do:
```python
def test_user(user):
    pass
```

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/conftest.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/conftest.py
@@ -10,5 +10,5 @@ def media_storage(settings, tmpdir):
 
 
 @pytest.fixture
-def user() -> User:
+def user(db) -> User:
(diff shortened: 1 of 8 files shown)


browniebroke (approved):
LGTM

browniebroke:
Thanks for the contribution! 🎉

PR #3844, opened 2022-08-24
Add article to README about how to use a hosted DB

professorhaseeb:
help article


browniebroke (approved):
Very nice and simple article. Just a small bit of feedback that it's actually not specific to AWS RDS, but would also work for any hosted database.

PR #3855, opened 2022-08-31
Store extended Celery task attributes in backend

danihodovic:
> Enables extended task result attributes (name, args, kwargs, worker,
> retries, queue, delivery_info) to be written to backend.

https://docs.celeryq.dev/en/stable/userguide/configuration.html#result-extended

Without this the information in the result backend is rather opaque. You
can only see the task_id's of tasks that have executed. With the setting
enabled you can see the task name, queue and other metadata.

To demonstrate - the top task is with CELERY_RESULT_EXTENDED enabled
while the rest have the value disabled.

image


danihodovic:
Ping @browniebroke

browniebroke:
Thanks!

PR #3866, opened 2022-09-09
Reload uvicorn on html file change

monosans:
## Description

Checklist:

## Rationale

After changing the html file uvicorn must be reloaded for Django to pick up the changes, otherwise it will show the old version of the template.


browniebroke (approved):

browniebroke:
Thanks!

PR #3871, opened 2022-09-15
Mailjet default api url does not work out of the box

brumiser1550:
It seems best to not include this setting and let AnyMail handle it. https://github.com/anymail/django-anymail/blob/a08052e7f808c800cb336df3725b63865dc36afc/anymail/backends/mailjet.py#L19


browniebroke (approved):
Sounds reasonable... 👍🏻

PR #3405, opened 2021-11-15
Fix ungraceful Celery workers shutdown in container

goatwu1993:
## Description

Checklist:

- [x] I've made sure that `tests/test_cookiecutter_generation.py` is updated accordingly (especially if adding or updating a template option)
- [x] I've updated the documentation or confirm that my change doesn't require any updates

## Rationale

I use this awesome project to bootstrap a containerized Django + Celery application on AWS ECS.

I found out when I rolled upgrade Celery deployment in ECS, the old Celery pods shut down ungracefully. These old pods kept receiving tasks even if they are dying in the near future. Any ongoing tasks will then hang in some weird state.

Fix #1743 

## How to reproduce

- Create a project with celery and docker enabled
- Start the application
   ```bash
   docker-compose -f production.yml up --build
   ```
- Open another terminal to follow the logs of celery
   ```bash
   docker logs -f my_awesome_project_celeryworker_1
   ```
- Stop the container
   ```bash
   docker stop my_awesome_project_celeryworker_1
   ```
- See how the logs of celeryworker ended.

Expected
- Should see a warm shutdown in Celery log.

Reality
- Not logs showing warm shutdown. After the timeout, the container shut down ungracefully.

## How to Fix

- Add exec before celery

## Explanation

- In docker-compose,  kubectl, and AWS ECS, the way the Control pane gracefully shut down a container is
  1. Control pane send a SIGTERM
  2. The dying Application handles the SIGTERM by doing the unfinished tasks or sending the task back to the queue.
  3. Control pane send a SIGKILL If grace period exceeds (configurable, usually 30s to 60s) (Ungraceful shutdown)

- Celery has a built-in SIGTERM handling (although undocumented...), by doing the ongoing task or pushing the task into the queue and stopping receiving any new task.
- AWS Explanation for ECS: https://aws.amazon.com/blogs/containers/graceful-shutdowns-with-ecs/  
   Note: This is not ECS specific but also Kubernetes, docker, and docker-compose. I paste this page only because it is easier to read.

## Some screenshots

Without graceful shutdown
Screen Shot 2021-11-15 at 2 29 31 PM

With graceful shutdown

Screen Shot 2021-11-15 at 2 30 51 PM

PR #3925, opened 2022-10-30
Fix graceful shutdown of local dev containers and use watchfiles for beat + flower

goatwu1993:
## Description

- Fix local dev container may not gracefully shutdown (similar to #3405)
- Watchfiles for celery beat/flower when local dev

Checklist:


{{cookiecutter.project_slug}}/compose/local/django/start
```diff
@@ -9,5 +9,5 @@ python manage.py migrate
 {%- if cookiecutter.use_async == 'y' %}
 uvicorn config.asgi:application --host 0.0.0.0 --reload --reload-include '*.html'
```
browniebroke, line 10: Should we put a `exec` in fron of this one too?
```suggestion
exec uvicorn config.asgi:application --host 0.0.0.0 --reload --reload-include '*.html'
```

browniebroke (approved):
Looks good otherwise.

browniebroke:
Thanks 🎉 🎉

PR #3562, opened 2022-02-03
Fix docker-compose config on Apple silicon

jorgeavaldez:
## Description

Adds an explicit platform to relevant python service definitions in both `local.yml` and `production.yml`.

Checklist:

- [x] I've made sure that tests are updated accordingly (especially if adding or updating a template option)
- [x] I've updated the documentation or confirm that my change doesn't require any updates

## Rationale
When running from scratch on Apple Silicon, the default `django` service would continuously poll for a postgres connection. After logging the `psycopg2` exception in the container entrypoint, I found that the script was silencing a connection error stating the following:

```
SCRAM authentication requires libpq version 10 or above
```

Upon further investigation, I found that there was an inconsistency with some versions of the `python` image that would install an outdated version of `libpq`. According to the [Docker Desktop documentation](https://docs.docker.com/desktop/mac/apple-silicon/#known-issues):

> Not all images are available for ARM64 architecture. You can add --platform linux/amd64 to run an Intel image under emulation.

We can specify this within the docker-compose files by setting the `platform` property on a service.

Issue #3770, opened 2022-07-06
Django server doesn't reload after changes in app


BenjaminArmijo3:
## What happened?
Hi i have used a previous version of cookiecutter-django but the last one doest reload the developent server when the files change

* Host system configuration:
    * OS name and version:  Mac M1

foarsitter:
Did you use it before on the Mac M1? Did you find some clues on the web about M1 + Django + reloading? Do you use docker?

BenjaminArmijo3:
the diff is that in the newer version have the line platform, i tried to run with the line commented but the bug persist
  django: &django
    platform: linux/x86_64

foarsitter:
Watchdog is configured through werkzeug in `requirements/local.txt`. Is your previous project using the `stat` reloader?

BenjaminArmijo3:
I change to the older version and it restart with stat.

it seems that the problem is with watchdog

foarsitter:
See [here how werkzeug](https://werkzeug.palletsprojects.com/en/2.1.x/serving/#reloader) decides which reloader is uses and [here how you can override](https://django-extensions.readthedocs.io/en/latest/runserver_plus.html?highlight=stat#configuration) the setting in runserver_plus.

camrail:
Hi all, same issue - I just today switched to apple silicon.  I had a working project on my intel based mac and then after migrating to the new laptop, I spun up the containers and django was no longer seeing changes.

I think this may be talking about the same problem in this [Watchdog issue](https://github.com/gorakhargosh/watchdog/issues/838)

BenjaminArmijo3:
> 

did you fixed it ?

camrail:
> > 
> 
> did you fixed it ?

Not a fix but a workaround is to change runserver_plus reloader to stat rather than Watchman given Watchman isn't working.  Thanks @foarsitter.

@BenjaminArmijo3 you can add the following line to your local.py settings file.
`RUNSERVERPLUS_POLLER_RELOADER_TYPE = 'stat'`

Unfortunately the reload is takes a bit longer to trigger than you may be used to.

Eugeny:
For me, switching the dev container to arm64 by removing `platform: linux/x86_64` from `local.yml` did the trick - apparently that's what prevented inotify from working on a mounted volume

browniebroke:
> switching the dev container to arm64 by removing `platform: linux/x86_64` from `local.yml` did the trick

[...]

PR #3957, opened 2022-11-22
Remove platform from compose file

leifurhauks:
## Description

I believe that the workaround introduced in PR #3562 is no longer needed. Using a fresh project created with cookiecutter-django, I was able to run the app with docker compose using the arm64 platform for the django and postgres images, without any database connection issue.

I compared the version of libpq on the arm64 image with the one on the amd64 image and they were identical.

If any core contributors are using an M1 mac, it would be good to test this before merging to ensure that this isn't just a quirk of my machine.

If it helps, I'm using Docker Desktop v4.14.1 with Engine v20.10.21 and compose v1.29.2, on macOS 12.6.

Checklist:

  - note: nothing to update for the tests here, but I did run tox locally

## Rationale

If the workaround is no longer needed, it is preferable on M1 macs to use the default arm64 platform for better performance.

### DIFF
--- a/{{cookiecutter.project_slug}}/local.yml
+++ b/{{cookiecutter.project_slug}}/local.yml
@@ -11,7 +11,6 @@ services:
       dockerfile: ./compose/local/django/Dockerfile
     image: {{ cookiecutter.project_slug }}_local_django
     container_name: {{ cookiecutter.project_slug }}_local_django
-    platform: linux/x86_64
     depends_on:
       - postgres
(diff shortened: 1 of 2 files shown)


browniebroke (approved):
Just tested, and it works fine, yes.

Thanks 👍🏻

browniebroke:
Nice side effect of this: #3770 is fixed 🎉 

Thanks!

PR #3961, opened 2022-11-22
feat(celery): Enable sending the sent task event by default

adinhodovic:
## Description

Enable sending task sent event by default. It's useful for e.g Flower and Celery-exporter. 

Checklist:

## Rationale

Helps differ between sent to workers and received by them.

Forgot that this is disabled in previous pr.


luzfcb:
@adinhodovic, thanks again. 🚢

PR #3967, opened 2022-11-26
Add Azure Storage as an option to serve static and media files

### LINKED ISSUE #2301: Add Blob Storage as an option for static and media files
## Description

Can I add Blob Storage (Microsoft Azure) as an option to serve static and media files?

## Rationale

I always use Amazon S3 for serving static and media files, but now I have a client that has a preference for Blob Storage (Microsoft Azure), so I thought about adding this as an option because it can be helpful for other developers.

In my case, I have to use it because a client has a preference for it. But maybe someone feels more comfortable using Azure instead of AWS or Google Cloud.

I can make the PR if you all think it's ok.

rguptar:
## Description
Add Azure Storage as an option to serve static and media files

Checklist:

## Rationale
Fix #2301


docs/project-generation-options.rst
```diff
@@ -66,7 +66,8 @@ cloud_provider:
 
     1. AWS_
     2. GCP_
-    3. None
+    3. AZURE_
```
browniebroke, line 69: In restructured text, this is a link, which target should be specified later in the file:

https://github.com/cookiecutter/cookiecutter-django/blame/0a8b15615a51f09083ce5386c4e09c5fe2ed2757/docs/project-generation-options.rst#L148-L149

docs/settings.rst
```diff
@@ -49,6 +49,9 @@ DJANGO_AWS_S3_CUSTOM_DOMAIN             AWS_S3_CUSTOM_DOMAIN        n/a
 DJANGO_AWS_S3_MAX_MEMORY_SIZE           AWS_S3_MAX_MEMORY_SIZE      n/a                                            100_000_000
 DJANGO_GCP_STORAGE_BUCKET_NAME          GS_BUCKET_NAME              n/a                                            raises error
 GOOGLE_APPLICATION_CREDENTIALS          n/a                         n/a                                            raises error
+DJANGO_AZURE_STORAGE_ACCOUNT_KEY        AZURE_ACCOUNT_KEY           n/a                                            raises error
+DJANGO_AZURE_STORAGE_ACCOUNT_NAME       AZURE_ACCOUNT_NAME          n/a                                            raises error
```
browniebroke, line 53: Shall we just call them `DJANGO_AZURE_ACCOUNT_KEY` and `DJANGO_AZURE_ACCOUNT_NAME`?

hooks/pre_gen_project.py
```diff
@@ -72,11 +72,18 @@
...
+        and "{{ cookiecutter.mail_service }}" == "Amazon SES"
+    )
+    or (
+        "{{ cookiecutter.cloud_provider }}" == "AZURE"
+        and "{{ cookiecutter.mail_service }}" == "Amazon SES"
+    )
+    or (
+        "{{ cookiecutter.cloud_provider }}" == "None"
+        and "{{ cookiecutter.mail_service }}" == "Amazon SES"
+    )
```
browniebroke, line 86: Maybe it's time to simplify this?

```suggestion
    (
        "{{ cookiecutter.mail_service }}" == "Amazon SES" 
        and "{{ cookiecutter.cloud_provider }}" != "AWS"
    )
```

cookiecutter.json
```diff
@@ -27,6 +27,7 @@
   "cloud_provider": [
     "AWS",
     "GCP",
+    "AZURE",
```
browniebroke, line 30: The other 2 (AWS and GCP) are in capital letters because they are acronyms. I think the name "Azure" is a single word. Shall we change that to `Azure` (not all caps)?

browniebroke (commented):
Thanks a lot for the change. I think it's covering most of it. I left a few suggestions around naming conventions as well as a broken link in the docs.

Could you fix these when you have a moment please?

Thanks again!

cookiecutter.json
rguptar: Yep, I've updated all the references.

docs/settings.rst
rguptar: I thought ACCOUNT_NAME and ACCOUNT_KEY was a bit ambiguous, but I agree that it makes sense to align with what `django-storages` uses.

docs/project-generation-options.rst
rguptar: Thanks for catching this, updated!

browniebroke (approved):
Looks good

browniebroke:
Thanks again 🎉

PR #3968, opened 2022-11-28
Populate user name field in social auth

Andrew-Chen-Wang:
## Description

The "name" field is not filled with social authentication using django allauth due to removing the first and last name field in the User model. This PR fills in the name field.

Checklist:

## Rationale

To fill in the name field during social auth


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/adapters.py
```diff
@@ -14,3 +14,13 @@ def is_open_for_signup(self, request: HttpRequest):
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
     def is_open_for_signup(self, request: HttpRequest, sociallogin: Any):
         return getattr(settings, "ACCOUNT_ALLOW_REGISTRATION", True)
+
+    def populate_user(self, request, sociallogin, data):
+        user = sociallogin.user
```
browniebroke, line 19: Could you please add a docstring to this method explaining what it does? And link to any relevant piece of documentation if such thing exist.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/adapters.py
```diff
@@ -14,3 +14,13 @@ def is_open_for_signup(self, request: HttpRequest):
...
+        if name := data.get("name"):
+            user.name = name
+        elif first_name := data.get("first_name"):
+            user.name = first_name
+            if last_name := data.get("last_name"):
+                user.name += f" {last_name}"
```
browniebroke, line 25: I seems like it would be more consistent with django-allauth if we were to do it in the `save_user()` method: https://github.com/pennersr/django-allauth/blob/7060e3fb6e00f01bd881890546244670929b5cf3/allauth/account/adapter.py#L224-L251

Unless there a specific reason not to?

browniebroke (commented):
Do you have an exemple of a social provider I could use to test this without too much setup? Also, it would be nice if we could write an automated test for it, this is

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/adapters.py
Andrew-Chen-Wang: It's been awhile, but I believe `save_user` isn't actually used for social registration (i.e. it's not called)

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/adapters.py
Andrew-Chen-Wang:
```suggestion
        """
        Populates user information from social provider info
        See: https://django-allauth.readthedocs.io/en/latest/advanced.html?highlight=populate_user#creating-and-populating-user-instances
        """
        user = sociallogin.user
```

Andrew-Chen-Wang:
> Do you have an exemple of a social provider I could use to test this without too much setup?

I find Google the easiest and Twitter the second easiest to set up.

> Also, it would be nice if we could write an automated test for it

We could add mock requests? Otherwise, not sure whether adding these OAuth providers and a bot for automated testing will trigger a captcha warning.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/adapters.py
```diff
@@ -14,3 +14,16 @@ def is_open_for_signup(self, request: HttpRequest):
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
     def is_open_for_signup(self, request: HttpRequest, sociallogin: Any):
         return getattr(settings, "ACCOUNT_ALLOW_REGISTRATION", True)
+
+        """
```
Andrew-Chen-Wang, line 18:
```suggestion
        def populate_user(self, request, sociallogin, data):
        """
```

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/adapters.py
```diff
@@ -14,3 +14,17 @@ def is_open_for_signup(self, request: HttpRequest):
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
     def is_open_for_signup(self, request: HttpRequest, sociallogin: Any):
         return getattr(settings, "ACCOUNT_ALLOW_REGISTRATION", True)
+
+        def populate_user(self, request, sociallogin, data):
```
Andrew-Chen-Wang, line 18:
```suggestion
    def populate_user(self, request, sociallogin, data):
```

browniebroke (approved):

PR #3969, opened 2022-11-28
Simplify production `DATABASES` setting to extend base definition

adinhodovic:
## Description

Remove production db settings being instantiated twice when using docker.

Checklist:

## Rationale

I believe each `env.db()` resets the dict. Keep shared settings in base.py. Otherwise we have to redefine every db parameter, such as atomic requests.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -22,9 +22,6 @@ ALLOWED_HOSTS = env.list("DJANGO_ALLOWED_HOSTS", default=["{{ cookiecutter.domai
 
 # DATABASES
 # ------------------------------------------------------------------------------
-{% if cookiecutter.use_docker == "n" -%}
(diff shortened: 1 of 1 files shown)


foarsitter:
This indeed looks like a duplicated definition, but why is there an exception needed for a non docker environment? The initialisation is the same right?

browniebroke:
> This indeed looks like a duplicated definition, but why is there an exception needed for a non docker environment? The initialisation is the same right?

Thanks for bumping this discussion and raising this point, I have the same question. It feels like we could remove this altogether?

adinhodovic:
> > This indeed looks like a duplicated definition, but why is there an exception needed for a non docker environment? The initialisation is the same right?
> 
> Thanks for bumping this discussion and raising this point, I have the same question. It feels like we could remove this altogether?

Hmm, makes sense when I took a second look can remove it altogether. Updated.

browniebroke (approved):

PR #4023, opened 2022-12-23
Configuration of the external port number of the django application

HenryJobst:
## Description

Possibility to change the default port (8000) of the django application with the new template option django_port.

Checklist:

## Rationale

With this addition it is possible to run different (generated) django applications at the same time.


foarsitter:
Although I agree on making the server port configurable I do not see the necessity to add it as an cookiecutter option. 

If we make the port configurable for development I would say we should do the same for production (port 5000). Besides that `gulpfile.js` and the documentation needs to be updated.

browniebroke:
I'm -1 on this. It doesn't seem like a common enough use case to justify making it an option... We need to keep a balance between making it flexible while keeping the number of questions not too high.

browniebroke:
I'm going to close this as it's pretty clear that it has grown beyond the original scope. 

If others are interested in the originally suggested feature, let's discuss in an issue and collect interest.

PR #4036, opened 2022-12-30
Fix link, add non-Docker commands

pamelafox:
## Description

The current coverage link is a 404 so I updated it to point to latest, similar to the pytest link. I also put the non-Docker version of commands first, to mirror what's done earlier in the article.

Checklist:

## Rationale

Main motivation is fixing the 404.


browniebroke (approved):
Thanks!

PR #4054, opened 2023-01-06
Add .git to .dockerignore

Kaffeetasse:
## Description

I don't want unnecessary files and folders in Docker images. The .git directory is not needed there.

Checklist:

## Rationale

I had a huge file in my git tree and while building the image I've seen that the build context was very big, also the repository is simply not needed in the Docker image.


browniebroke (approved):
Good catch! Thanks!

PR #4061, opened 2023-01-06
Update Celery instructions

pamelafox:
## Description

I got confused when I first followed the Celery instructions as I'm new to Celery and didn't know that I needed to explicitly queue the task. I've modified the instructions to show how to do that with the built-in get_users_count task.

Checklist:

## Rationale

Hopefully this makes Celery setup more clear for other celery noobs.

### DIFF
--- a/docs/developing-locally.rst
+++ b/docs/developing-locally.rst
@@ -147,12 +147,12 @@ If the project is configured to use Celery as a task scheduler then, by default,
 
 The project comes with a simple task for manual testing purposes, inside `<project_slug>/users/tasks.py`. To queue that task locally, start the Django shell, import the task, and call `delay()` on it::
 
-    $ python manage.py shell 
+    $ python manage.py shell
     >> from <project_slug>.users.tasks import get_users_count
     >> get_users_count.delay()
-    
+
 Next, make sure `redis-server` is installed (per instructions at https://redis.io/topics/quickstart) and run the server in one terminal::
-    
+
     $ redis-server
 
 Now that a task is queued and Redis is running, the final step is to start the Celery worker locally. In another terminal, run the following command::
--- a/docs/developing-locally.rst
+++ b/docs/developing-locally.rst
@@ -141,15 +141,23 @@ In production, we have Mailgun_ configured to have your back!
 Celery
 ------
 
-If the project is configured to use Celery as a task scheduler then by default tasks are set to run on the main thread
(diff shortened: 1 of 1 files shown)


docs/developing-locally.rst
```diff
@@ -141,15 +141,23 @@ In production, we have Mailgun_ configured to have your back!
 Celery
 ------
 
-If the project is configured to use Celery as a task scheduler then by default tasks are set to run on the main thread
-when developing locally. If you have the appropriate setup on your local machine then set the following
-in ``config/settings/local.py``::
+If the project is configured to use Celery as a task scheduler then, by default, tasks are set to run on the main thread when developing locally instead of getting sent to a broker. However, if you have Redis setup on your local machine, you can set the following in ``config/settings/local.py``::
```
pamelafox, line 144: I was also confused by the original wording of this, as I wasn't sure what "appropriate setup" means, and whether I had the appropriate setup. After doing some research, I think it means having a broker setup, so I reworded it to that.

docs/developing-locally.rst
```diff
@@ -141,15 +141,23 @@ In production, we have Mailgun_ configured to have your back!
...
+    >> from <project_slug>.users.tasks import get_users_count
+    >> get_users_count.delay()
+
+Next, make sure `redis-server` is installed (per instructions at https://redis.io/topics/quickstart) and run the server in one terminal::
+
+    $ redis-server
+
+Now that a task is queued and Redis is running, the final step is to start the Celery worker locally. In another terminal, run the following command::
+
+    $ celery -A config.celery_app worker --loglevel=info
```
browniebroke, line 160: I'm not sure whether it's clear with the current wording, which seems to imply that it's a one-off thing. Maybe we could also add a sentence at the end to explain that the worker can/should be left running in the background, which would enable other tasks to be queued, and the worker will pick tasks and process them as they come. No need to expand too much, and maybe Celery has a page explaining it better... What do you think?

docs/developing-locally.rst
browniebroke: I think that's clearer, yes 👍🏻

browniebroke (approved):

docs/developing-locally.rst
pamelafox: Good point @browniebroke! I've reworded, reordered, and added links. Let me know what you think of the revision.

browniebroke (approved):
Looks good

PR #4062, opened 2023-01-06
Add dump.rdb to gitignore

pamelafox:
## Description

After running Redis locally, I noticed dump.rdb show up in my repository, which is apparently the Redis dump file. This PR adds dump.rdb to gitignore. I thought about putting it behind a "use_celery = y" guard, but realized Redis is also the caching mechanism, so it seems relevant for all configurations.

Checklist:

## Rationale

I don't believe the dump file belongs in version control.

### DIFF
--- a/{{cookiecutter.project_slug}}/.gitignore
+++ b/{{cookiecutter.project_slug}}/.gitignore
@@ -326,6 +326,9 @@ Session.vim
 # Auto-generated tag files
 tags
 
+# Redis dump file
+dump.rdb
+
 ### Project template
 {%- if cookiecutter.use_mailhog == 'y' and cookiecutter.use_docker == 'n' %}
 MailHog


browniebroke:
Can you explain shortly dow did you have this file show up in your project? I don't remember seeing it... Perhaps it's because I don't usually run Redis on my host machine?

pamelafox:
Yep, I started `redis-server` command from the root of my project directory (my_awesome_project), and that saves dumps of the DB every so often.

```
my_awesome_project % redis-server
```

I suspect this is what creates dump.rdb:

```
98819:M 06 Jan 2023 11:06:00.036 * Background saving started by pid 4499
4499:C 06 Jan 2023 11:06:00.046 * DB saved on disk
```

I started it locally so that I could check how caching and celery worked locally before testing them on production.

browniebroke (approved):

PR #4099, opened 2023-01-24
Update postgres and redis to point to mini tiers

theptrk:
## Description

Changing the commands for heroku postgres and redis

## Rationale

The free tiers no longer exist so these commands are broken


browniebroke:
Could you please also make sure that the build is green? Thanks

PR #4086, opened 2023-01-18
Rename ROOT_DIR to BASE_DIR

tildebox:
## Description

Checklist:

- [ ] I've made sure that tests are updated accordingly (especially if adding or updating a template option)
- [ ] I've updated the documentation or confirm that my change doesn't require any updates

## Rationale

Fix #4075

PR #4105, opened 2023-01-26
Refactor `merge_production_dotenvs_in_dotenv.py`

browniebroke:
## Description

Following up from #4086 and #4102, I wanted to rename a few more variables and then realised that this file could be modernised since it uses Pathlib. 

The tests was pretty much reimplementing the same logic as the function, and I think it would benefit from being "dumber" and simply write a bunch of input and compare it with the extected output.

Checklist:

I think we could look at replacing this script by more standard tools, but that's probably one for another day:

- On Linux and macOS, it's pretty simple to do `cat file1 file2 ... > .env`  
- On Windows, I found this, but I have no idea if that's a viable option: https://www.shellhacks.com/windows-cat-equivalent-cmd-powershell/

## Rationale

Modernize script and make it more readable.

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -293,6 +293,7 @@ def set_flags_in_settings_files():
 def remove_envs_and_associated_files():
     shutil.rmtree(".envs")
     os.remove("merge_production_dotenvs_in_dotenv.py")
+    shutil.rmtree("tests")
 
 
 def remove_celery_compose_dirs():
--- a/{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
+++ b/{{cookiecutter.project_slug}}/merge_production_dotenvs_in_dotenv.py
@@ -18,13 +18,12 @@ def merge(
     merged_files: Sequence[Path],
     append_linesep: bool = True,
 ) -> None:
-    with open(output_file, "w") as output_file:
-        for merged_file_path in merged_files:
-            with open(merged_file_path) as merged_file:
-                merged_file_content = merged_file.read()
-                output_file.write(merged_file_content)
-                if append_linesep:
-                    output_file.write(os.linesep)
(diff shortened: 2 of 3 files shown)


luzfcb:
@browniebroke can we move the tests to another file? I guess it shouldn't be mandatory to have pytest installed to run this specific file

browniebroke:
> @browniebroke can we move the tests to another file? I guess it shouldn't be mandatory to have pytest installed to run this specific file

Yes I wanted to do that but I think there might be some imports issue I'll need to address.

browniebroke:
Another benefit of moving to a separate file: tests are [actually running](https://github.com/cookiecutter/cookiecutter-django/actions/runs/4019996894/jobs/6907458459#step:5:2436) now ([compared to master](https://github.com/cookiecutter/cookiecutter-django/actions/runs/4018856764/jobs/6904986697#step:5:2422))

browniebroke:
@luzfcb WDYT? I think it's ready to go now. Let me know if you have any other suggestions, otherwise I'll merge it a bit later

luzfcb (approved):
LGTM 👍🏽

hooks/post_gen_project.py
```diff
@@ -293,6 +293,7 @@ def set_flags_in_settings_files():
 def remove_envs_and_associated_files():
     shutil.rmtree(".envs")
     os.remove("merge_production_dotenvs_in_dotenv.py")
+    shutil.rmtree("tests")
```
luzfcb, line 296: Why not remove /tests/test_merge_production_dotenvs_in_dotenv.py only?

luzfcb (commented):
.

hooks/post_gen_project.py
browniebroke: That should be the only file in there. It's a new tests directory at the top level.

PR #4112, opened 2023-01-28
Remove BrokenLinkEmailsMiddleware

browniebroke:
## Description

Reverting this middleware that was added in #2355 

FYI @danihodovic 

Checklist:

## Rationale

While it's good to keep an eye on broken links, I don't think we should enable it by default. I got spammed with a few emails from seemingly bots trying to access the site URLs with `/favicon.ico` at the end, for example: `/about//favicon.ico`.

If the site suddenly receives a spike of traffic and this goes out of hand, it could cause significant costs from the email service.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -152,7 +152,6 @@ MIDDLEWARE = [
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
-    "django.middleware.common.BrokenLinkEmailsMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
 ]
 


foarsitter:
It is an annoying feature imho. Never gave me some useful feedback. LGTM.

danihodovic (approved):

browniebroke:
👍🏻  Thanks for the feedback folks

tildebox:
Since BrokenLinkEmailsMiddleware has been removed, the following setting may also be removed. What do you think?

config\settings\base.py

`MANAGERS = ADMINS`

MANAGERS
A list in the same format as ADMINS that specifies who should get broken link notifications when BrokenLinkEmailsMiddleware is enabled.

Source: https://docs.djangoproject.com/en/4.2/ref/settings/#managers

browniebroke:
Great point! Looks like the setting was there before we started [using the middleware](https://github.com/cookiecutter/cookiecutter-django/pull/2355), though. The middleware uses it via the `mail_managers` function, which is in the [documented API of Django](https://docs.djangoproject.com/en/4.2/topics/email/#mail-managers), so one might use this setting independently from the middleware... All in all, I'm not sure if it's worth removing it.

tildebox:
I think this setting can be removed because Cookiecutter Django doesn't use the `mail_managers` function.

agusmakmun:
Issues:
1. Generating alert fatigues when the URLs is not from our app, for example when someone scan our app.
2. Costing the email provider

Solution:

- Modify BrokenLinkEmailsMiddleware to only catch the 404 error when urls/path's is come from our application
- Django have capability to generate all url patterns, use this as a solution wether the requested path is match to our URL patterns or not.

Example:

```python
from django.urls import get_resolver

def get_urls_from_pattern(pattern, prefix=''):
    if hasattr(pattern, 'url_patterns'):
        # For included URL patterns
        return sum(
            [get_urls_from_pattern(p, prefix + str(pattern.pattern)) for p in pattern.url_patterns],
            []
        )
    elif hasattr(pattern, 'lookup_str'):
        # For APIViews (django-rest-framework)
        return [prefix + str(pattern.pattern)]
    else:
        # For regular paths
        return [prefix + str(pattern.pattern)]

def get_all_urls():
    resolver = get_resolver()
    all_urls = []

    for pattern in resolver.url_patterns:
        all_urls.extend(get_urls_from_pattern(pattern))

    return all_urls


# Example usage
all_urls = get_all_urls()
print(all_urls)
```

Issue #4119, opened 2023-01-30
Upgrade Traefik docker configurations to the latest Traefik version


luzfcb:
Currently, the Docker-based project is shipping with Traefik 2.2.11. 
The latest version of Traefik is 2.9.6, but 3.0.0 is close to being released.

It would be good for us to update to the latest available version of Traefik.
The files that maybe should be updated are:

https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/Dockerfile

https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/traefik.yml

Perhaps to perform the update you only need to update the Traefik version in the Dockerfile, however, it would be good to check if we are using deprecated or no longer recommended API (where are the Traefik Release Notes?), since there were many releases between 2.2.11 and 2.9.6

https://doc.traefik.io/traefik/deprecation/releases/

https://doc.traefik.io/traefik/deprecation/features/

Question: Is there some kind of service like dependabot/pyup but to update versions in Dockerfile/docker-compose.yml? Which are?

browniebroke:
> Question: Is there some kind of service like dependabot/pyup but to update versions in Dockerfile/docker-compose.yml? Which are?

[Dependabot supports Docker](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem). The question is whether dependabot works with Dockerfile that include Jinja tags. It's not really an issue for Traefik, but would be nice it worked across the board...

luzfcb:
Fixed on https://github.com/cookiecutter/cookiecutter-django/pull/4124

PR #4124, opened 2023-02-02
Update template docker images using Dependabot

browniebroke:
## Description

We're not currently doing it, which means we might be running some outdated images, as noted in #4119. I've tested it [on my fork](https://github.com/browniebroke/cookiecutter-django/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+compose) and it seems that Depenabot can handle Dockerfile with Jinja templates

Checklist:

## Rationale

To keep our template up to date

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,56 @@ updates:
       interval: "daily"
     labels:
       - "update"
+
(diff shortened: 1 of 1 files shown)


luzfcb (approved):
LGTM

PR #4128, opened 2023-02-03
Fix django-webpack-loader setup when running tests

### LINKED ISSUE #4126: Webpack and django.test.client.Client
Using the django test client in combination with webpack raises an error during tests because the stats file is missing. 

```python
webpack_loader.exceptions.WebpackLoaderBadStatsError: The stats file does not contain valid data. Make sure webpack-bundle-tracker plugin is enabled and try to run webpack again.
```

Seems related to https://github.com/django-webpack/django-webpack-loader/issues/187

browniebroke:
## Description

While running some tests with Django's test Client, Webpack fails to load the stat file and cannot find the bundles. This is known limitation of django-webpack-loader which is [documented in their readme](https://github.com/django-webpack/django-webpack-loader#usage-in-tests).

Checklist:

## Rationale

I couldn't make the first suggested solution work, but the second one does the trick. I like the 2nd solution better as it's more robust across environments, it works the same way locally as in CI, without the need to fiddle with files.

Fix #4126


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_views.py
```diff
@@ -91,6 +92,13 @@ def test_authenticated(self, user: User, rf: RequestFactory):
 
         assert response.status_code == 200
 
+    def test_authenticated_integration(self, user: User, client: Client):
+        client.force_login(UserFactory())
+        response = client.get(
+            reverse("users:detail", kwargs={"username": user.username})
+        )
+        assert response.status_code == 200
```
browniebroke, line 100: I think it would be nice to include some tests covering this, but I'm not happy where it is right now... 

I'm thinking that it might make more sense to be in a new file called `test_integration.py` (or something like this) to signal that it's for testing the whole request/response, from URL, view, form all the way to the DB, and that these are expected to be slower. We could even add a pytest marker for them.

That starting to sound like a bigger change, and I might save it for another PR. Thoughts?

foarsitter: Is it an idea to move it to utils and call it `test_webpack_loader`?

foarsitter:
Good to follow the recommendations of django-webpack-loader. Thanks for your fast response.

{{cookiecutter.project_slug}}/config/settings/test.py
```diff
@@ -29,5 +29,13 @@
 # ------------------------------------------------------------------------------
 TEMPLATES[0]["OPTIONS"]["debug"] = True  # type: ignore # noqa F405
 
+{%- if cookiecutter.frontend_pipeline == 'Webpack' %}
+# django-webpack-loader
+# ------------------------------------------------------------------------------
+WEBPACK_LOADER["DEFAULT"][  # noqa F405
+    "LOADER_CLASS"
+] = "webpack_loader.loader.FakeWebpackLoader"
```
foarsitter, line 37: I was surprised to see that they had already had an loader for this, how could I have missed that!? Great job for fixing it upstream!

browniebroke: I've just added it :)

foarsitter: I like it

PR #4130, opened 2023-02-04
Add note about which service to request when running locally with Docker & Webpack or Gulp

duffn:
## Description

Add a section about using Webpack or Gulp with Docker locally in order to make it clear which service should be addressed in order to see proper styles.

Checklist:

## Rationale

It was unclear to me (per https://github.com/cookiecutter/cookiecutter-django/issues/4129) that I needed to request the `node` service and not the `django` service when using the Webpack `frontend_pipeline` option. One should use the `django` service when using any other option, but using the `node` service when using the Webpack or Gulp option.

### DIFF
--- a/docs/developing-locally-docker.rst
+++ b/docs/developing-locally-docker.rst
@@ -213,6 +213,11 @@ By default, it's enabled both in local and production environments (``local.yml`
 
 .. _`Flower`: https://github.com/mher/flower
 
+Using Webpack
+~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


browniebroke (commented):
Thanks for the addition, just a note that this is also true for Gulp. Could you please rephrase to cover both? Much appreciated :+1:

browniebroke (approved):

villancikos:
Hi guys, this happened to me and because I am not using docker but I am using Webpack locally, I suggest you also add this text to the following https://cookiecutter-django.readthedocs.io/en/latest/developing-locally.html#setting-up-development-environment

villancikos:
Just sent one: https://github.com/cookiecutter/cookiecutter-django/pull/4413

PR #4163, opened 2023-02-15
Trim leading and trailing space in `domain_name` and `email`

thuvh:
## Description

remove trailing space in domain_name and email

Checklist:

## Rationale

Before: if I input domain_name with "   example.com  " then there are some settings like that
```
ADMINS = [("""root""", "root@   example.com  ")]
ALLOWED_HOSTS = env.list("DJANGO_ALLOWED_HOSTS", default=["   example.com  "])
```
After: 
```
ADMINS = [("""root""", "root@example.com")]
ALLOWED_HOSTS = env.list("DJANGO_ALLOWED_HOSTS", default=["example.com"])
```

### DIFF
--- a/hooks/pre_gen_project.py
+++ b/hooks/pre_gen_project.py
@@ -17,6 +17,14 @@ INFO = "\x1b[1;33m [INFO]: "
 HINT = "\x1b[3;33m"
 SUCCESS = "\x1b[1;32m [SUCCESS]: "
 
+# The content of this string is evaluated by Jinja, and plays an important role.
+# It updates the cookiecutter context to trim leading and trailing spaces
+# from domain/email values
+"""
+{{ cookiecutter.update({ "domain_name": cookiecutter.domain_name | trim }) }}
+{{ cookiecutter.update({ "email": cookiecutter.email | trim }) }}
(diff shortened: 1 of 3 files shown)


hooks/pre_gen_project.py
```diff
@@ -80,3 +80,8 @@
         "Mail Service for sending emails."
     )
     sys.exit(1)
+
+"""
+{{ cookiecutter.update({ "domain_name": cookiecutter.domain_name | trim }) }}
+{{ cookiecutter.update({ "email": cookiecutter.email | trim }) }}
+"""
```
browniebroke, line 87: That is pretty hacky, but nice trick... Since it's not very obvious what this does, could we:

- Add a comment explaining that it's not just a string, it serves a purpose.
- Add a test for it: generate a project with extra spaces in the domain/email and check that the outpur doesn't include them

Should we move this string more at the top of the file? If we start using these values in the file, I guess they wouldn't be cleaned yet?

thuvh: ok, let's me update this pr.

browniebroke (approved):

PR #4166, opened 2023-02-17
Disable caching for linter job on GitHub actions

DelphineLemire:
Hello, 
It is my first contribution .
I use cookiecutter-django for training .

## Description

When I push my project to github, i have an Linter's issue . 
image

Checklist:

## Rationale

My fix is based on the installation instructions used by pytest. that seems good

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
@@ -30,10 +30,6 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.10"
-          cache: pip
-          cache-dependency-path: |
-            requirements/base.txt
-            requirements/local.txt
 
       {%- if cookiecutter.open_source_license != 'Not open source' %}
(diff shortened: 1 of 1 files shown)


vsha96:
Fixed this in my project in same way 👍 
I think, we need to merge it

vsha96 (approved):
Fixes issue with `no file in ... matched to **/requirements.txt` on CI workflow

{{cookiecutter.project_slug}}/.github/workflows/ci.yml
```diff
@@ -31,6 +31,9 @@ jobs:
         with:
           python-version: "3.10"
           cache: pip
+          cache-dependency-path: |
+            requirements/base.txt
+            requirements/local.txt
```
browniebroke, line 36: The `ache-dependency-path` was removed in #4151 because the following actions aren't installing from these requirements, but I didn't realise that it breaks because the default doesn't match anything.

We should just disable caching in this case:

```suggestion
```

PR #4201, opened 2023-03-09
Fix the omit configuration for coverage

mpsantos:
## Description

When running coverage, directories _tests_ and _migrations_ still appear on the report, despite being included in the omit configuration.
Coverage documentation states that: 
`* matches any number of file name characters, not including the directory separator`
The directories _tests_ and _migrations_ are not omitted because of the missing directory separator.

Checklist:

## Rationale

Configuration for coverage is not working as intended.

### DIFF
--- a/{{cookiecutter.project_slug}}/setup.cfg
+++ b/{{cookiecutter.project_slug}}/setup.cfg
@@ -35,6 +35,6 @@ ignore_errors = True
 
 [coverage:run]
 include = {{cookiecutter.project_slug}}/**
-omit = *migrations*, *tests*
+omit = */migrations/*, */tests/*
 plugins =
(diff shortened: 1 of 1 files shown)


browniebroke (approved):
Good catch. I probably missed that with the [upgrade to Coverage 7](https://coverage.readthedocs.io/en/stable/changes.html#version-7-0-0b1-2022-12-03):

> - Changes to file pattern matching, which might require updating your configuration:
>    - Previously, * would incorrectly match directory separators, making precise matching difficult. This is now fixed, closing [issue 1407](https://github.com/nedbat/coveragepy/issues/1407).
>    - Now ** matches any number of nested directories, including none.

PR #4234, opened 2023-03-28
Social google login added

M-Usman-Dev:
## Description

Checklist:


browniebroke:
I'm going to assume that this was open by mistake. There is no interest in adding Google login to the default template, and even if there was, this is vastly incomplete as it stands (e.g. introducing a new dependency without proper installation).

Issue #4209, opened 2023-03-15
Serving media files with NGINX issue


maponzzi:
Hello Everyone,
I'm trying to enable serving media files in production using nginx, as @arrys solution on his PR [#2457](https://github.com/cookiecutter/cookiecutter-django/pull/2457) but it doesn't works for me.

I implemented manually this PR because when I ran cookiecutter installation, and I chose None as cloud server doesn't added nginx as media server. (I don't know why? because it's suppose this PR its already added to the master branch)

The only one modification I made was in /compose/production/django/Dockerfile, I added this line:
`RUN mkdir ${APP_HOME}/proveemax/media && chown django:django ${APP_HOME}/proveemax/media`

in order to create my own media folder with django:django as the owner:group, because in the PR doesn't include the media folder creation and when I build containers this media folder was created with root:root ownership, so my django app crashed when I tried to upload any file because doesn't have permissions to write on that folder.

At this moment, with my own media folder (as mentioned above) I can upload media files, even when I connect to nginx container, media files are created in the right folder, (/usr/share/nginx/media/ ready to be served) which means media volume is also working.

I think traefik router (web-media-router) it's not working, because when I request a media file I just got a 404 not found, and I can't see anything on nginx logs (I guess I should see the request at least)

Any Idea how to solve this ? or how to test if the traefik router it's working ?

_**/compose/production/traefik/traefik.yml**_
```yml
log:
  level: INFO

entryPoints:
  web:
    # http
    address: ":80"
    http:
      # https://docs.traefik.io/routing/entrypoints/#entrypoint
      redirections:
        entryPoint:
          to: web-secure

  web-secure:
    # https
    address: ":443"

certificatesResolvers:
  letsencrypt:
    # https://docs.traefik.io/master/https/acme/#lets-encrypt
    acme:
      email: "..."
      storage: /etc/traefik/acme/acme.json
      # https://docs.traefik.io/master/https/acme/#httpchallenge
      httpChallenge:
        entryPoint: web

http:
  routers:
    web-secure-router:
      rule: "Host(`proveemaxindustrial.com`) || Host(`www.proveemaxindustrial.com`)"
      entryPoints:
        - web-secure
      middlewares:
        - csrf
      service: django
      tls:
        # https://docs.traefik.io/master/routing/routers/#certresolver
        certResolver: letsencrypt

    web-media-router:
      rule: "Host(`proveemaxindustrial.com`) && PathPrefix(`/media/`)"
[...]

PR #4244, opened 2023-04-01
Fix traefik rule priority for media router

arrys:
## Description

Traefik looks at the length of rules to determine priority when no explicit priority is set.
The `web-media-router` needs to always have priority over the` web-secure-router` in order for the media files to be served.

The` web-secure-router` could end up having priority over the `web-media-router` when the `www` subdomain is used.
```
    web-secure-router:
      {%- if cookiecutter.domain_name.count('.') == 1 %}
      rule: "Host(`{{ cookiecutter.domain_name }}`) || Host(`www.{{ cookiecutter.domain_name }}`)"
      {%- else %}
      rule: "Host(`{{ cookiecutter.domain_name }}`)"
      {%- endif %}
```
Let's say `n =  len(cookiecutter.domain_name)`.
Here the `web-secure-router` would have a priority of `18 + 2 * n)` or, in the second branch, `8 + n`.
Because the rule for the `web-media-router` only has one option, we have a priority of `34 + n`.
```
  web-media-router:
     rule: "Host(`{{ cookiecutter.domain_name }}`) && PathPrefix(`/media/`)"
```
In some cases where `n > 16`, this will mean the priority of the  `web-media-router` will be lower than the priority of the `web-secure-router`.
This in turn breaks the routing, as nginx will never be reached and media files will never be served.

This fix adds the `www` subdomain to the `web-media-router` if it's also present in the `web-secure-router` rule.

Checklist:

The tests and documentation don't need changes as far as I am aware.

## Rationale

This should address one issue discovered by @edwardigarashi in #4209.
I am not yet sure if it solves the issue completely.


{{cookiecutter.project_slug}}/compose/production/traefik/traefik.yml
```diff
@@ -60,7 +60,11 @@ http:
     {%- if cookiecutter.cloud_provider == 'None' %}
 
     web-media-router:
+      {%- if cookiecutter.domain_name.count('.') == 1 %}
+      rule: "(Host(`{{ cookiecutter.domain_name }}`) || Host(`www.{{ cookiecutter.domain_name }}`))&& PathPrefix(`/media/`)"
```
browniebroke, line 64:
```suggestion
      rule: "(Host(`{{ cookiecutter.domain_name }}`) || Host(`www.{{ cookiecutter.domain_name }}`)) && PathPrefix(`/media/`)"
```

browniebroke (approved):

browniebroke:
Thanks for the fix @arrys!

PR #4247, opened 2023-04-02
Remove secure browser setting

MortenKaae:
## Description

The setting `SECURE_BROWSER_XSS_FILTER` was deprecated and removed in Django 4.0, so I don't think it makes sense to include this in the base-project settings file.

You can view the deprecation in the Django 4.0 release notes [here](https://docs.djangoproject.com/en/4.1/releases/4.0/#securitymiddleware-no-longer-sets-the-x-xss-protection-header).

Checklist:

## Rationale

This project is for Django 4.0 as mentioned in its README, which is why it shouldn't include settings that were deprecated in Django 4.0.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -22,7 +22,6 @@ DATABASE_URL                            DATABASES                   auto w/ Dock
 DJANGO_ADMIN_URL                        n/a                         'admin/'                                       raises error
 DJANGO_DEBUG                            DEBUG                       True                                           False
 DJANGO_SECRET_KEY                       SECRET_KEY                  auto-generated                                 raises error
-DJANGO_SECURE_BROWSER_XSS_FILTER        SECURE_BROWSER_XSS_FILTER   n/a                                            True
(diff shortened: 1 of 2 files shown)


docs/settings.rst
```diff
@@ -22,7 +22,6 @@ DATABASE_URL                            DATABASES                   auto w/ Dock
 DJANGO_ADMIN_URL                        n/a                         'admin/'                                       raises error
 DJANGO_DEBUG                            DEBUG                       True                                           False
 DJANGO_SECRET_KEY                       SECRET_KEY                  auto-generated                                 raises error
-DJANGO_SECURE_BROWSER_XSS_FILTER        SECURE_BROWSER_XSS_FILTER   n/a                                            True
```
browniebroke, line 25: That was wrong actually, since there is no env variable for `DJANGO_SECURE_BROWSER_XSS_FILTER`...

browniebroke (approved):
Good catch, thanks for the contribution!

PR #4250, opened 2023-04-03
fix: Syntax for ignoring specific noqa errors

adinhodovic:
Noticed when started using pygrep-hooks. Syntax is `noqa: <code>` with a colon.

## Description

Fix the noqa syntax for ignoring specific codes.

Checklist:


luzfcb:
@adinhodovic Great. Thanks for the contribution. 🚢 :

PR #4028, opened 2022-12-27
Upgrade to Django 4.1

foarsitter:
With the merge of https://github.com/django-extensions/django-extensions/pull/1775 we are close to upgrading to Django 4.1. The new release of django-extensions will resolve #3941.

This pull request gets everything ready for upgrading tot Django 4.1 and can be merged after the release of django-extensions and will close #3826.

PR #4255, opened 2023-04-04
Replace `runserver_plus` with `runserver`

### LINKED ISSUE #3941: Docker Django 4.1 template reload behavior
Using Django 4.1 under Docker gives trouble in development with the changed default behavior of the template loading. Templates are cached until restart since the cached loader is added as default: https://docs.djangoproject.com/en/4.1/releases/4.1/#templates

My current workaround is adding the next snippet to local.py

```
TEMPLATES[0]["OPTIONS"]["loaders"] = [
    (
        "django.template.loaders.filesystem.Loader",
        [APPS_DIR / "templates"],
    ),
    "django.template.loaders.app_directories.Loader",
]
```

foarsitter:
The `runserver_plus` command does not handle template reloads correctly as discussed in #3941. This pull requests uses the build-in `runserver` command to serve django. 

## Description

Checklist:

## Rationale

The `runserver_plus` command is incompatible with django >= 4.1, so this unblocks #4028 

Fix #3941

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -182,7 +182,7 @@ def handle_js_runner(choice, use_docker, use_async):
             dev_django_cmd = (
                 "uvicorn config.asgi:application --reload"
                 if use_async
-                else "python manage.py runserver_plus"
+                else "python manage.py runserver"
             )
             scripts.update(
                 {


browniebroke:
Thanks! 🎉

Issue #3915, opened 2022-10-25
Python 3.11


foarsitter:
To my big surprise my codebase passed all tests with Python 3.11. With python 3.10 merged a view hours ago I hope we can adopt python 3.11 a little faster so we can benefit the faster cpython effort.

jugurtha114:
i can't wait how fast django will be with the new 3.11 version of python

foarsitter:
Just go ahead and try it! Posting your experiences over here will be valuable.

Andrew-Chen-Wang:
This should be closed.

browniebroke:
Fixed in #4256

PR #4256, opened 2023-04-04
Upgrade Python to version 3.11 (Faster CPython)

foarsitter:
With Django updated to 4.1 we can upgrade to Python 3.11. 

Cloeses #3915 
## Description

Checklist:


docs/deployment-on-pythonanywhere.rst
```diff
@@ -34,7 +34,7 @@ Make sure your project is fully committed and pushed up to Bitbucket or Github o
 
     git clone <my-repo-url>  # you can also use hg
     cd my-project-name
-    mkvirtualenv --python=/usr/bin/python3.10 my-project-name
+    mkvirtualenv --python=/usr/bin/python3.11 my-project-name
```
browniebroke, line 37: It looks like Python anywhere doesn't support 3.11 yet? I don't see it mentioned in their documentation...

browniebroke (approved):

PR #4361, opened 2023-05-30
Install Django and DRF stubs with `compatible-mypy` extra (as per offical recommendation)

browniebroke:
## Description

Django stubs and DRF stubs are by default leting us install any version of mypy as long as it's not too old. The version restriction can be made more constrained with an extra require specifier.

Checklist:

## Rationale

This might make upgrades more annoying, with the tradeoff that we're more sure to install a compatible version. Overpinning has its drawbacks, especially when the project isn't updated in a timely manner, but these 2 projects are well maintained (at least now). These are optional and could be easily reverted if it's adding too much friction.


mpsantos:
This will break with mypy 1.3.0 (#4327)

browniebroke:
Correct, and that's actually the point. Right now, our build doesn't fail and we might ship with a version of `mypy` which is incompatible with `django-stubs`. Incompatibilities wouldn't show up on the vanilla project that we generate, but that doesn't mean that a project generated with Cookiecutter Django won't be impacted shortly after, when features start to be added to it.

PR #4389, opened 2023-06-13
Add djLint for HTML formatting and linting

### LINKED ISSUE #3827: Linter/Formatter for Django templates
## Description

I propose we add either [djlint](https://github.com/Riverside-Healthcare/djLint) or [djhtml](https://github.com/rtts/djhtml) as a local requirement and pre-commit hooks.

## Rationale

I think this is a missing part of the tooling available for the cookiecutter-django ecosystem. What do you think?

monosans:
## Description

Checklist:

## Rationale

Closes #3827


.pre-commit-config.yaml
```diff
@@ -44,6 +44,14 @@ repos:
     hooks:
       - id: flake8
 
+  - repo: https://github.com/Riverside-Healthcare/djLint
+    rev: v1.31.0
+    hooks:
+      - id: djlint-reformat-jinja
+        types_or: [html]
+      - id: djlint-jinja
+        types_or: [html]
```
browniebroke, line 53: That's nice that we can lint/reformat templated files, but is that actually working? Asking as we exclude the whole templated project on the first line of this config:

https://github.com/cookiecutter/cookiecutter-django/blob/7ad2e1fdae82e011fa823f5f42590bf6ce3638ee/.pre-commit-config.yaml#L1

tests/test_cookiecutter_generation.py
```diff
@@ -239,6 +239,32 @@ def test_django_upgrade_passes(cookies, context_override):
...
+@auto_fixable
+@pytest.mark.parametrize("context_override", SUPPORTED_COMBINATIONS, ids=_fixture_id)
+def test_djlint_check_passes(cookies, context_override):
+    """Check whether generated project passes djLint --check."""
+    result = cookies.bake(extra_context=context_override)
+
+    try:
+        sh.djlint("--check", ".", _cwd=str(result.project_path))
+    except sh.ErrorReturnCode as e:
+        pytest.fail(e.stdout.decode())
```
browniebroke, line 265: I noticed that the lint isn't auto-fixable but the check is. Was going to suggest to change that, but actually it makes makes sense after looking at their docs. Nice one!

browniebroke (approved):
Excellent work!

.pre-commit-config.yaml
monosans: Yes, thank you, I hadn't noticed. I ended up removing the pre-commit hook after all.

browniebroke (approved):

PR #4396, opened 2023-06-15
Configure Flake 8 to be compatible with Black (88 char line limit)

JamesParrott:
Make Flake8 accept Black's automatic reformatting, 

https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8
"
[Flake8](https://pypi.org/p/flake8/) is a code linter. It warns you of syntax errors, possible bugs, stylistic errors, etc. For the most part, Flake8 follows [PEP 8](https://www.python.org/dev/peps/pep-0008/) when warning about stylistic errors. There are a few deviations that cause incompatibilities with Black.
Also, as like with isort, flake8 should be configured to allow lines up to the length limit of 88, Black’s default. This explains max-line-length = 88.
"
## Description

Add the args to Flake 8 that Black recommends.

## Rationale

Avoids doom loops, in which commits are impossible, due to Black's automatic changes being rejected by Flake 8

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,7 @@ repos:
     rev: 6.0.0
     hooks:
       - id: flake8
+        args: ['--max-line-length=88']
 
 ci:
   autoupdate_schedule: weekly
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,7 @@ repos:
     rev: 6.0.0
     hooks:
       - id: flake8
-        args: ['--max-line-length=88']
+        args: ["--max-line-length=88"]
 
 ci:
   autoupdate_schedule: weekly


JamesParrott:
I apologise.  This change unlocks a host of further frustrations with the CI tools.  I don't understand the value most of them add, let alone why that justifies a plethora of often conflicting linters and auto formatters.  They are discouraging me from committing often in my project, so I've simply nerfed almost everything and am using ruff and https://github.com/pre-commit/pre-commit-hooks only.  

Even then, the only error is about line lengths, which I really couldn't care less about.  My fellow Python fans!  Is it possible we spent so long building fantastic software tools, to avoid irritating each other personally and having to have a conversation, that we instead achieved automatic tools irritate us and contradict each other.  

At the end of the day, even the highly efficient, fantastic Rust implementation, Ruff, is nothing more than a little minded Hobgloblin.

I'm quite happy to add a commit to this PR including my new pre-commit file if you want me to, but I suspect that merely opens a bigger can of worms.

browniebroke:
I don't understand why you have this problem, the flake8 line length and black line length are both configured to 119, in `.flake8` and `pyproject.toml` respectively

JamesParrott:
You star Brownie - thankyou!  A missing .flake8 file was exactly the problem.

PR #4413, opened 2023-06-23
Added webpack tip for port 3000 instead of 8000 in developing-locally…

villancikos:
….rst

## Description

Checklist:

### DIFF
--- a/docs/developing-locally.rst
+++ b/docs/developing-locally.rst
@@ -80,9 +80,11 @@ First things first.
 
     $ python manage.py runserver 0.0.0.0:8000
 
-or if you're running asynchronously: ::
+   or if you're running asynchronously: ::
 
     $ uvicorn config.asgi:application --host 0.0.0.0 --reload --reload-include '*.html'
+
+   If you've opted for Webpack or Gulp as frontend pipeline, please see the :ref:`dedicated section <bare-metal-webpack-gulp>` below.
 
 .. _PostgreSQL: https://www.postgresql.org/download/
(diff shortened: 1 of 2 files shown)


docs/developing-locally.rst
```diff
@@ -187,6 +187,12 @@ If you've opted for Gulp or Webpack as front-end pipeline, the project comes con
 
 .. note:: The task will start 2 processes in parallel: the static assets build loop on one side, and the Django server on the other. You do NOT need to run Django as your would normally with ``manage.py runserver``.
 
+Using Webpack or Gulp
+~~~~~~~~~~~~~~~~~~~~~
+
+When using Webpack or Gulp as the ``frontend_pipeline`` option, you should access your application at the address of the ``node`` service in order to see your correct styles. This is http://localhost:3000 by default. When using any of the other ``frontend_pipeline`` options, you should use the address of the ``django`` service, http://localhost:8000.
```
browniebroke, line 193: Thanks for the clarification. The previous section "Sass Compilation & Live Reloading" might be suited for that, since it's about Webpack or Gulp, but I realise that it is easy to miss. Maybe we could merge both sections?

What do you think of:

1. Renaming the previous section to "Using Webpack or Gulp"
2. Add numbered item at the end saying that the site should be accessed at `http://localhost:3000`, to go through the node service

PS: to keep the counter of the ordered list and prevent it to reset, you might have to indent the `.. note::` a bit

browniebroke (approved):

PR #4418, opened 2023-06-26
Remove prettier webpack in pre-commit

Andrew-Chen-Wang:
## Description

Remove prettier webpack integration in pre-commit

Checklist:

## Rationale

I use nvm and PyCharm, so this pre-commit hook does not apply due to different IDE and because I'm not using node based frontend.

### DIFF
--- a/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
+++ b/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
       - id: check-case-conflict
       - id: check-docstring-first
       - id: detect-private-key
-{%- if cookiecutter.frontend_pipeline != "Webpack" or cookiecutter.frontend_pipeline != "Gulp" or cookiecutter.use_pycharm %}
+{%- if cookiecutter.frontend_pipeline in ["Webpack", "Gulp"] %}
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v3.0.0-alpha.9-for-vscode
     hooks:
--- a/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
+++ b/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
@@ -17,13 +17,15 @@ repos:
       - id: check-docstring-first
       - id: detect-private-key
 {%- if cookiecutter.frontend_pipeline in ["Webpack", "Gulp"] %}
+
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -16,14 +16,14 @@ repos:
       - id: check-case-conflict
       - id: check-docstring-first
       - id: detect-private-key
-
+{%- if cookiecutter.frontend_pipeline != "Webpack" or cookiecutter.frontend_pipeline != "Gulp" or cookiecutter.use_pycharm %}
```
browniebroke, line 19: I don't understand. What's the problem with having this pre-commit hook? Prettier also formats YAML, markdown and JSON, which may be present in your repos even if you don't use a node based frontend. I use PyCharm too and I still find value in including this...

Andrew-Chen-Wang: I don't have node globally (I use nvm), and even if I did use it, my frontend repo is usually separate from Django repo, so node wouldn't be in my PATH. Initially, I didn't really understand why pre commit was failing till I saw it try to activate node, so that was the primary reason for opening this: confusion.

But I guess i'm stuck in 2012, cuz I still use bootstrap mostly thanks to cookiecutter-django haha

browniebroke: I also use nvm and I don't have any issues with this. I do however have a global node, coming from Homebrew, which now that I think about it, might be a problem with [node based pre-commit](https://pre-commit.com/#node) hooks 🤔  Or maybe not. 


I understand better where the friction comes from regarding Webpack and Gulp, however I don't see how it relates to the usage of PyCharm...

browniebroke: Did you mean this?

```suggestion
{%- if cookiecutter.frontend_pipeline in ["Webpack", "Gulp"] %}
```

Andrew-Chen-Wang: it looked like a vscode only thing given the name, but if not, then for sure can remove

browniebroke: Yes, not sure why the tags include vscode in the name, they are not even present anymore in the upstream repo (the pre-commit repo is just a mirror).

PR #4429, opened 2023-06-29
Improve type hints for `UserSerializer`

browniebroke:
## Description

I recently came across the fact that `djangorestframework-stubs` supports using generics in model serializers, which has the effect of improving type hints: `self.instance` is known to be an instance of the model instead of `Any`.

It would be nice to show that feature in the only serializer that we include, the `UserSerializer`.

Checklist:

## Rationale

Improve discoverability of a nice feature, improve type checking, improve auto-completion.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/api/serializers.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/api/serializers.py
@@ -1,12 +1,7 @@
-from __future__ import annotations
-
-import typing
-
 from django.contrib.auth import get_user_model
 from rest_framework import serializers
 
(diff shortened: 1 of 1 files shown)


foarsitter:
It is good to know about this feature. Do you have some more background on this?

browniebroke:
I've stumbled upon that in a [closed issue on the DRF-stubs](https://github.com/typeddjango/djangorestframework-stubs/issues/164#issuecomment-1613314994) and I'm intending to send a pull request to document these better in the upstream project. For now, I'm learning more about them in one of my projects.

browniebroke:
Unless there is any objections, I'm going to merge this in the next couple of days, it seems pretty safe AFAIK

foarsitter (approved):

PR #4431, opened 2023-07-01
Add Postgresql 15

sanchimenea:
## Description

Add PostgreSQL 15 to the versions of choice. 

Checklist:

## Rationale

Psycopg 3 supports PostgreSQL from version 10 to 15.


browniebroke (approved):
LGTM! Thanks for keeping us up to date. 

PS: reminds me that we should drop older versions of PostgreSQL as [Django 4.2 only supports 12 and higher](https://docs.djangoproject.com/en/4.2/releases/4.2/#dropped-support-for-postgresql-11). Can be done in a separate PR though.

PR #4434, opened 2023-07-02
Add missing trailing space in EMAIL_SUBJECT_PREFIX

tildebox:
## Description

A missing space at the end of `EMAIL_SUBJECT_PREFIX`.

Checklist:

## Rationale

It doesn't look good when the `EMAIL_SUBJECT_PREFIX` is displayed with the subject without a space.

[The documentation](https://docs.djangoproject.com/en/dev/ref/settings/#email-subject-prefix) also contains the following note:

> "You’ll probably want to include the trailing space."


browniebroke:
Good catch, thanks for the fix!

PR #4450, opened 2023-07-10
Change how we remove prettier pre-commit hook from the config

browniebroke:
## Description

Removes the prettier pre-commit hook in post generate hooks rather than by templating the pre-commit config config as per #4418, to make sure we can run `pre-commit autoupgrade` on the config in the template. 

Checklist:

## Rationale

The pre-commit config needs to be valid to be kept up to date automatically, and [the autoupgrade workflow](https://github.com/cookiecutter/cookiecutter-django/actions/workflows/pre-commit-autoupdate.yml) started failing shortly after https://github.com/cookiecutter/cookiecutter-django/pull/4418 was merged (FYI @Andrew-Chen-Wang).

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -207,6 +207,24 @@ def handle_js_runner(choice, use_docker, use_async):
         remove_gulp_files()
 
 
+def remove_prettier_pre_commit():
+    with open(".pre-commit-config.yaml", "r") as fd:
+        content = fd.readlines()
+
+    removing = False
+    new_lines = []
+    for line in content:
+        if removing and "- repo:" in line:
+            removing = False
+        if "mirrors-prettier" in line:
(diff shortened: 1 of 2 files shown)


hooks/post_gen_project.py
```diff
@@ -207,6 +207,24 @@ def handle_js_runner(choice, use_docker, use_async):
...
+    for line in content:
+        if removing and "- repo:" in line:
+            removing = False
+        if "mirrors-prettier" in line:
+            removing = True
+        if not removing:
+            new_lines.append(line)
+
+    with open(".pre-commit-config.yaml", "w") as fd:
+        fd.writelines(new_lines)
```
browniebroke, line 225: I've tried to use PyYAML, but I'm struggling to control the formatting of the output. This should hopefully do the job.

PR #4455, opened 2023-07-11
fix /tmp bind mount

masavini:
minor fix, the previous mount (host `~/.ssh` into contaner `/tmp`) had no sense at all...


browniebroke (approved):

masavini:
it would be nice to fix mount sources for windows users, as well, but (un)fortunately i'm not... :)

PR #4457, opened 2023-07-11
changed settings.py vars to reflect django-4.2

### LINKED ISSUE #4443: Migrate the media & static settings to the new unified `STORAGES`
## Description

Django 4.2 introduced a unified [`STORAGES` setting](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES) which unifies static & media setup and simplifies adding a custom storage that might be required by an application (e.g. private media).

## Rationale

The [`DEFAULT_FILE_STORAGE`](https://docs.djangoproject.com/en/4.2/ref/settings/#default-file-storage) and [`STATICFILES_STORAGE`](https://docs.djangoproject.com/en/4.2/ref/settings/#staticfiles-storage) settings are now both deprecated and we should migrate to the new setting before Django 5.1

chipndell:
## Description

Checklist:

## Rationale

Fixes #4443

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -106,34 +106,34 @@ AZURE_CONTAINER = env("DJANGO_AZURE_CONTAINER_NAME")
 # STATIC & MEDIA
 # ------------------------
 STORAGES = {
-{%- if cookiecutter.use_whitenoise == 'y' -%}
+{%- if cookiecutter.use_whitenoise == 'y' %}
     "staticfiles": {
         "BACKEND": "whitenoise.storage.CompressedManifestStaticFilesStorage",
-    }
-{%- elif cookiecutter.cloud_provider == 'AWS' -%}
+    },
+{%- elif cookiecutter.cloud_provider == 'AWS' %}
     "default": {
         "BACKEND": "{{cookiecutter.project_slug}}.utils.storages.MediaRootS3Boto3Storage",
     },
     "staticfiles": {
         "BACKEND": "{{cookiecutter.project_slug}}.utils.storages.StaticRootS3Boto3Storage",
-    }
-{%- elif cookiecutter.cloud_provider == 'GCP' -%}
+    },
+{%- elif cookiecutter.cloud_provider == 'GCP' %}
     "default": {
         "BACKEND": "{{cookiecutter.project_slug}}.utils.storages.MediaRootGoogleCloudStorage",
     },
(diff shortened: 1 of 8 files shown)


{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -62,6 +62,8 @@
 # https://docs.djangoproject.com/en/dev/ref/middleware/#x-content-type-options-nosniff
 SECURE_CONTENT_TYPE_NOSNIFF = env.bool("DJANGO_SECURE_CONTENT_TYPE_NOSNIFF", default=True)
 
+STORAGES = dict()
```
browniebroke, line 65: I think this overrides the [default value](https://docs.djangoproject.com/en/4.2/ref/settings/#storages). If the user selects `cloud_provider='None'`, none of the branches below are hit, and the setting will probably be invalid.

browniebroke (commented):
Thanks for your contribution. I've started to tweak the code style to conform to black, but then I spotted a potential issue. WDYT?

{{cookiecutter.project_slug}}/config/settings/production.py
```diff
@@ -103,35 +103,49 @@
...
-STATIC_URL = f"https://storage.googleapis.com/{GS_BUCKET_NAME}/static/"
-{% elif cookiecutter.cloud_provider == 'Azure' -%}
-STATICFILES_STORAGE = "{{cookiecutter.project_slug}}.utils.storages.StaticRootAzureStorage"
-STATIC_URL = f"https://{AZURE_ACCOUNT_NAME}.blob.core.windows.net/static/"
-{% endif -%}
+STORAGES = {
+{%- if cookiecutter.use_whitenoise == 'y' %}
+    "staticfiles": {
+        "BACKEND": "whitenoise.storage.CompressedManifestStaticFilesStorage",
+    },
```
browniebroke, line 112: I've added a `profile_picture` field to the user model and I tried to deploy with `cloud_provider = 'None'` and `use_whitenoise = 'y'` and got this error when trying to upload an image:

```
InvalidStorageError: Could not find config for 'default' in settings.STORAGES.
```

We really need to specify both default and staticfiles storages:

```suggestion
    "default": {
        "BACKEND": "django.core.files.storage.FileSystemStorage",
    }
    "staticfiles": {
        "BACKEND": "whitenoise.storage.CompressedManifestStaticFilesStorage",
    },
```

browniebroke:
Why did you close it? I think it's nearly there...

PR #4460, opened 2023-07-15
Change `MEDIA_URL` to an absolute URL in tests

itisnotyourenv:
## Description

I recently noticed that in cases where we have a default value for models' **ImageField** like  in the example below
```
class Book(models.Model):
    cover = models.ImageField(upload_to="images/", default=DEFAULT_IMAGE, verbose_name="Изображение")
```

and try to compare the field from the response with the expected field from a serializer
```
class TestUserViewSet:
    def test_get_user_by_id(self, api_client: APIClient, user: User, admin_user: User):
        url = reverse("api:users-detail", kwargs={"id": admin_user.id})
        expected_data = UserSerializer(admin_user).data

        response = api_client.get(url)

        assert response.status_code == status.HTTP_200_OK
        assert response.data.get("image") == expected_data.get("image")
```

we encounter a failure because Django adds a prefix  "http://testserver" to the ImageField and for other areas that use **MEDIA_URL**.
`{'image': 'http://testserver/media/default/user-logo.jpg'} != {'image': '/media/default/user-logo.jpg'}`

Checklist:

## Rationale
This will help to avoid incomprehensible errors in tests and these changes can be quite in demand.


browniebroke:
Thanks for your contribution. 

Is there a reason to use a fixture to set this? Since it's a contant value, it's better to set it in [the dedicated setting file](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/test.py) for tests, as it's where other similar ones are set. The `media_storage` fixture defined above is an exception because it uses a pytest fixture `tmpdir`.

Small note that this URL prefix is actually added by DRF (https://github.com/encode/django-rest-framework/blob/7658ffd71d4ad07ddada20b2b9538b889ec02403/rest_framework/fields.py#L1539-L1554), to make the URL absolute. Although, it's true that it's calling an API from Django: https://github.com/django/django/blob/0016a4299569a8f09ff24053ff2b8224f7fa4113/django/http/request.py#L201.

Which leads me to the fact that I'm slightly against this change, for the following reasons: 

- By setting the `MEDIA_URL` value to an absolute URI, you're hiding the fact that this transformation happens in the serialization, which is a kind of detail that I find useful to see in my tests. 
- I'd argue that your test should compare the result of the serialization to something more static: `assert response.data["image"] == "http://testserver/media/default/user-logo.jpg"`. The way it's written in your example feels to me like testing that `1 + 1 == 1 + 1`, if that makes sense.

Although the counter argument to the 1st point here is that the `MEDIA_URL` [is often absolute on prod](https://github.com/cookiecutter/cookiecutter-django/blob/69505fb7d6b709d9e030f785eb30f263623e0840/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/production.py#L124-L135) anyway... I guess we could set it to an absolute URL different from the default server URL? e.g. `https://testassets/media/`.

browniebroke:
Hey @itisnotyourenv 👋🏻 What do you think of my suggestions I mentioned above? Do you want to make any edits to this PR?

itisnotyourenv:
> Hey @itisnotyourenv 👋🏻 What do you think of my suggestions I mentioned above? Do you want to make any edits to this PR?

Sorry bro. I forgot about this PR, and didn't notice your last question. Thanks for merging

PR #4464, opened 2023-07-17
Display custom 403_csrf.html while getting a csrf error in prod

### LINKED ISSUE #3316: Both 403_csrf.html and 403.html should exist
## What happened?
The custom template (403.html) was not displayed for a 403 error. It was a csrf 403, not a regular 403 error.

The 403.html template should have been displayed.

## Additional details
Reproduce (all tests pass):

```python
from unittest.mock import patch

from django.test import Client, override_settings
from django.urls import reverse
from django.views import defaults as default_views


class TestCsrfTemplatesInUse:

    @override_settings(DEBUG=True)
    def test_403_csrf_debug_enabled(self):
        csrf_client = Client(enforce_csrf_checks=True)
        csrf_client.cookies.load({settings.CSRF_COOKIE_NAME: "notavalidtoken"})
        response = csrf_client.post(reverse("account_login"))
        # Show that CSRF_FAILURE_TEMPLATE is unexpectedly being used instead of using the project's custom "403.html".
        # https://github.com/django/django/blob/main/django/views/csrf.py#L15-L100
        assert 'CSRF verification failed. Request aborted.' in str(response.content)

    @override_settings(DEBUG=False)
    def test_403_csrf_debug_disabled(self):
        csrf_client = Client(enforce_csrf_checks=True)
        csrf_client.cookies.load({settings.CSRF_COOKIE_NAME: "notavalidtoken"})
        response = csrf_client.post(reverse("account_login"))
        # Show that CSRF_FAILURE_TEMPLATE is unexpectedly being used instead of using the project's custom "403.html".
        # https://github.com/django/django/blob/main/django/views/csrf.py#L15-L100
        assert 'CSRF verification failed. Request aborted.' in str(response.content)

    @patch("django.views.generic.base.TemplateView.get")
    def test_403_non_csrf(self, template_view_get):
        def get(request, *args, **kwargs):
            return default_views.permission_denied(request, Exception("Permission Denied"))
        template_view_get.side_effect = get

        client = Client()
        response = client.get(reverse("home"))
        # Show that non-csrf 403 errors are showing the 403.html template.
        assert response.templates[0].name == '403.html'
```

Fix:
```bash
$ cd {{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/
$ cp 403.html 403_csrf.html
```

infraredCoding:
## Description

As mentioned in #3316 , when `debug=False`, we still see the default 403 csrf template being displayed, not the custom `403.html` file.

Checklist:

- [x] I've made sure that tests are updated accordingly and no existing tests were affected

## Rationale

Adding a custom `403_csrf.html` which overrides the default csrf template. 
Fix #3316

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403_csrf.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403_csrf.html
@@ -1,0 +1,14 @@
+{% raw %}{% extends "base.html" %}
+
+{% block title %}Forbidden (403){% endblock title %}
+{% block content %}
+  <h1>Forbidden (403)</h1>
+  <p>
+    {% if exception %}
+      {{ exception }}
+    {% else %}
+      You're not allowed to access this page.
+    {% endif %}
+  </p>
+{% endblock content %}
+{%- endraw %}


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403_csrf.html
```diff
@@ -0,0 +1,14 @@
+{% raw %}{% extends "base.html" %}
+
+{% block title %}Forbidden (403){% endblock title %}
+{% block content %}
+  <h1>Forbidden (403)</h1>
+  <p>
+    {% if exception %}
+      {{ exception }}
+    {% else %}
+      You're not allowed to access this page.
```
browniebroke, line 10: Should we change the content? Because it's a bit difficult to see that it was a CSRF failure here and it was not a regular 403.

```suggestion
      CSRF verification failed. Request aborted.
```

browniebroke (approved):

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/403_csrf.html
infraredCoding: In the original issue, it says that the message "CSRF verification  failed..." was supposed to be "enexpected". So I was not sure if I should add this error message to the new file or keep it the way it is. But yeah if you want me to change it, I'll do it.

browniebroke: I see. Yes I guess one could argue either way... I'm not sure showing what I suggest is any more helpful for users actually. Let's leave it like this for now.

PR #4514, opened 2023-08-11
Update README.md

FatemeFouladkar:
## Description

Added a new article on how to make a custom Django cookiecutter template.

## Rationale

As many people might want to implement their own standards in their Django projects, they might need to design their own cookiecutter templates and add the tools they might usually use. This article shows how people can make their own Django templates using the cookiecutter package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,7 @@ experience better.
 
 ## Articles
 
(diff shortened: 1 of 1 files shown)


browniebroke (approved):

browniebroke:
Nice post!

PR #4534, opened 2023-08-19
Override _after_postgeneration to force save in User factory

thorrak:
## Description

Saving in `_after_postgeneration` was automatic in Factory Boy versions below 4.0, but will need to be done explicitly in v4.0+. Until then, _not_ overriding  `_after_postgeneration` (or explicitly setting `skip_postgeneration_save`) results in a warning when using `UserFactory` in tests. 

Checklist:

## Rationale

Eliminates a warning for recent versions of FactoryBoy and ensures that behavior will not change in future major releases.


browniebroke (approved):
Nice one

PR #4541, opened 2023-08-26
add compatibility options for black and flake8

### LINKED ISSUE #4538: Addressing E203 Whitespace Error in Pre-commit using flake8 Configuration
## Description

I propose adding the `extend-ignore = E203` option to the flake8 configuration in the project's pre-commit configuration file. This would address the issue of the `E203 "whitespace before ':'"` error that appears in the code snippet below:

`foo[index + 1 :]`

## Rationale

This issue seems to have been a long-standing topic of discussion, as seen in this GitHub issue: https://github.com/PyCQA/pycodestyle/issues/373. It appears to be a recurring problem for many developers.

Furthermore, the recommendation to add the `extend-ignore = E203` option comes from Black's official documentation in the context of ensuring compatibility with other tools like flake8: https://github.com/psf/black/blob/main/docs/guides/using_black_with_other_tools.md#flake8.

Is it acceptable if I submit a pull request to address this issue?

SungMinWoo:
## Description
Propose adding the E203 to the extend-ignore option of Flake8.

Checklist:

## Rationale
The official documentation of Black states that Flake8's E203 causes incompatibility with Black.

[1] https://github.com/psf/black/blob/main/docs/guides/using_black_with_other_tools.md#flake8
[2] https://github.com/PyCQA/pycodestyle/issues/373

Fix #4538

### DIFF
--- a/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
+++ b/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
@@ -50,6 +50,7 @@ repos:
     rev: 6.1.0
     hooks:
       - id: flake8
+        args: [ '--config=setup.cfg' ]
 
   - repo: https://github.com/Riverside-Healthcare/djLint
     rev: v1.32.1
--- a/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
+++ b/{{cookiecutter.project_slug}}/.pre-commit-config.yaml
@@ -50,7 +50,6 @@ repos:
     rev: 6.1.0
     hooks:
       - id: flake8
-        args: [ '--config=setup.cfg' ]
 
   - repo: https://github.com/Riverside-Healthcare/djLint
     rev: v1.32.1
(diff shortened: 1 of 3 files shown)


{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -50,6 +50,7 @@ repos:
     rev: 6.1.0
     hooks:
       - id: flake8
+        args: [ '--config=setup.cfg' ]
```
browniebroke, line 53: Do we need that? I was under the impression that [it's a location](https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations) where flake8 looks out of the box

SungMinWoo: Thank you for leaving a comment. I appreciate it. I have confirmed that this is the location that flake8 checks by default.

browniebroke (approved):

PR #4603, opened 2023-09-27
Fix ownership for /start-flower script in production Dockerfile

### LINKED ISSUE #4602: Flower on production raises Permission denied
## What happened?

Flower **doesn't** start on production using Dockerfile under `django` user

Flower **should** start on production using Dockerfile under `django` user

## Additional details

```bash
$ docker-compose -f production.yml logs flower
<project_name> | PostgreSQL is available
<project_name> | /bin/bash: /start-flower: Permission denied
```

I think rest of the report is irelevant. I'll provide a small PR to fix this.

SukiCZ:
## Description

Fixing a bug which seems like a missing a small typo. I needed to apply these changes to production env to be able to start Flower

Checklist:

The documentation doesn't need to be updated but I don't know how to test such change (it works on production tho)

## Rationale

Fix #4602

This project is optionally shipped with Flower which doesn't run in production without this change. This seems like a small typo as other files are changing ownership to Django user. I hope I'm not missing something.

Thanks for a great project BTW ❤️ 

EDIT: I could add some small test suite for Flower if requested (just need more time) 🤞🏿

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
@@ -103,7 +103,7 @@ RUN sed -i 's/\r$//g' /start-celerybeat
 RUN chmod +x /start-celerybeat
 
 
-COPY ./compose/production/django/celery/flower/start /start-flower
+COPY --chown=django:django ./compose/production/django/celery/flower/start /start-flower
 RUN sed -i 's/\r$//g' /start-flower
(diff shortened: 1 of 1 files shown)


browniebroke (approved):
Good catch. Wondering how it's been broken all this time. Oh well...

browniebroke:
Regarding testing, although I wish I had a better answer, there isn't much we can do right now, so don't worry about it.

Issue #4644, opened 2023-10-25
Python 3.12


foarsitter:
Abused the `create_django_issue.py` script to create a table for Python 3.12. 

There is a lot of work to do: 

## base.txt

| Name | Version in Master | 3.12 Compatible Version | OK|  Issue / PR |
| ---- | :---------------: | :-----------------------------: | :-: |:-:|
| [python-slugify](https://github.com/un33k/python-slugify) | 8.0.1 |  | ✅ |https://github.com/un33k/python-slugify/issues/138|
| [Pillow](https://python-pillow.org) | 10.1.0 | 10.1.0 | ✅ ||
| [rcssmin](http://opensource.perlig.de/rcssmin/) | 1.1.1 | 1.1.2 | ✅ |https://github.com/cookiecutter/cookiecutter-django/pull/4614|
| [argon2-cffi](https://pypi.org/project/argon2-cffi/) | 23.1.0 | 23.1.0 | ✅ ||
| [whitenoise](https://github.com/evansd/whitenoise) | 6.6.0 | 6.6.0 | ✅ ||
| [redis](https://github.com/redis/redis-py) | 5.0.3 |  | ❌ |https://github.com/redis/redis-py/pull/2979|
| [hiredis](https://github.com/redis/hiredis-py) | 2.3.2 | 2.3.1 | ✅ |https://github.com/redis/hiredis-py/issues/172|
| [celery](https://docs.celeryq.dev/) | 5.3.5 |  | ✅ |https://github.com/celery/celery/pull/8549|
| [django-celery-beat](https://github.com/celery/django-celery-beat) | 2.6.0 |  | ✅ |https://github.com/celery/django-celery-beat/issues/691|
| [flower](https://github.com/mher/flower) | 2.0.1 |  | ❌ |https://github.com/mher/flower/pull/1334|
| [uvicorn](https://pypi.org/project/uvicorn/) | 0.24.0.post1  | 0.24.0  | ✅ |https://github.com/encode/uvicorn/pull/2145|
| [django](https://www.djangoproject.com/) | 4.2.8 | 4.2.8  | ✅ |https://github.com/django/django/pull/17331 |
| [django-environ](https://django-environ.readthedocs.org) | 0.11.2 | >0.11.2  | ✅| https://github.com/joke2k/django-environ/pull/508|
[...]

PR #4692, opened 2023-11-22
Update developing-locally.rst

reyesvicente:
Update python version to 3.12

## Description
Updated python version to 3.12

Checklist:


browniebroke (changes requested):
See #4644

PR #4707, opened 2023-11-30
Move navbar to the right

reyesvicente:
move nav links to the right

## Description

Checklist:

## Rationale

Just moved the nav links to the right.


foarsitter:
@reyesvicente what is the reason behind this PR? Can you provide some more context?

browniebroke:
I too don't see value in merging this, this is a subjective cosmetic tweak, and folks would be 50-50 to get what we currently have vs what is suggested. So wihtout a clear reason, keep it as is.

reyesvicente:
> @reyesvicente what is the reason behind this PR? Can you provide some more context?

I don't understand why it's in the left in the first place. Normally, navlinks are at the right side of the page.

browniebroke:
No clear benefits, so priority to keep it as is. Thanks for the suggestion.

PR #4710, opened 2023-12-02
 Update mailpit to latest

GitBib:
## Description

Updated Mailpit Docker container from 1.8 to 'latest'.

## Checklist:

## Rationale
Enhances performance and security.


browniebroke (commented):
Since it's just used in local dev, and that it's apparently not automatically updated, should we specify its version with 'latest'? Or just pin the major version (if it's possible)...

GitBib:
@browniebroke It looks like a better use of 'latest'

browniebroke (approved):

PR #4726, opened 2023-12-09
Add missing __init__.py file to api module

Hraesvelg:
## Description

fix a missing init file as discuss here : https://github.com/cookiecutter/cookiecutter-django/discussions/4721

Checklist:

## Rationale

init file seems not necessary based on https://peps.python.org/pep-0420/ but the common usage imply to have this file


luzfcb (approved):
LGTM. Thanks @Hraesvelg

PR #4733, opened 2023-12-11
Add django-upgrade to pre-commit

foarsitter:
Browsing for some extra content I discovered that `django-upgrade` takes care of https://github.com/cookiecutter/cookiecutter-django/issues/4443. I was surprised we didn't include `django-upgrade` already but here it is.   

I used in the past for a legacy codebase and it is really helpful.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,6 +47,12 @@ repos:
     hooks:
       - id: flake8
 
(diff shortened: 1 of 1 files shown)


foarsitter:
Nevermind, I'm overlooking https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.pre-commit-config.yaml

PR #4790, opened 2024-01-16
Add a test to cover `DJANGO_ADMIN_FORCE_ALLAUTH`

### LINKED ISSUE #4789: `DJANGO_ADMIN_FORCE_ALLAUTH` is not covered by tests
## Description

The line in `{{ cookiecutter.project_slug }}.users.admin` after `DJANGO_ADMIN_FORCE_ALLAUTH` is not covered in tests, resulting in < 100% coverage in the rendered project. Namely, in this block:

```python
if settings.DJANGO_ADMIN_FORCE_ALLAUTH:
    # Force the `admin` sign in process to go through the `django-allauth` workflow:
    # https://docs.allauth.org/en/latest/common/admin.html#admin
    admin.site.login = decorators.login_required(admin.site.login)  # type: ignore[method-assign]
```

the line reassigning the value of `admin.site.login` is never tested.

## Rationale

Because the readme says this template renders a project with 100% coverage.

mfosterw:
## Description

Checklist:

## Rationale

Fixes #4789 and restores 100% test coverage.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
@@ -1,3 +1,8 @@
+from importlib import reload
+
+import pytest
+from django.contrib import admin
+from django.contrib.auth.models import AnonymousUser
 from django.urls import reverse
 
 from {{ cookiecutter.project_slug }}.users.models import User
@@ -47,3 +52,23 @@ class TestUserAdmin:
         url = reverse("admin:users_user_change", kwargs={"object_id": user.pk})
         response = admin_client.get(url)
         assert response.status_code == 200
+
+    @pytest.fixture
+    def force_allauth(self, settings):
+        settings.DJANGO_ADMIN_FORCE_ALLAUTH = True
+        # Reload the admin module to apply the setting change
+        import {{ cookiecutter.project_slug }}.users.admin as users_admin  # pylint: disable=import-outside-toplevel
+
+        try:
+            reload(users_admin)
+        except admin.sites.AlreadyRegistered:
+            pass
+
+    @pytest.mark.django_db
+    def test_allauth_login(self, rf, settings, force_allauth):  # pylint: disable=unused-argument
+        request = rf.get("/fake-url")
+        request.user = AnonymousUser()
+        response = admin.site.login(request)
+
+        # The `admin` login view should redirect to the `allauth` login view
+        assert response.status_code == 302
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
@@ -65,7 +65,8 @@ class TestUserAdmin:
             pass
 
     @pytest.mark.django_db
-    def test_allauth_login(self, rf, settings, force_allauth):  # pylint: disable=unused-argument
+    @pytest.mark.usefixtures("force_allauth")
+    def test_allauth_login(self, rf):
         request = rf.get("/fake-url")
         request.user = AnonymousUser()
         response = admin.site.login(request)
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
```diff
@@ -47,3 +52,23 @@ def test_view_user(self, admin_client):
...
+        # Reload the admin module to apply the setting change
+        import {{ cookiecutter.project_slug }}.users.admin as users_admin  # pylint: disable=import-outside-toplevel
+
+        try:
+            reload(users_admin)
+        except admin.sites.AlreadyRegistered:
+            pass
+
+    @pytest.mark.django_db
+    def test_allauth_login(self, rf, settings, force_allauth):  # pylint: disable=unused-argument
```
browniebroke, line 68: Why do we need to pass the `settings` fixture? It's not directly used by this method, is it?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
```diff
@@ -47,3 +52,23 @@ def test_view_user(self, admin_client):
...
+        request = rf.get("/fake-url")
+        request.user = AnonymousUser()
+        response = admin.site.login(request)
+
+        # The `admin` login view should redirect to the `allauth` login view
+        assert response.status_code == 302
```
browniebroke, line 74: Should we assert the `response.url` goes where we expect it to?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
browniebroke: Could probably also replace the `force_allauth` argument by using the decorator `@pytest.mark.use_fixtures("force_allauth")`, which would avoid the pylint issue.

mfosterw: Thanks, the settings were a mistake and I didn't know about that marker.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
mfosterw: Sure, I could add something like
```python
assert isinstance(response, HttpResponseRedirect) # for mypy
assert response.url == reverse(settings.LOGIN_URL) + "?next=" + request.path
```
(In which case I should add back the settings fixture). The `next` parameter has to be appended manually since it's not actually part of the defined url pattern.

mfosterw:
Also, I wanted to note that in the project I'm working on, I implemented this test using the `TestClient` with `follow=True` rather than `RequestFactory`, but when working on a blank rendering of the template I couldn't get this approach to work, since the `admin:login` view refused to redirect when I was testing. I don't know why it works in my project but not a new template. I don't think it makes a huge difference, but if you have any idea why it would behave differently in different projects I'm all ears.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
mfosterw: I implemented it with `assertRedirects` instead

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/test_admin.py
```diff
@@ -47,3 +53,25 @@ def test_view_user(self, admin_client):
...
+    @pytest.mark.django_db
+    @pytest.mark.usefixtures("force_allauth")
+    def test_allauth_login(self, rf, settings):
+        request = rf.get("/fake-url")
+        request.user = AnonymousUser()
+        response = admin.site.login(request)
+
+        # The `admin` login view should redirect to the `allauth` login view
+        target_url = reverse(settings.LOGIN_URL) + "?next=" + request.path
+        assertRedirects(response, target_url, fetch_redirect_response=False)
```
mfosterw, line 77: I had to use `fetch_redirect_response=False` since the response would need a client attribute otherwise. I can add `response.client = Client()` if fetching the redirect is important.

browniebroke:
Thanks 🎉

PR #4477, opened 2023-07-21
Migrate to the unified `STORAGES` setting added in Django 4.2

browniebroke:
## Description

There was a PR open for it in #4457 but the author closed it before it could be merged. Re-opening it with commits attributions.

Checklist:

- [x] I've made sure that tests are updated accordingly (especially if adding or updating a template option)
- [x] I've updated the documentation or confirm that my change doesn't require any updates

## Rationale

Fix #4443

PR #4803, opened 2024-01-24
Replace custom static & media storage classes by passing options in the `STORAGES` setting

jkaeske:
The custom cloud storages module has been deleted, and the settings have been updated to use the storage backend settings directly from each cloud provider's storage backend libraries. These changes have simplified the cloud storage configuration for each cloud provider in the production settings file.

This follows the recommendations from the `django-storages` docs and has been discussed in PR #4477

## Description

Checklist:

## Rationale

This will make maintenance easier and reduce complexity since the `utils` folder is not needed anymore and if no storage provider is selected the `storages.py` file does not need to be deleted.
It is inline with the recommendation from `django-storages`.


browniebroke (approved):
Nice one, looks good!

browniebroke:
Thanks! 🎉

jkaeske:
Thanks for merging this!

PR #4804, opened 2024-01-25
Add registry to Docker images names

timfreund:
## Description

Prepending `docker.io/` to image names in local.yml, production.yml and Dockerfiles. 

Checklist:

`tox -e py` was run successfully against Python 3.11.7

## Rationale

This change allows projects to run in alternate container runtimes such as podman-compose without additional registry configuration. The docker client is also able to pull the images in their fully qualified form.


browniebroke (approved):
> This change allows projects to run in alternate container runtimes such as podman-compose without additional registry configuration.

I didn't know about podman, thanks. This looks good to me, have no objections against that.

Issue #4810, opened 2024-01-27
Hybrid development 


foarsitter:
Running a single inside a container gives me a little more overhead then I would like. Therefore I'm using python locally and leveraging docker for all the external services like `redis` and `postgres`. I'm calling it `hybrid`. 

I would like to adjust the template for adding `hybrid` as an option alongside local & docker. 

What is your opinion about this approach?

JamesParrott:
Make a fork and lets have a look at it.  

It would need launcher scripts instead of letting docker compose spin everything up.  I wouldn't use it, but there's still value for users new to django in making a template for the exact example from the django docs introductory app, but with postgres, nginx and traefik etc. working out of the box, using containers.  That would be a perfect opportnuity to trim out some of the hacks in cookiecutter-django too (like the non-standard directory structure, and the python script in a string inside a bash script, that waits for a postgres connection).

But for the core mission of cookiecutter-django, and promise it makes to help create:  "jump start production-ready Django projects quickly", containerising the django run time has the following advantages:
- Currently local/production.yml is a great reference for what's running in your cookiecutter-django stack.

- Containerising Django allows the env it's in to be easily reproduced for celery, or for the user's custom services that need to use Django.

- Containerising makes the stack more portable and cloud agnostic (unlike running locally, it doesn't necessarily need a server.  It could run on a container run time PAAS like fly.io)

- One extra security layer - if there's a vulnerability in Django future, attackers are straight into the host OS, not just the container under the non-root user `django`.  I would not assume a container is bullet proof, but it does raise the cost to an attacker.

The docker network is a little nicer to work than ports, and avoids having to double check the ports of services running at OS level, against the firewall and network table.

Plus others that don't immediately spring to mind.

foarsitter:
My proposal is to use a local interpreter for development instead of the containerized one. So the production environment wouldn't change. 

Your bullets are valid points and should left untouched. The only change I'm proposing is to use a system interpreter so tests are running faster :)

JamesParrott:
The tests will indeed run faster, whihc would be nice I agree.   Is there a second route to running the tests directly in the repo, or can they be reorganised?  

But the subset of tests that can run locally outside a container, will not definitively test any interaction with the containerised environment, that is still the one actually being deployed.  So it could lead to false confidence, or late discovery of bugs.

browniebroke:
[...]

Issue #4539, opened 2023-08-24
Why prefer python than bash for wait the service start


AdrianPardo99:
## Description

In this file

https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/django/entrypoint

Why don't use some tcp listener for docker like [wait-for](https://github.com/eficode/wait-for) or [wait-for-it.sh](https://github.com/vishnubob/wait-for-it)

## Rationale

I think executing the python live code its a perfect idea but if we can implement a native script (bash) for implement another services that need to start first in the app maybe can replicate in other projects or services, maybe its only fork the script or the bash function.

Or maybe its only wait for the start of the service in a single file for wait until the database service is already started and exit.

browniebroke:
I honestly don't know why it's been done like this, maybe the original pull request has some details. Have you tried to search in the git history? 

I personally don't really have an issue with the current solution, but simplifications are always welcome. I'd be happy to replace it with an off the shelves solution.

foarsitter:
Adding wait-for to the project also comes in handy as `HEALTHCHECK` as discussed in https://github.com/cookiecutter/cookiecutter-django/issues/4428

PR #4811, opened 2024-01-27
Use POSTGRES_* and REDIS_URL as fallback

foarsitter:
This PR extends the base configuration in two ways:

1) When `DATABASE_URL` is missing it is using `POSTGRES_*` variables for configuring the database. 
2) `CELERY_BROKER_URL` defaults to `REDIS_URL` 

When using #4810 the `POSTGRES_*` and `REDIS_URL` variables are present but I'm not using the `entrypoint` script. This modification gives the possibility to run without `entrypoint` and without further modifications by using the variables already present.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -45,16 +45,16 @@ LOCALE_PATHS = [str(BASE_DIR / "locale")]
 # ------------------------------------------------------------------------------
 # https://docs.djangoproject.com/en/dev/ref/settings/#databases
 try:
-    { % if cookiecutter.use_docker == "y" - %}
+    {%- if cookiecutter.use_docker == "y" %}
     DATABASES = {"default": env.db("DATABASE_URL")}
-    { % - else %}
+    {%- else %}
     DATABASES = {
         "default": env.db(
             "DATABASE_URL",
             default="postgres://{% if cookiecutter.windows == 'y' %}localhost{% endif %}/{{cookiecutter.project_slug}}",
         ),
     }
-    { % - endif %}
+    {%- endif %}
 except environ.ImproperlyConfigured:
     DATABASES = {
         "default": {
@@ -297,7 +297,11 @@ if USE_TZ:
     # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std:setting-timezone
     CELERY_TIMEZONE = TIME_ZONE
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std:setting-broker_url
-CELERY_BROKER_URL = env("CELERY_BROKER_URL")
+try:
+    CELERY_BROKER_URL = env("CELERY_BROKER_URL")
+except environ.ImproperlyConfigured:
+    CELERY_BROKER_URL = env("REDIS_URL")
+
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std:setting-result_backend
 CELERY_RESULT_BACKEND = CELERY_BROKER_URL
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#result-extended
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -44,16 +44,28 @@ LOCALE_PATHS = [str(BASE_DIR / "locale")]
 # DATABASES
 # ------------------------------------------------------------------------------
 # https://docs.djangoproject.com/en/dev/ref/settings/#databases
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -44,16 +44,28 @@
...
+    DATABASES = {
+        "default": {
+            "ENGINE": "django.db.backends.postgresql",
+            "NAME": env.str("POSTGRES_DB"),
+            "USER": env.str("POSTGRES_USER"),
+            "PASSWORD": env.str("POSTGRES_PASSWORD"),
+            "HOST": env.str("POSTGRES_HOST"),
+            "PORT": env.str("POSTGRES_PORT"),
+        }
+    }
```
browniebroke, line 68: I'm not a big fan of this `except` 🤔 I can't quite really put a finger on why, though (sorry, that's unhelpful)

{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -285,7 +297,11 @@
     # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std:setting-timezone
     CELERY_TIMEZONE = TIME_ZONE
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std:setting-broker_url
-CELERY_BROKER_URL = env("CELERY_BROKER_URL")
+try:
+    CELERY_BROKER_URL = env("CELERY_BROKER_URL")
+except environ.ImproperlyConfigured:
+    CELERY_BROKER_URL = env("REDIS_URL")
```
browniebroke, line 303: When I look at this, I wonder why we need `CELERY_BROKER_URL` in the first place 🤔

browniebroke (commented):
Looking at the issue about hybrid dev, I'd like to ask a few questions, in the hope that we can maybe achieve the same outcome, without the try/except:

1. Why is it easier to set 5 env variables (`POSTGRES_...`) than 1 (`DATABASE_URL`) in your hybrid dev mode (without Docker)? 
2. The `POSTGRES_...` vars are set in Docker, so do we actually need the `DATABASE_URL`?
3. Could we instead replace the entrypoint script by an off-the-shelf solution as suggested in https://github.com/cookiecutter/cookiecutter-django/issues/4539?

Then perhaps we could always use the `POSTGRES_...` variables instead?

{{cookiecutter.project_slug}}/config/settings/base.py
browniebroke: I think it's because in case of problems, it because harder to figure out where the DB credentials are coming from. Do I have a `DATABASE_URL` set? Do I have anything missing from the required `POSTGRES_...` env? It's only one more place to check, but it can be _really_ confusing when your app won't conect to the DB

{{cookiecutter.project_slug}}/config/settings/base.py
foarsitter: For a project, in the time that Heroku had free redis instances, I used a three instances: a cache backend, celery broker and result backend. But yeah, we can get rid of it since it doesn't add any value.

foarsitter:
> 1. Why is it easier to set 5 env variables (`POSTGRES_...`) than 1 (`DATABASE_URL`) in your hybrid dev mode (without Docker)?

When I'm talking about hybrid I' talking about a local interpreter with docker services.  In a non docker setup It wouldn't matter to me to set 5 or just 1 composed out of 5.  

> 2. The `POSTGRES_...` vars are set in Docker, so do we actually need the `DATABASE_URL`?

`DATABASE_URL` is there for deploys on heroku I suppose. 

> 3. Could we instead replace the entrypoint script by an off-the-shelf solution as suggested in [Why prefer python than bash for wait the service start #4539](https://github.com/cookiecutter/cookiecutter-django/issues/4539)?

If we get rid of `CELERY_BROKER_URL` then we can replace the entrypoint script. `DATABASE_URL` isn't required for the docker setup

> Then perhaps we could always use the `POSTGRES_...` variables instead?

We can put  `DATABASE_URL` around an statement that checks for `use_heroku`.  

Sweet, I like where we are going!

{{cookiecutter.project_slug}}/config/settings/base.py
foarsitter: Relaying on exceptions here is a lazy way of doing things I suppose. When a faulty url is given it would try to use the postgres vars, so I agree on your reluctance and it should be improved.

foarsitter:
Closing this in favor of  https://github.com/cookiecutter/cookiecutter-django/pull/4861 and https://github.com/cookiecutter/cookiecutter-django/pull/4862

PR #4836, opened 2024-02-07
Generic UserManager

foarsitter:
This PR fixes three mypy errors by implementing a generic `UserManager[User]` and ignoring the overrides:

```
users/managers.py:23: error: "_T" has no attribute "password"  [attr-defined]
users/managers.py:27: error: Signature of "create_user" incompatible with supertype "UserManager"  [override]
users/managers.py:32: error: Signature of "create_superuser" incompatible with supertype "UserManager"  [override]
```


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/managers.py
```diff
@@ -1,8 +1,13 @@
+from typing import TYPE_CHECKING
+
 from django.contrib.auth.hashers import make_password
 from django.contrib.auth.models import UserManager as DjangoUserManager
 
+if TYPE_CHECKING:
+    from ruff1.users.models import User  # noqa: F401
```
browniebroke, line 7: Where is that `ruff1` coming from? 🤔

browniebroke (approved):

PR #4838, opened 2024-02-07
Extend docker test with deploy check

### LINKED ISSUE #4462: Improve test coverage of production build/config
## Description

Add some automated check for the production build and configuration. While it would be tricky to do an actual deployment, it would be nice to at least try to install dependencies for prod, and run a simple Django command like [`manage.py check --deploy`](https://docs.djangoproject.com/en/4.2/howto/deployment/checklist/#run-manage-py-check-deploy).

## Rationale

Currently our CI isn't covering the production setup at all, and some issues won't be caught unless we do an actual deployment, which needs some manual work.

foarsitter:
As the [docs](https://docs.djangoproject.com/en/5.0/ref/django-admin/#cmdoption-check-deploy) already mention, `manage.py check --deploy` requires the production settings, otherwise it fails on security related settings. In turn, running the production settings require the production dependencies. 

This PR references the production requirements in `requirements/local.txt` so the dependencies like `collectfast` are available during development. This gives us the possibility to use the production settings in the local docker image and run `check --deploy`. For convenience I added the `--database` to check since they aren't run by default.

Fixes #4462


{{cookiecutter.project_slug}}/requirements/local.txt
```diff
@@ -1,4 +1,4 @@
--r base.txt
+-r production.txt
```
browniebroke, line 1: Next step: merge `base.txt` and `production.txt`? 😄

browniebroke (approved):
I was thinking of adding more job to the CI, but actually that's much better.

Do you think it would be worth changing `test_bare.sh` to run this deploy check as well? 

I haven't thought about this much, and I'm assuming that it's a similar level of complexity as `test_docker.sh`. If it's more complicated than that (need extra services to be runnning on CI), I wouldn't do it as part of this PR.

{{cookiecutter.project_slug}}/requirements/local.txt
foarsitter: I've suggested that before in #2810 but felt out of scope for this PR.

foarsitter:
> Do you think it would be worth changing `test_bare.sh` to run this deploy check as well?

I don't think it is worth it since the production settings do not contain any logic that differ between bare and docker.

{{cookiecutter.project_slug}}/requirements/local.txt
browniebroke: Oh yes, I agree: 100% out of scope. I was merely pointing out that it might help us get there without changing too much our requirements files.

browniebroke:
> > Do you think it would be worth changing `test_bare.sh` to run this deploy check as well?
> 
> I don't think it is worth it since the production settings do not contain any logic that differ between bare and docker.

fair enough. Anyway, LGTM! I will feel a tiny bit more confident in merging updates to our prod dependencies now. Thanks for that.

PR #4843, opened 2024-02-08
Allauth elements & MFA

### LINKED ISSUE #4691: Implement django-allauth's MFA
## Description

Can we add this?

## Rationale

For added security for the project

### LINKED ISSUE #4819: Allauth template elements
Looking into #4691 I stumbled upon the [elements](https://github.com/pennersr/django-allauth/tree/main/allauth/templates/allauth/elements) directory.  

Since 0.58.0 (2023-10-26) it is possible to override just the elements instead of overriding the templates completely: https://docs.allauth.org/en/latest/common/templates.html#styling-the-existing-templates

This feature lowers our effort to keep up-to-date with `allauth` but I'm not sure if we can achieve the same quality.

foarsitter:
This PR implements #4819 and #4691

The design's are very minimalistics but it will be much easier to keep up-to-date with allauth upstream. For example used texts are now the same as allauth uses so we can benefit from their translations.

The way I handle tags isn't the most pretty one but we can address that in another PR. 

Fix #4819
Fix #4691


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/account/verification_sent.html
browniebroke, line 1: It looks like this page is now unstyled
image

I don't understand where it's coming from 🤔

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/account/login.html
browniebroke, line 1: This page has no navbar, is that intended?

image

It [used to be there](https://browniebrokewnymaster-a0616fed8aca.herokuapp.com/accounts/login/), would be nice to keep it.

browniebroke (commented):
Deployed this branch to Heroku: https://browniebrokewny-eaa7f24dd6c4.herokuapp.com/

And the master branch, as baseline: https://browniebrokewnymaster-a0616fed8aca.herokuapp.com/

Looks like a few pages have lost some bits of styling...

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/account/verification_sent.html
foarsitter: The `m-auto` isn't adding any margin when there is a long text. Fixed it with a `col-md-4` so each component (login, register, verification sent text) has the same size. Centered the content with `justify-content-center`.

browniebroke (approved):

PR #4861, opened 2024-02-15
Remove CELERY_BROKER_URL in favor of REDIS_URL

foarsitter:
As discussed in https://github.com/cookiecutter/cookiecutter-django/pull/4811 this is how it looks like to remove `CELERY_BROKER_URL`. 

Considering the amount of deletions, this looks like the right path.

### DIFF
--- a/docs/2-local-development/developing-locally.rst
+++ b/docs/2-local-development/developing-locally.rst
@@ -55,8 +55,6 @@ First things first.
 #. Set the environment variables for your database(s): ::
 
     $ export DATABASE_URL=postgres://postgres:<password>@127.0.0.1:5432/<DB name given to createdb>
-    # Optional: set broker URL if using Celery
-    $ export CELERY_BROKER_URL=redis://localhost:6379/0
 
    .. note::
 
--- a/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
@@ -125,9 +125,6 @@ RUN chown -R django:django ${APP_HOME}
 USER django
 
 RUN DATABASE_URL="" \
-  {%- if cookiecutter.use_celery == "y" %}
-  CELERY_BROKER_URL="" \
-  {%- endif %}
   DJANGO_SETTINGS_MODULE="config.settings.test" \
   python manage.py compilemessages
 
--- a/{{cookiecutter.project_slug}}/compose/production/django/entrypoint
+++ b/{{cookiecutter.project_slug}}/compose/production/django/entrypoint
@@ -4,12 +4,6 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-
-{% if cookiecutter.use_celery == 'y' %}
-# N.B. If only .env files supported variable expansion...
-export CELERY_BROKER_URL="${REDIS_URL}"
-{% endif %}
-
 if [ -z "${POSTGRES_USER}" ]; then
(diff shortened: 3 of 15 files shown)


browniebroke (commented):
Looks good at first glance - could you please merge/rebase the latest master to do some final chackes?

docs/2-local-development/developing-locally.rst
```diff
@@ -55,8 +55,6 @@ First things first.
 #. Set the environment variables for your database(s): ::
 
     $ export DATABASE_URL=postgres://postgres:<password>@127.0.0.1:5432/<DB name given to createdb>
-    # Optional: set broker URL if using Celery
-    $ export CELERY_BROKER_URL=redis://localhost:6379/0
```
browniebroke, line 59: Should this tell people to set the `REDIS_URL`?

{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
```diff
@@ -125,9 +125,6 @@ RUN chown -R django:django ${APP_HOME}
 USER django
 
 RUN DATABASE_URL="" \
-  {%- if cookiecutter.use_celery == "y" %}
-  CELERY_BROKER_URL="" \
-  {%- endif %}
```
browniebroke, line 130: I think that was required because the env variable has no default in the settings, which is still the case with `REDIS_URL`

{{cookiecutter.project_slug}}/compose/production/django/entrypoint
```diff
@@ -4,12 +4,6 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-
-{% if cookiecutter.use_celery == 'y' %}
-# N.B. If only .env files supported variable expansion...
-export CELERY_BROKER_URL="${REDIS_URL}"
-{% endif %}
```
browniebroke, line 11: :tada:

{{cookiecutter.project_slug}}/docs/conf.py
```diff
@@ -26,9 +26,6 @@
     sys.path.insert(0, os.path.abspath(".."))
 {%- endif %}
 os.environ["DATABASE_URL"] = "sqlite:///readthedocs.db"
-{%- if cookiecutter.use_celery == 'y' %}
-os.environ["CELERY_BROKER_URL"] = os.getenv("REDIS_URL", "redis://redis:6379")
-{%- endif %}
```
browniebroke, line 31: This one might be still be required too...

{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
foarsitter: What do you think of this default: 
```python
env("REDIS_URL", default="redis://{% if cookiecutter.use_docker == 'n' %}localhost{%else%}redis{% endif %}:6379/0")
```

browniebroke: Yes sounds good. Might read better with a "truthy" condition `if use_docker == y`? But yea, would be simpler with a default value in settings

{{cookiecutter.project_slug}}/compose/production/django/entrypoint
foarsitter: Adding a default value resolves this one to

browniebroke (approved):

foarsitter:
Thanks for your valuable feedback @browniebroke!

PR #4863, opened 2024-02-15
Speed up CI pytest by 30%: Change ci.yml to only build images relevant for running pytest

### LINKED ISSUE #4865: ci.yml for running pytest is building docs image unnecessarily (run time 4:30m instead of 3:00m)
## What happened?

With docker setup enabled ci.yml is building all images (including the docs) during the Github Actions workflow. But only the images related to the django service are needed to run pytest in the next step. This causes the CI workflow to last 4m 30s instead of just below 3m (incl. linter).

Only the images related to running the django service and therefore the pytest command should be build.
Please see my pull request for the minor required change: #4863

mtmpaulwulff:
In order to run pytest we can constrain the build command to only build the images related to the django service. 

This mainly prevents the docs image from being built unnecessarily.

This cuts run time of the Github Actions workflow **from 4min 30s to below 3min**.

Fix #4865

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.0
 
-  # With no caching at all the entire ci process takes 4m 30s to complete!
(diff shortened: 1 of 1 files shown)


foarsitter (approved):
Smart!

browniebroke (approved):

PR #4879, opened 2024-02-20
You probably don't need `get_user_model`

### LINKED ISSUE #4872: You probably don't need `get_user_model`
## Description

Import `User` directly, rather than using `get_user_model`.

## Rationale

`get_user_model` is meant for *reusable* apps, while it is my understanding this project is targeted more towards creating websites than packages. Especially within the `users` app it doesn't make any sense to use it (are we expecting users to create a new user model in a custom app but then keep the `users` app in their project?) , and switching can prevent new users from assuming they have to use it in all their custom apps (which may or may not be what happened to me). For a more in-depth explanation of why it's an anti-pattern, read this [blog post](https://adamj.eu/tech/2022/03/27/you-probably-dont-need-djangos-get-user-model/).

mfosterw:
## Description

Checklist:

## Rationale

Fixes #4872


{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/api/views.py
```diff
@@ -1,13 +1,12 @@
...
 from rest_framework.decorators import action
 from rest_framework.mixins import ListModelMixin, RetrieveModelMixin, UpdateModelMixin
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
-from .serializers import UserSerializer
+from {{ cookiecutter.project_slug }}.users.models import User
 
-User = get_user_model()
+from .serializers import UserSerializer
```
mfosterw, line 9: You use a local import here and nowhere else; might I suggest the entire `User` app be changed to use local imports? I personally think it's cleaner, and it can encourage new developers to encapsulate their apps and make them more reusable (which I know is the opposite of what this pull request is doing, but still)

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tasks.py
```diff
@@ -1,8 +1,5 @@
-from django.contrib.auth import get_user_model
-
 from config import celery_app
-
-User = get_user_model()
+from {{ cookiecutter.project_slug }}.users.models import User
 
 
 @celery_app.task()
```
mfosterw, line 5: Per [celery docs](https://docs.celeryq.dev/en/stable/userguide/tasks.html#how-do-i-import-the-task-decorator), 

> If you’re using Django... then you probably want to use the shared_task() decorator

mfosterw: I can open a new pr and write a fix if you have no objections

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/factories.py
```diff
@@ -1,11 +1,12 @@
 from collections.abc import Sequence
 from typing import Any
 
-from django.contrib.auth import get_user_model
 from factory import Faker
 from factory import post_generation
```
mfosterw, line 5: Ruff split this import into 2 lines, was this expected?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tasks.py
browniebroke: I had the vague impression that it was for reusable apps, but that was from a long time ago. I'm happy to follow the latest recommendations on this, yes. Agree that I'd rather be in a separate PR

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tests/factories.py
browniebroke: We just switched to Ruff and we didn't really discuss that. Feel free to open an issue if you want to discuss it further, I don't have any strong opinion, only a slight habit of grouping the imports, but that's mostly following what isort does by default, I think. I guess the argument for putting them on separate line is that it makes diff clearer?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/admin.py
```diff
@@ -1,19 +1,18 @@
...
 from django.contrib import admin
 from django.contrib.auth import admin as auth_admin
-from django.contrib.auth import decorators
-from django.contrib.auth import get_user_model
+from django.contrib.auth.decorators import login_required
 from django.utils.translation import gettext_lazy as _
 
 from {{ cookiecutter.project_slug }}.users.forms import UserAdminChangeForm
 from {{ cookiecutter.project_slug }}.users.forms import UserAdminCreationForm
+from {{ cookiecutter.project_slug }}.users.models import User
```
mfosterw, line 9: Ruff doesn't like my import order for some reason and I can't figure out why

mfosterw: TIL ruff wants two lines before class definitions but one line before statements

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/api/views.py
mfosterw: Ok, I'll open a PR as soon as this one is merged

browniebroke (approved):

browniebroke:
Thanks a lot, much simpler 🎉

PR #4881, opened 2024-02-20
Switch to celery.shared_task

mfosterw:
Per [celery docs](https://docs.celeryq.dev/en/stable/userguide/tasks.html#how-do-i-import-the-task-decorator), 

> If you’re using Django... then you probably want to use the shared_task() decorator

Checklist:

## Rationale

As mentioned in https://github.com/cookiecutter/cookiecutter-django/pull/4879#discussion_r1496379763


browniebroke (approved):
Looks good to me. Would you mind resolving the merge conflict, please?

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/tasks.py
```diff
@@ -1,8 +1,8 @@
-from config import celery_app
+from celery import shared_task
```
browniebroke, line 1:
```suggestion
from celery import shared_task

```

Issue #4880, opened 2024-02-20
Allow multiple imports on one line


mfosterw:
## Description

Remove the following from `pyproject.toml`:

```toml
[tool.ruff.lint.isort]
force-single-line = true
```

## Rationale

I've never heard of forcing every import to be on its own line, and I am concerned that doing so will lead import blocks to be unnecessarily long. I haven't heard of many arguments in favor of this style, but I am open to them if anyone disagrees.

browniebroke:
I personally don't mind too much. The main argument for forcing to be on a single line is https://github.com/asottile/reorder-python-imports#why-this-style:

> The style chosen by `reorder-python-imports` has a single aim: reduce merge conflicts.
> 
> By having a single import per line, multiple contributors can add / remove imports from a single module without resulting in a conflict.
> 
> Consider the following example which causes a merge conflict:
> 
> ```diff
> # developer 1
> -from typing import Dict, List
> +from typing import Any, Dict, List
> ```
> 
> ```diff
> # developer 2
> -from typing import Dict, List
> +from typing import Dict, List, Tuple
> ```
> 
> no conflict with the style enforced by `reorder-python-imports`:
> 
> ```diff
> +from typing import Any
>  from typing import Dict
>  from typing import List
> +from typing import Tuple
> ```

As I said, I don't feel strongly about it.

JamesParrott:
> I've never heard of forcing every import to be on its own line

> I haven't heard of many arguments in favor of this style,

FYI:

https://peps.python.org/pep-0008/#imports

> Imports should usually be on separate lines:
> 
> # Correct:
> import os
> import sys
> 
> # Wrong:
> import sys, os

mfosterw:
> Yes: from collections.abc import Mapping, Sequence
> import os
> import sys
> from typing import Any, NewType
> 
> No: import os, sys

This rule is not about splitting imports from different modules, but putting imports from the *same* module on different lines, i.e.

> from typing import Any, NewType

-->

> from typing import Any
> from typing import NewType

It is this kind of splitting that I had never heard of. I thought that this was meant to just reduce diff size, but the point about merge conflicts does actually resonate with me somewhat, but I'm curious what other people think.

luzfcb:
For imports of multiple items in a single _**import**_ clause, I prefer

```python
from myapp import (
    a_module,
    b_module,
    c_function
)
```
instead of 

```python
from myapp import a_module, b_module, c_function
```

[...]

PR #4883, opened 2024-02-21
Switch to local imports within app

mfosterw:
## Description

Switch to using local imports within app.

Checklist:

## Rationale

Encourage encapsulation and reusability by importing modules within an app using local imports, instead of absolute imports, as discussed in https://github.com/cookiecutter/cookiecutter-django/pull/4879#discussion_r1496455646

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/urls.py
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/urls.py
@@ -1,8 +1,8 @@
 from django.urls import path
 
-from {{ cookiecutter.project_slug }}.users.views import user_detail_view
-from {{ cookiecutter.project_slug }}.users.views import user_redirect_view
-from {{ cookiecutter.project_slug }}.users.views import user_update_view
+from .views import user_detail_view
+from .views import user_redirect_view
+from .views import user_update_view
 
 app_name = "users"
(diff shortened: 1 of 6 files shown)


mfosterw:
I tried also changing the imports in `users/api` to be relative to `users` (with the understanding it was dubious to begin with) but ruff complained so I reverted it.

{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/users/urls.py
```diff
@@ -1,8 +1,8 @@
 from django.urls import path
 
-from {{ cookiecutter.project_slug }}.users.views import user_detail_view
-from {{ cookiecutter.project_slug }}.users.views import user_redirect_view
-from {{ cookiecutter.project_slug }}.users.views import user_update_view
+from .views import user_detail_view
+from .views import user_redirect_view
+from .views import user_update_view
```
mfosterw, line 5: If we do keep the `force-single-line` rule for imports, we might want to change this style of importing views, and instead do:
```python
from . import views

urlpatterns = [
    path("~redirect/", view=views.user_redirect_view, name="redirect"),
    ...
```

This is one case where I think the issue of import blocks being too long could be an actual problem #4880

browniebroke (approved):
Looks good

Issue #4478, opened 2023-07-21
Upload media permission denied in production


Tshibek:
Hello, with new Django version 4.2.3, i need add in STORAGES default
`STORAGES = {
    "default": {
        "BACKEND": "django.core.files.storage.FileSystemStorage",
    },
    "staticfiles": {
        "BACKEND": "whitenoise.storage.CompressedStaticFilesStorage",
    },
}`
and after it i added, i tried upload files in production and get ERROR500, in console is
log like that `PermissionError: [Errno 13] Permission denied: '/app/{project_name}/media/file'`

browniebroke:
**We have a set of issue templates, please help us help you: follow them to report a bug, they have questions we _really_ need answers from, otherwise it's really hard to answer.**

Might be related to https://github.com/cookiecutter/cookiecutter-django/issues/4209 which OP mentioned a permission issue:

> in order to create my own media folder with django:django as the owner:group, because in the PR doesn't include the media folder creation and when I build containers this media folder was created with root:root ownership, so my django app crashed when I tried to upload any file because doesn't have permissions to write on that folder.

Tshibek:
From the above just did 
> The only one modification I made was in /compose/production/django/Dockerfile, I added this line: RUN mkdir ${APP_HOME}/proveemax/media && chown django:django ${APP_HOME}/proveemax/media
> I think @edwardigarashi might be onto something. The rule for `web-media-router` needs to have priority over the rule for `web-secure-router`. Traefik looks at the rule length to determine this when no explicit priority is set.
> 
> These rules are defined in [this file (traefik.yml)](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/traefik/traefik.yml).
> 
> Adding a high priority like 1000 (just has to be higher than the length of the rule in `web-secure-router`) will make Traefik prioritize this rule.
> 
> ```
>     web-media-router:
>       rule: "Host(`{{ cookiecutter.domain_name }}`) && PathPrefix(`/media/`)"
>       entryPoints:
>         - web-secure
>       middlewares:
>         - csrf
>       service: django-media
>       tls:
>         certResolver: letsencrypt
>       priority: 1000
> ```
> 
> Alternatively, you can manually set the priority for both rules. Which might be a better solution to merge in case this solves the issue. Or we could even add the `www` subdomain to the media router rule like this:
> 
> ```
[...]

PR #4889, opened 2024-02-24
Fix permissions for media files when served by nginx

s-emanuilov:
## Description

Minor changes: adding a recursive flag in the Django Docker file is needed for proper permissions.

Checklist:

## Rationale

Without this flag in the Docker file, local media files cannot be saved (with no cloud providers) as there are no permissions. Having all folders and files under django:django -> solves those (and similar) problems.

Ref #4478 
Ref #4209


browniebroke (approved):
Nice catch, has been reported a couple of times I think

PR #4894, opened 2024-02-27
Fix invalid HTML in django-allauth field element template

reggieriser:
## Description

Checklist:

## Rationale

I found the HTML validation errors below in the new `django-allauth` overridden field element template. Without fixing these, you might see some side-effects depending on how your browser handles the invalid HTML.
- hard-coded `for` attribute value for a radio button (found on the e-mail management page)
- missing `</div>` for help text which was caused by an out of place `{% endslot %}` (found on the MFA activation page)


foarsitter (approved):
This indeed looks like faulty behavior you are fixing here. Thanks @reggieriser!

PR #4905, opened 2024-03-06
Fix fully qualified docker images

### LINKED ISSUE #4903: Incorrect Python version in Docker.
## What happened?
Aliases for Docker image builds aren't functioning as intended. Currently, Python version 3.11 is specified, but the image is built from 3.12 because aliases aren't used in stages 2 and 3, and the reference to Python is made without specifying the version. As a result, besides having the incorrect Python version, each image also occupies about ~1.5 gigabytes instead of ~800 megabytes.

The image should be built based on the base image with the specified version.

GitBib:
## Description

Fix Python Docker image version not being recognized after adding registry to image names

Checklist:

## Rationale

Fix #4903


browniebroke (approved):
Looks good 👍🏻

PR #4950, opened 2024-03-29
Change pytest import mode to importlib

GvS666:
## Description

Change pytest import mode to importlib

Checklist:

## Rationale

Based on https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#choosing-an-import-mode :

"For historical reasons, pytest defaults to the prepend import mode instead of the importlib import mode we recommend for new projects. The reason lies in the way the prepend mode works:

Since there are no packages to derive a full package name from, pytest will import your test files as top-level modules. The test files in the first example (src layout) would be imported as test_app and test_view top-level modules by adding tests/ to sys.path.

This results in a drawback compared to the import mode importlib: your test files must have unique names. (...)
The importlib import mode does not have any of the drawbacks above, because sys.path is not changed when importing test modules.
"

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -1,7 +1,7 @@
 # ==== pytest ====
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--ds=config.settings.test --reuse-db"
+addopts = "--ds=config.settings.test --reuse-db --import-mode=importlib"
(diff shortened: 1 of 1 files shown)


foarsitter (approved):
Thanks @GvS666

traviswaelbro:
It's worth noting that this currently results in a `pytest` output issue, where `assert` failures are not having their differences output as expected. For reference, see pytest-dev/pytest#12659.

PR #4954, opened 2024-04-01
Use main over master for branch name in deployment-on-heroku instruction

PatDuJour:
## Description

Checklist:

## Rationale

The newer convention(although it has been years) is to use `main` as the default branch over `master`. Newly created cookiecutter projects are more likely to have the default branch as `main` rather than `master`


foarsitter (approved):
Thanks @PatDuJour

PR #4960, opened 2024-04-02
Remove double RUN cmd in ci.yml

DavidPacioianu:
Remove double RUN cmd

## Description

Remove `run: docker compose -f local.yml build docs`

## Rationale

Not fixing this throws the following error during pre-commit: `found duplicate key "run" with value "docker compose -f local.yml build docs" (original value: "docker compose -f local.yml build django")`


browniebroke:
Fixed in #4972 - the intent was to keep the line building the docs, but it should have been on a separate step. Thanks for the help

PR #4971, opened 2024-04-06
Add options to reload django when file changes in windows docker

### LINKED ISSUE #4970: Why doesn't add RunServerPlus options for reload in windows when using docker?
## Description

Add this code when window = y when generating project, for reloading django code in window.

```
end of base.py
# RunServerPlus
{% if cookiecutter.windows == 'y' %}
# After how many seconds auto-reload should scan for updates in poller-mode
RUNSERVERPLUS_POLLER_RELOADER_INTERVAL = 5

# Werkzeug reloader type [auto, watchdog, or stat]
RUNSERVERPLUS_POLLER_RELOADER_TYPE = 'stat'
{% endif %}
```

## Rationale

In window with docker, somehow(might be filesytem problem) reload doens't work with default RELOADER_TYPE.
So we should add above code but when the first time using cookiecutter, all of users don't realize what happens like me.
So it should be set by default when windows option as y in generating project step.

quroom:
## Description
Supporting django auto reload in windows docker.

## Rationale
Now it doesn't reload when django code is changed in windows docker env.
The problem is related to wsl2 and Werkzeug's built-in reloading system.
So it can be fixed with stat polling options.

FIX #4970

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -83,15 +83,15 @@ if env("USE_DOCKER") == "yes":
         # The node container isn't started (yet?)
         pass
     {%- endif %}
-{% if cookiecutter.windows == 'y' %}
-# RunServerPlus
-# ------------------------------------------------------------------------------
-# After how many seconds auto-reload should scan for updates in poller-mode
-RUNSERVERPLUS_POLLER_RELOADER_INTERVAL = 5
-
-# Werkzeug reloader type [auto, watchdog, or stat]
-RUNSERVERPLUS_POLLER_RELOADER_TYPE = 'stat'
-{% endif %}
+    {%- if cookiecutter.windows == 'y' %}
+    # RunServerPlus
+    # ------------------------------------------------------------------------------
+    # This is a custom setting for RunServerPlus to fix reloader issue in Windows docker environment
+    # Werkzeug reloader type [auto, watchdog, or stat]
+    RUNSERVERPLUS_POLLER_RELOADER_TYPE = 'stat'
+    # If you have CPU and IO load issues, you can increase this poller interval e.g) 5
+    RUNSERVERPLUS_POLLER_RELOADER_INTERVAL = 1
+    {%- endif %}
 {%- endif %}
 
 # django-extensions
--- a/{{cookiecutter.project_slug}}/config/settings/local.py
+++ b/{{cookiecutter.project_slug}}/config/settings/local.py
@@ -83,6 +83,15 @@ if env("USE_DOCKER") == "yes":
         # The node container isn't started (yet?)
         pass
     {%- endif %}
+{% if cookiecutter.windows == 'y' %}
+# RunServerPlus
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -83,6 +83,15 @@
         # The node container isn't started (yet?)
         pass
     {%- endif %}
+{% if cookiecutter.windows == 'y' %}
```
browniebroke, line 86: I'm not sure it will change much in the final formatting, but I tend to add a minus sign to [trim empty lines before the Jinja tags](https://jinja.palletsprojects.com/en/2.10.x/templates/#whitespace-control), and add an explicit emtpy line after:

```suggestion
{%- if cookiecutter.windows == 'y' %}

```

I found it easier to see where the empty line will be when the codition is true or false.

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -83,6 +83,15 @@
...
+# RunServerPlus
+# ------------------------------------------------------------------------------
+# After how many seconds auto-reload should scan for updates in poller-mode
+RUNSERVERPLUS_POLLER_RELOADER_INTERVAL = 5
```
browniebroke, line 90: 5 seconds seems a bit long, no? The default value of 1 second seems more appropriate:

```suggestion
RUNSERVERPLUS_POLLER_RELOADER_INTERVAL = 1
```

Actually, since it's the default interval, do we need to define it? I think we could omit it...

{{cookiecutter.project_slug}}/config/settings/local.py
```diff
@@ -83,6 +83,15 @@
...
+
+# Werkzeug reloader type [auto, watchdog, or stat]
+RUNSERVERPLUS_POLLER_RELOADER_TYPE = 'stat'
+{% endif %}
```
browniebroke, line 94: To avoid leaving an extra empty line:

```suggestion
{%- endif %}
```

browniebroke (commented):
Just left a few comments. Would be interested if you went with 5 seconds for a specific reason?

quroom:
> Just left a few comments. Would be interested if you went with 5 seconds for a specific reason?

In [offical document](https://django-extensions.readthedocs.io/en/stable/runserver_plus.html#io-calls-and-cpu-usage), Default setting can make pretty brute force and continously issues stat system calls which causes the CPU and IO load.
So I just test default setting and 5 seconds too.
In my env, nothing much diffrent like reload performance and others too.
That's why I set it like that , But as you said It could be omit also, It can be set 1 second and leave some comment like this
```
    # If you have CPU and IO load issues, you can increase this poller interval e.g) 5
    RUNSERVERPLUS_POLLER_RELOADER_INTERVAL = 1
```

{{cookiecutter.project_slug}}/config/settings/local.py
quroom: Yeah, it looks better. and also it should be add indent too.

quroom:
@browniebroke I made a new commit to relect your comments. Could you review that?

browniebroke (approved):

PR #4973, opened 2024-04-06
Feature/add missing migrations file check

arnav13081994:
## Description

Added a local pre-commit hook that fails in case the user forgot to commit a migration file.

Checklist:


browniebroke (commented):
I'm -1 on this. I don't think pre-commit is the right place to check for that.

Also, from the template maintenance perspective, this file if free of templating logic, which enable us to run `pre-commit autoupdate` to keep versions up to date. Accepting this PR would break this workflow. I think the trade-off is not worth it.

PR #4976, opened 2024-04-08
Fix Django Debug Toolbar Visibility and INTERNAL_IPS in Docker Environments

DavidPacioianu:
Problem: The Django Debug Toolbar does not display in Docker (tested on MacOS) because `socket.gethostname()` returns Docker's default IP (172.21.0.1), not the host's IP.

Solution: Switch `socket.gethostname()` to `host.docker.internal` to accurately reference the host machine's IP from within Docker containers. This change ensures the the host's IP is added to INTERNAL_IPS and the Django debug toolbar is visible when developing in Docker.


foarsitter:
This might be working on MacOS, but doesn't work on linux: 
```
Python 3.11.7 (main, Feb  1 2024, 05:12:23) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.gethostbyname_ex("host.docker.internal")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
socket.gaierror: [Errno -2] Name or service not known
```

PR #4978, opened 2024-04-09
Fix start command for docs

farwill:
## Description

Checklist:

## Rationale
Because docs service moved to docs.yml from local.yml. change the command in documents.rst


browniebroke (approved):

browniebroke:
Thanks!

PR #5012, opened 2024-04-19
Update start-flower in flower to wait until all celery workers are online

arnav13081994:
Flower needs to start only after the celery workers come online. otherwise one will see errors like `inspect method failed` etc

## Description

Checklist:


foarsitter (approved):
Not an expert at this but I like what I'm seeing here!

browniebroke (approved):

PR #5020, opened 2024-04-23
Disable UP038 Ruff rule to avoid introducing slower code

luzfcb:
## Description

Disable the [UP038](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/) Ruff rule to avoid introducing valid code, which is slower due to the lack of optimization in the Python interpreter itself.

This rule also does not exist in `pyupgrade` for the same reason.

Checklist:

## Rationale

Although Python 3.10 starts to support the use of Union in `isinstance` and `issubclass`, there is a considerable performance loss compared to using `tuple`, which has not yet been fully resolved in the Python interpreter.

[Python 3.11 has improved performance but is still 4x slower than using tuples](https://github.com/python/cpython/issues/91603#issue-1206079854). Maybe it will be resolved in Python 3.13.
Check https://github.com/astral-sh/ruff/issues/7871


foarsitter (approved):
Nice catch!

PR #5059, opened 2024-05-08
Remove obsolete `version` element from docker compose files

mpsantos:
## Description

Removed the property 'version' from the docker compose files

Checklist:

## Rationale

The property 'version' now gives a warning that is deprecated. As this is an [optional property](https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-and-name-top-level-elements), I think it's a good idea to remove it.


browniebroke (approved):

browniebroke:
Thanks!

PR #5073, opened 2024-05-15
Update Two scoops of Django book link & image

GeoMaciolek:
## Description

The book link for "Two Scoops of Django 3.x" was a 404; updated to the current page & pointed to a current cover image. Changed the HTML `` tags to use markdown instead, removed `<p>` center tags

Checklist:

  - (Not relevant for documentation)

## Rationale

Just small updates to direct folks correctly to be able to support the relevant author.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,19 +65,15 @@ This project is an open source project run by volunteers. You can sponsor us via
 
 Projects that provide financial support to the maintainers:
 
----
+### Two Scoops of Django
 
-<p align="center">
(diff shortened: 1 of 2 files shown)


GeoMaciolek:
Looks like `faq.rst` also needs the link updated; I'm not sure what the cleanest way to add that in would be. (Probably I should clone the repo locally and do an actual edit rather than the simplistic github web-based editor - oops!)

browniebroke (approved):

PR #5094, opened 2024-05-26
Rename SendInBlue ESP to Brevo

paramkpr:
## Description

This is a really tiny change! [SendInBlue renamed themselves to Brevo in 2023](https://www.brevo.com/blog/becoming-brevo/). I am simply suggesting that we reflect that in the cookiecutter.json so that users can see it when they're configuring a new project. I propose that we change the option name to `Brevo (Previously SendInBlue)`. 

Checklist:

## Rationale

Moving forward, an increasing number of users will be familiar with 'Brevo' instead of 'SendInBlue' and might simply assume that django-cookiecutter (or rather AnyMail) doesn't support Brevo. This change would avoid confusion.


paramkpr:
Fixed all the errors and tested that these changes do actually work for sends mails using Brevo. 

Anymail has also refactored to support the name change so it was fairly straightforward.

browniebroke (approved):

browniebroke:
Thanks for this, I didn't know they changed their name.

PR #5127, opened 2024-06-07
Honor env variable WEB_CONCURRENCY when running gunicorn

areski:
## Description

At the moment, `WEB_CONCURRENCY` env variable is only used on Heroku & PythonAnywhere.
We expect it to work when running with docker, this patch allow to configure the number of worker which by default is 4 on gunicorn.

Checklist:

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/django/start
+++ b/{{cookiecutter.project_slug}}/compose/production/django/start
@@ -27,6 +27,10 @@ if compress_enabled; then
   python /app/manage.py compress
 fi
 {%- endif %}
+
+# The number of worker processes for handling requests. [default: 4]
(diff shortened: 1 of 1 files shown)


browniebroke:
> WEB_CONCURRENCY env variable is only used on Heroku

I don't see it in the command line of the Procfile, do you know how it's used on Heroku?

https://github.com/cookiecutter/cookiecutter-django/blob/db38d8b00f057430744bf6c7a487208d939517b1/%7B%7Bcookiecutter.project_slug%7D%7D/Procfile#L3-L5

areski:
After some digging -> https://github.com/cookiecutter/cookiecutter-django/commit/52619563563c6347ff66b166abe0c9a01bf26a80
It

https://docs.gunicorn.org/en/stable/settings.html#workers

So maybe we don't need this patch.

PR #5164, opened 2024-06-24
Fix keywords case warning in Dockerfile

browniebroke:
## Description

Recently noticed the following warning in my Docker builds (using https://github.com/docker/build-push-action):

> WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match

I don't think it's causing any harm, apart perhaps some small noise in the logs... Seems like a good thing to do for consistency of keywords.

Checklist:

## Rationale

- Fix noisy warning
- More consistent usage of keywords in Dockerfile
- https://docs.docker.com/reference/build-checks/from-as-casing/


luzfcb (approved):
LGTM

PR #5248, opened 2024-07-24
fix non existent of two scoops of django image

thuvh:
## Description

fix image url

Checklist:


luzfcb (approved):

luzfcb:
@thuvh Good catch. thanks.

PR #5252, opened 2024-07-26
Resolve https documentation around local development environment

### LINKED ISSUE #2759: /etc/host configuration docs needed
## What happened?
This is a bug w/ the docs, I think there is a missing section about configuring the `/etc/host` file to allow for local connections through a url like `my-dev-env.local`. I asked the question on [SO](https://stackoverflow.com/questions/63525029/unable-to-connect-to-django-app-in-docker-container-with-nginx) and the response allowed this to work for me.

Am I missing something here or is it necessary to update `etc/host`?

Details about configuring the `/etc/host` file on the local machine to forward the request.

[//]: # (Any or all of the following:)
[//]: # (* Host system configuration: OS, Docker & friends' versions etc.)
[//]: # (* Replay file https://cookiecutter.readthedocs.io/en/latest/advanced/replay.html)
[//]: # (* Logs)

### LINKED ISSUE #2840: https://my-dev-env.local serving pages from both django and docs containers

### LINKED ISSUE #5184: Add ngrok for developing locally with HTTPS
## Description

Change the documentation for developing locally with HTTPS to use ngrok. 

## Rationale

The existing documentation no longer works as described. 

## Implementation 

ngrok allows a user to connect localhost to the internet for testing applications and APIs

1. Install ngrok 

- `brew install ngrok/ngrok/ngrok`
or
- `npm install -g ngrok`

2. Register for account - get AUTH_TOKEN

4. `ngrok config add-authtoken <AUTH_TOKEN>`

5. `ngrok http localhost:8000 - get URL`

6. In config/settings/local.py

- add ngrok domain to ALLOWED_HOSTS
- `ALLOWED_HOSTS` = ["localhost", "0.0.0.0", "127.0.0.1", `".ngrok-free.app"]`

- add HTTPS configuration
- `CSRF_TRUSTED_ORIGINS = ['https://*.ngrok-free.app']`
- `SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')`
- `SESSION_COOKIE_SECURE = True`
- `CSRF_COOKIE_SECURE = True`

ngrok URL should now serve localhost

Afrowave:
## Description
This PR seeks to solve outdated instructions on how to set up a secure development environment in Docker.

I have created a blog post that I can easily keep updated and goes into more detail on how to set up the environment with the default `nginx` Docker image. I have also added Nginx context to the `webpack` section so that it works as a stand-alone explanation.

Checklist:

## Rationale
This is to help those who want to set up a secure local web environment for development.

This fixes the issues that came about in:

- Fix #2840 
- Fix #5184 
- Fix #2759
- Close #3111


browniebroke:
Thanks!

PR #5306, opened 2024-08-17
Migrate from traefik+nginx to angie on production

ghostforpy:
## Description

Checklist:


luzfcb:
Thanks for the PR,  but we are not going to migrate to Angie at this time.

PR #5312, opened 2024-08-21
Add type hint for UserFactory

browniebroke:
## Description

Since factory-boy 3.3.1, we can type annotate factories:

https://github.com/FactoryBoy/factory_boy/pull/1059

Setting that up for the only factory of the project

Checklist:

- [x] I've made sure that tests are updated accordingly (especially if adding or updating a template option): rely on existing mypy test

## Rationale

Demonstrate best practice as part of the project


foarsitter (approved):
This is awesome!

PR #5327, opened 2024-08-27
Use wait-for-it in favor of the custom python script when waiting for postgres

foarsitter:
Proposing the usage of wait-for-it in favor of our own python script as raised in https://github.com/cookiecutter/cookiecutter-django/issues/4539

wait-for-it can also be used in the healthcheck command but that is out of scope for now.


luzfcb:
Does `wait-for-it` ensure that host + port is reachable and the host + port + password + database set is 100% functional and can execute queries?

As far as I know, it is one thing the PostgreSQL process to be active on a certain port, and another thing for PostgreSQL to be active and able to answer queries for a specific set of user + password + database.

Specifically for health checks, there is a case where the PostgreSQL process is active on a port, but is not yet ready to receive queries. This may take a few milliseconds or seconds. That said, I don't remember where I read this. Maybe I'm wrong.

PR #5335, opened 2024-08-30
Pin watchdog to 4.0.2

novucs:
Watchdog 5.0.0 was recently released. This update causes Werkzeug to reload the Django server much more frequently than desired (causing the local docker development build to barely load).

## Description

Checklist:

## Rationale

Without this change, playing around with the local docker development build consistently fails with ECONNRESET due to page loads triggering inotify events and therefore django server reboots

I've not tried to reproduce this issue on other environments, but this is the case on my M3 Mac.

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/local.txt
+++ b/{{cookiecutter.project_slug}}/requirements/local.txt
@@ -1,5 +1,6 @@
 -r production.txt
 
+watchdog==4.0.2 # https://github.com/gorakhargosh/watchdog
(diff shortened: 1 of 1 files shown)


browniebroke:
Would it be worth pinning it in werkzeug instead?

browniebroke (approved):
Tested on a project, I had the same issue and this fixed it

browniebroke:
I think this has now been fixed in werkzeug 3.0.5 so I'll remove this pin when I merge the upgrade: https://github.com/pallets/werkzeug/issues/2945

Issue #5344, opened 2024-09-04
Bump NGINX to the latest version


Hoffmann77:
## Description

The version of NGINX used in /compose/production/nginx/Dockerfile is outdated and should be bumped to the newest version.

## Rationale

Why should this feature be implemented?

PR #5345, opened 2024-09-05
Add nginx container to dependabot automation

luzfcb:
- Add nginx container to dependabot automation
- Fix the URL to the dependabot documentation.

Checklist:

## Rationale

- Helps prevent #5344 from happening again


luzfcb:
okay, looks like trying to update dependabot configuration makes it starting to fail 
https://github.com/dependabot/dependabot-core/issues/10525

browniebroke:
> okay, looks like trying to update dependabot configuration makes it starting to fail [dependabot/dependabot-core#10525](https://github.com/dependabot/dependabot-core/issues/10525)

We've just received a bunch of PRs to update the docker images for Python so I it might be failing just in the Dependabot file chek that runs on CI...

browniebroke (approved):

luzfcb:
Dependabot changes the syntax. "Special" characters need to be escaped with \\ in the dependabot config file.

https://github.com/dependabot/dependabot-core/issues/10525#issuecomment-3618116627

Issue #5346, opened 2024-09-05
Docs (sphinx-autobuild) fails with Docker


ljfp:
## What happened?
The container executes the following command at startup:
sphinx-autobuild -b html --host 0.0.0.0 --port 9000 --watch /app -c . . ./_build/html
It fails with "Errno 2" because the directory '_build' doesn't exist.
The container stops, docs are not generated.

## What should've happened instead?
Container should run and docs should be accessible at http://localhost:9000/

## Additional details

Logs from the container:
2024-09-05 08:13:45 sphinx-autobuild -b html --host 0.0.0.0 --port 9000 --watch /app -c . . ./_build/html
2024-09-05 08:13:46 Traceback (most recent call last):
2024-09-05 08:13:46   File "/usr/local/bin/sphinx-autobuild", line 8, in <module>
2024-09-05 08:13:46     sys.exit(main())
2024-09-05 08:13:46              ^^^^^^
2024-09-05 08:13:46   File "/usr/local/lib/python3.12/site-packages/sphinx_autobuild/__main__.py", line 34, in main
2024-09-05 08:13:46     args, build_args = _parse_args(list(argv))
2024-09-05 08:13:46                        ^^^^^^^^^^^^^^^^^^^^^^^
2024-09-05 08:13:46   File "/usr/local/lib/python3.12/site-packages/sphinx_autobuild/__main__.py", line 101, in _parse_args
2024-09-05 08:13:46     args.outdir = Path(sphinx_args.outputdir).resolve(strict=True)
2024-09-05 08:13:46                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-09-05 08:13:46   File "/usr/local/lib/python3.12/pathlib.py", line 1240, in resolve
2024-09-05 08:13:46     s = self._flavour.realpath(self, strict=strict)
2024-09-05 08:13:46         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-09-05 08:13:46   File "<frozen posixpath>", line 427, in realpath
2024-09-05 08:13:46   File "<frozen posixpath>", line 462, in _joinrealpath
2024-09-05 08:13:46 FileNotFoundError: [Errno 2] No such file or directory: '_build'
2024-09-05 08:13:46 make: *** [Makefile:20: livehtml] Error 1

- Host system configuration:

[...]

PR #5347, opened 2024-09-05
Fix Makefile and make.bat files used for automatic generation of documentation.

ljfp:
## Description

Fixing this bug --> https://github.com/cookiecutter/cookiecutter-django/issues/5346
Checklist:

I've used Tox to run the templates test suite.
This doesn't seem relevant for the fix.

## Rationale

Fix: #5346

Both the "Makefile" and "make.bat" files are passing a non-existent argument ("help") to the -M/-b option. These are the only valid arguments: https://www.sphinx-doc.org/en/master/usage/builders/index.html

From that list, the most logical Builder to use is "html" since that's the one that is used later with sphinx-autobuild.

### DIFF
--- a/{{cookiecutter.project_slug}}/docs/Makefile
+++ b/{{cookiecutter.project_slug}}/docs/Makefile
@@ -13,11 +13,11 @@ APP = /app
 APP = ../{{cookiecutter.project_slug}}
 {% endif %}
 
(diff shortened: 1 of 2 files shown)


foarsitter (approved):

foarsitter:
Thanks for your contribution @ljfp!

PR #5351, opened 2024-09-05
Merge setup.py in pyproject.toml

foarsitter:
The releases to PyPI have been abandoned since early 2018, so the setup.py file is no longer needed. I have merged all the metadata into pyproject.toml but left the original values intact, as it doesn’t interfere with anything but can be improved in the future.


pyproject.toml
```diff
@@ -48,3 +48,27 @@ indent_size = 2
...
+    "Intended Audience :: Developers",
+    "Natural Language :: English",
+    "License :: OSI Approved :: BSD License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Software Development",
+]
+keywords = ["cookiecutter", "Python", "projects", "project templates", " django", "skeleton", "scaffolding", "project directory", "setup.py"]
```
browniebroke, line 72:
```suggestion
keywords = ["cookiecutter", "Python", "projects", "project templates", " django", "skeleton", "scaffolding", "project directory"]
```

browniebroke (approved):
Looks good. Perhaps it would be easier to find if the section was at the top of the `pyproject.toml` file?

pyproject.toml
foarsitter: If we decide to change the keywords I would opt for removing Python, projects, and project directory to.

foarsitter:
Added https://pypi.org/project/pyproject-fmt/ as pre-commit hook for an uniform formatting and removed some keywords.

pyproject.toml
browniebroke: Fair point, yes 👍🏻

browniebroke (approved):

foarsitter:
Thanks for your feedback @browniebroke!

PR #5356, opened 2024-09-06
Replace pip with uv for the project files

foarsitter:
## Description

`uv` is maturing out and I played around a bit by replacing pip with uv for the project files. The generated project template stay's untouched. 

### Dependencies
The dependencies located in requirements.txt are now included in `pyproject.toml` in the `project.dependencies` section. `pyproject-fmt` sorts them alphabetical so I had to remove the comments. There is also a lock file (uv.lock) but dependabot doesn't have support for this at the moment.  

### Actions
Each action uses `astral-sh/setup-uv@v2` for installing and caching the dependencies. From what I saw the actions are a few seconds faster but not much. 

### Scripts

The scripts located in `/scripts/` have now a dependencies section for standalone execution: https://docs.astral.sh/uv/guides/scripts/ and are run by invoking `uv run scripts/create_django_issue.py`

Looking forward to implement `uv` in the project template to!


browniebroke:
Sounds cool! 

> There is also a lock file (uv.lock) but dependabot doesn't have support for this at the moment.

Found the issue tracking that https://github.com/dependabot/dependabot-core/issues/10478

.github/workflows/ci.yml
```diff
@@ -23,14 +23,17 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v2
+        with:
+          enable-cache: "true"
       - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
```
browniebroke, line 32: https://github.com/astral-sh/setup-uv?tab=readme-ov-file#do-i-still-need-actionssetup-python-alongside-setup-uv

> This action is modelled as a drop-in replacement for actions/setup-python when using uv. With setup-uv, you can install a specific version of Python using uv python install rather than relying on actions/setup-python.

.github/workflows/ci.yml
```diff
@@ -54,12 +57,15 @@ jobs:
...
-          cache: pip
       - name: Install dependencies
-        run: pip install -r requirements.txt
+        run: uv sync
```
browniebroke, line 68: While it might be benefical to keep the install and run steps separated, I _think_ this step is not 100% needed as uv will do it automatically for you when needed.

pyproject.toml
```diff
@@ -47,7 +67,7 @@ known_first_party = [
   "hooks",
 ]
 
-# ==== djLint ====
```
browniebroke, line 50: 🤦🏻  good catch

scripts/create_django_issue.py
```diff
@@ -1,3 +1,12 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "gitpython",
+#     "PyGithub",
+#     "jinja2",
+#     "requests",
+# ]
+# ///
```
browniebroke, line 9: ✨ 

Question: does uv try to restrict these versions from the project lock file, or does it just use the versions as specified here?

browniebroke:
Alternatively, renovate already has uv support it seems: https://github.com/renovatebot/renovate/pull/30756

.github/workflows/ci.yml
foarsitter: https://docs.astral.sh/uv/guides/integration/github/#setting-up-python 

> Alternatively, the official GitHub setup-python action can be used. This can be faster, because GitHub caches the Python versions alongside the runner.

Installing python can be done through `uv` but it may be slower then `setup-python`. That's why I have chosen for `setup-python`.

scripts/create_django_issue.py
foarsitter: From my understanding it uses the versions specified in the project lock file.

.github/workflows/ci.yml
foarsitter: Removed the `setup-python` step since `uv sync` and `uv run` install python in the virtualenv.

scripts/create_django_issue.py
foarsitter: `uv run` installs the correct environment including python.

foarsitter: A single location for pinning the dependencies has my preference so I removed the inline script metadata. Also I adopted the --frozen flag to avoid updates to the environment.

.github/workflows/.github/workflows/dependabot-uv-lock.yml
browniebroke, line 1: Added this file to try to keep the lock file up to date as Dependabot sends us PRs until they add support for it. What do you think?

browniebroke: New workflows don't run from a fork, so I created this to test it: https://github.com/cookiecutter/cookiecutter-django/pull/5419

browniebroke: Ok that doesn't work very well, I'll remove these commits from here and keep experimenting on the other PR

foarsitter:
Time to merge this before it is getting out of sync with requirements.txt again.

browniebroke:
Nice work!

luzfcb:
Just some additional context reaffirming that:

This change is not intended to replace `pip` with `uv` in the generated project, that is, in the project you get after running `cookiecutter https://github.com/cookiecutter/cookiecutter-django`.

As far as I know, it is not yet possible to replace pip with uv in the generated project because Heroku does not support it yet on the [Heroku's Classic Python Buildpack](https://github.com/heroku/heroku-buildpack-python) and because the Heroku's Classic Python Buildpack requires the `requirements.txt` file on the project root.

In parallel with [Heroku's Classic Python Buildpack](https://github.com/heroku/heroku-buildpack-python), Heroku is developing a new [buildpack standard](https://github.com/heroku/buildpacks) and a new buildpack for Python, the [Heroku's Cloud Native Buildpack for Python applications](https://github.com/heroku/buildpacks-python), which may eventually support `pyproject.toml`, `uv`, etc. However, the `Heroku's Cloud Native Buildpack for Python applications` is still under development and is not production-ready yet.


Related to:

https://github.com/heroku/heroku-buildpack-python/pull/834
https://github.com/heroku/heroku-buildpack-python/issues/796

foarsitter:
The generated project is indeed untouched. But not for long if it is up to me. 

In one of my projects running on Heroku I'm using the pre-commit hook to generate a requirements.txt as described over here: https://docs.astral.sh/uv/guides/integration/pre-commit/ works like a charm because uv is blazing fast.

PR #5376, opened 2024-09-14
Update create_django_issue.py

SantoraLOgfdfaf:
## Description

Checklist:


browniebroke (changes requested):
Why do we need this change? That seems like a step backwards removing some documentation as to why this script exists.

PR #5395, opened 2024-09-23
added PDM dependencies manager.

Aliaksandr-Litvinau:
## Description

Checklist:

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -13,6 +13,7 @@
     "Apache Software License 2.0",
     "Not open source"
   ],
+  "dependency_manager": ["requirements.txt", "PDM"],
   "username_type": ["username", "email"],
   "timezone": "UTC",
   "windows": "n",
--- a/{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
@@ -13,13 +13,20 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   # psycopg dependencies
   libpq-dev
 
+{% if cookiecutter.dependency_manager == 'PDM' %}
+# Устанавливаем PDM
+RUN pip install pdm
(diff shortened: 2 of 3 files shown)


foarsitter:
Hello @Aliaksandr-Litvinau, thanks for your contribution. 

You left the rationale section empty while your choice for pdm is very interesting. Considering the state of poetry and uv, why do you want to support pdm?

Aliaksandr-Litvinau:
@foarsitter Hi,
Reasons for Supporting PDM:
1. Compatibility with PEP 582:
PDM supports PEP 582, which offers an alternative way to manage dependencies without using virtual environments. This can simplify dependency setup and usage, especially for new Python users.
2. Dependency and Package Management:
PDM provides convenient commands for managing dependencies and packages, making it comparable to tools like Poetry and Pipenv. It also supports working with pyproject.toml, making it compatible with modern Python standards.
Speed and Lightweight:
PDM is designed with performance and ease of use in mind. It can be faster and less resource-intensive compared to some other dependency management tools.
4. Integration with PEP 517/518:
PDM fully supports PEP 517 and PEP 518, making it compatible with modern build and dependency management standards in Python.
5. Active Development and Support:
PDM is actively developed and supported by the community. This means it receives regular updates and improvements, making it a reliable choice for dependency management.
Comparison with Poetry and Pipenv:
Poetry:
Poetry is also an excellent tool for managing dependencies and packages, but it can be more complex to set up and use for some users. Additionally, Poetry has had performance and stability issues in the past, although these are actively being addressed.
Pipenv:
Pipenv was popular some time ago, but its development has slowed down, and it does not always support the latest Python standards. PDM can be a more modern and convenient choice.
Conclusion:
Supporting PDM in the project provides users with another powerful and convenient tool for managing dependencies and packages. This can enhance the development experience and simplify environment setup for new users.

foarsitter:
What does ChatGPT say about the fact that PEP 582 has been rejected?

PR #5409, opened 2024-09-28
Update tox.ini typo

Mwauracollins:
## Description

- [ ] I've updated the tox.ini file which had a typo
## Rationale

i don't really know what the codebase is about. I am just excited to contribute


browniebroke (changes requested):
It's not a typo, it's `-r requirements.txt` without the space, which is optional.

PR #5419, opened 2024-10-02
Test uv lock workflow by running on non-fork repo

browniebroke:
## Description

Checklist:


.github/workflows/dependabot-uv-lock.yml
```diff
@@ -0,0 +1,25 @@
+name: uv
+
+on:
+  pull_request:
+    paths:
+      - "../../pyproject.toml"
```
browniebroke, line 6:
```suggestion
      - "pyproject.toml"
```

PR #5421, opened 2024-10-03
Add workflow to run uv lock 

browniebroke:
## Description

Build on top of https://github.com/cookiecutter/cookiecutter-django/pull/5356 to add a uv lock workflow

Checklist:


browniebroke:
Opened from work repo

PR #5422, opened 2024-10-03
Add workflow to run uv lock

browniebroke:
## Description

Build on top of https://github.com/cookiecutter/cookiecutter-django/pull/5356 to add a uv lock workflow

Checklist:


foarsitter:
Adding a dependency through `uv add` also updates  `uv.lock`. So the lock file is kept in sync. 

If we commit 'uv.lock', is a workflow like this then required?

browniebroke:
This workflow is mainly for dependabot PRs, which AFAIK doesn't support `uv.lock` yet. Since it does support deps in `pyproject.toml`, as it's an earlier standard, PRs would leave the lock file out of date.

foarsitter:
Clever!

.github/workflows/dependabot-uv-lock.yml
```diff
@@ -0,0 +1,34 @@
...
+  lock:
+    runs-on: ubuntu-latest
+    env:
+      GH_PAT: ${{ secrets.GH_PAT }}
+    steps:
+      - name: Checkout with token
+        uses: actions/checkout@v4
+        if: ${{ env.GH_PAT != '' }}
+        with:
+          token: ${{ env.GH_PAT }}
```
browniebroke, line 22: This is needed to make sure that CI runs again on pushes, see: https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs

.github/workflows/dependabot-uv-lock.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Checkout without token
+        uses: actions/checkout@v4
+        if: ${{ env.GH_PAT == '' }}
```
browniebroke, line 26: Added this fallback in case the secret isn't accessible, which unfortunately the case sometimes...

foarsitter (approved):

browniebroke:
Ok, let's try this

Issue #3804, opened 2022-07-23
PyUP going to only paid plans


pydanny:
PyUP just sent me notice that on August 1 all repos must be on paid plans.

- [ ] Find an alternative (Perhaps Snyk)
- [ ] Switch from PyUP to alternative

rdegges:
No problemo! [Snyk](https://snyk.io/) is happy to sponsor cookiecutter =) Will send DM with details.

browniebroke:
I noticed today that [our template dependencies](https://github.com/cookiecutter/cookiecutter-django/pull/5612) aren't being kept up to date by dependabot, presumably due to a bug in the bot and also perhaps because we use uv in the template project.

Are there been any progress on this? @rdegges I would be happy to pick this up, I think I should have enough access to this repo. Does Snyk support uv as package manager? 

Alternatively, I've used [Renovate](https://www.mend.io/renovate/) which seemingly support everything under the sun (including uv) and is highly customisable. Also supports auto-merging after a delay for certain types of upgrades (patch/minor, but not major), hence reducing the maintenance efforts.

Issue #3083, opened 2021-03-06
Django container complains "error walking file system" when venv is installed


jmoppel:
## What happened?
After creating a fresh project, and then docker-compose `build` and `up`, The Django Docker container infinitely complains "error walking file system".
## What should've happened instead?
The Django container should start without any errors.
## Additional details
This happens if you have a Python `venv` environment installed anywhere inside the project directory. The problem is, inside the docker container, venv symlinks to files that don't exist in the container (e.g. `/usr/bin/python3.8`).

### Steps to reproduce
Inside a new or existing cookiecutter project directory
```
docker-compose -f local.yml build
docker-compose -f local.yml up
```
Stop the containers (ctrl-c). Now, create a new python venv
```
python -m venv venv
```

Now bring up your containers and the Django Container will constantly complain "error walking file system":
```
docker-compose -f local.yml up
```
I assume this is `watchgod` complaining. Let me know if this is correct or not.

If you're open to this fix, I'll create a PR.

luzfcb:
@jmoppel I was unable to reproduce the issue.
What operating system are you using?

If you want, could you please record the steps you made in the terminal using https://asciinema.org/ and after sending the recording link? (Be careful to avoid showing things that could compromise security, such as passwords, tokens, etc.)

Install asciinema


```bash
pip3 install asciinema
```

Create an account to able to manage your asciinema records

```bash
asciinema auth
```

Start record the steps

```bash
asciinema rec -t "cookiecutter-django issue 3083"
```

jmoppel:
https://asciinema.org/a/L0Rzbuo6EgWi5EZpqgiOL6Vf7

One minor correction to the video. Before creating the video, I did a docker-compose build _and then_ a docker-compose up. In the video I stated it was the other way around, which is not correct.

For anyone who might find this issue in the future, the exact error message in the Docker console is:
`django          | error walking file system: FileNotFoundError [Errno 2] No such file or directory: '/app/venv/bin/python3.8'`

I have created a Docker file that corrects this problem on my machine.

bryanus1:
I have the same error when install a venv, docker show this log: error walking file system: FileNotFoundError [Errno 2] No such file or directory: '/app/venv/bin/python'

jmoppel:
[...]

PR #5434, opened 2024-10-06
Uv generated project

foarsitter:
The project itself now uses `uv` and this is an attempt to bring `uv` to the generated project too.

All checks are green but heroku and the various CI providers haven't been tested. 

Generating the docs with `make html` requires activating the virtualenv manually. 

I used the `watch` directive to prevent the .venv being overridden in docker but I didn' test the production image. 

Feel free to contribute to this draft.


foarsitter:
Related to #3083 #1988

foarsitter:
Ready for testing :tada:

{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -140,3 +140,146 @@ extend-unsafe-fixes = [
...
+[project]
+name = "cookiecutter-django"
+version = "2024.10.04"
+description = "A Cookiecutter template for creating production-ready Django projects quickly."
+readme = "README.md"
+
+license = { text = "BSD" }
+authors = [
+  { name = "Daniel Roy Greenfeld", email = "pydanny@gmail.com" },
+]
```
browniebroke, line 193: That was copied from the template I suppose? These shouk have the variables provided by the user

{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -140,3 +140,146 @@ extend-unsafe-fixes = [
 
 [tool.ruff.lint.isort]
 force-single-line = true
+
+[tool.uv]
+dev-dependencies = [
```
browniebroke, line 145: Might want to use the dependency groups here

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -48,7 +48,14 @@ repos:
     hooks:
       - id: djlint-reformat-django
       - id: djlint-django
-
+{%- if cookiecutter.use_heroku == "y" %}
```
browniebroke, line 51: This file doesn't contain any jinja tags right now in order to make `pre-commit autoupdate` work. Unwanted blocks are removed in post-generation hooks.

Unless pre-commit handles that now, this will break.

browniebroke (commented):
Did a first pass and noted a few things. Haven't reviewed the Dockerfiles yet

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
browniebroke: https://github.com/cookiecutter/cookiecutter-django/blob/fb27881b33aabe4273dab9703df0f420db67c70d/hooks/post_gen_project.py#L220

foarsitter: Ah yes, that is much cleaner. 

Planning to do the same for dependencies in pyproject.toml so we can have a single lockfile for the complete project, but not in this PR ;)

{{cookiecutter.project_slug}}/pyproject.toml
foarsitter: Mist [PEP-735](https://peps.python.org/pep-0735/) but an easy adjustment

steveputman:
Hi, there. I played around with using your fork to update some existing projects and eventually got them working but ended up reverting them to their production (pip) versions so that I could revisit when I had more time.

When working on an adapting an existing project, I got an error during the build, which was reproduced when I built a project from scratch using `cookiecutter https://github.com/foarsitter/cookiecutter-django --checkout uv-generated-project`:

 > [django python-build-stage 4/6] RUN --mount=type=cache,target=/root/.cache/uv     --mount=type=bind,source=uv.lock,target=uv.lock     --mount=type=bind,source=pyproject.toml,target=pyproject.toml     uv sync --frozen --no-install-project --no-dev:
0.112 Using CPython 3.12.7 interpreter at: /usr/local/bin/python
0.112 Creating virtual environment at: .venv
0.113 error: Could not find root package `my-awesome-project`
'

I believe that it's caused by having the blank `uv.lock`. I thought since the local project had the host file `rw` that it would generate a usable `uv.lock` during the local build. Is there a way to get a usable version of `uv.lock` on the host filesystem without running `uv sync` on the host filesystem? Thanks for this--very cool!

foarsitter:
@steveputman thanks looking into this! 


During a build you cannot write to the host system. 

Your interpreter is inside the container together with uv so you can run uv:
```
docker compose -f docker-compose.local.yml run django uv lock
```
Struggled a lot with this and it is not ideal. So hopefully someone has a better solution.

steveputman:
@foarsitter 
The Docker documentation was sort of vague on that, but now I know.

Your solution is better than mine (I had uv installed on my host system so just ran `uv lock` in the project directory.

Thanks for the help! Will let you know if anything else comes up.

{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR ${APP_HOME}
 
 COPY ./package.json ${APP_HOME}
 RUN npm install && npm cache clean --force
-COPY . ${APP_HOME}
+ADD . ${APP_HOME}
```
luzfcb, line 9: Could you explain why the COPY command was replaced with ADD in the Dockerfile?
What specific problem does this change aim to solve?

foarsitter: The https://github.com/astral-sh/uv-docker-example uses ADD, I don't know the difference.

luzfcb: The official documentation on docker ADD vs COPY best practices and some usage examples is very enlightening:

https://docs.docker.com/build/building/best-practices/#add-or-copy

JonathanPiaget: I would also suggest to change from:

```
ADD . ${APP_HOME}
```

to: 
```
COPY ./pyproject.toml ./uv.lock ./
```

in the local django Dockerfile

JonathanPiaget: I would also suggest to change from:

```
ADD . ${APP_HOME}
```

to: 
```
COPY ./pyproject.toml ./uv.lock ./
```

in the local django Dockerfile

foarsitter: Can you elaborate a little more on your requested change?

foarsitter: Reverted the change, using COPY in favor of ADD

rtpg:
I think you need to run `uv sync` on various start scripts (like celery), at least locally. Because you're mounting a folder to `/app/.venv` we actually don't have the installed virtualenv created in the base `Dockerfile`.

Of course we're also not sharing a virtualenv here, by my read/attempt to use this. It might be possible to just share a mount across all the services, but I'm having trouble getting that to work.

pievalentin:
the dependabot would need updating too. I dont know if it support pyproject declarations

leandrodesouzadev:
> The dependabot would need updating too. I dont know if it supports pyproject declarations

I believe it does, I've found some discussions on the project, here's the link to the [issue](https://github.com/dependabot/dependabot-core/issues/3290#issuecomment-1428610595).

browniebroke:
> The dependabot would need updating too. I dont know if it supports pyproject declarations

This has been added now: https://github.com/dependabot/dependabot-core/issues/10478#issuecomment-2691330949

pievalentin:
a git pull away from a merge 😿

foobacca:
Tried it.  I wonder if the requirements currently in `requirements/production.txt` should be moved to an optional dependencies section in `pyproject.toml`, so something like:

``` toml
[project.optional-dependencies]
production = [
  "gunicorn==23.0.0",
  # ...
]
```

Currently they appear in the main project dependencies, so they will be installed in the development virtualenv.

They can then be installed by running `uv sync --extra=production` - or explicitly excluded with `uv sync --no-extra=production`

[uv docs about optional dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies).

foobacca:
@foarsitter I see your point (though you can of course install the production dependencies if you want with `uv sync --extra=production` ).

I don't have a strong preference, but it does seem like a change from the master branch, where there is a separate `requirements/production.txt` file, suggesting that the maintainers liked to keep the production dependencies separate.  So I wanted to make sure that change was explicit and intended.

I'll leave it to the maintainers.  If they're happy with the change I have no objection.  (And I'm quite capable of changing my project after running cookiecutter anyway :slightly_smiling_face: )

browniebroke:
> I don't have a strong preference, but it does seem like a change from the master branch, where there is a separate `requirements/production.txt` file, suggesting that the maintainers liked to keep the production dependencies separate. So I wanted to make sure that change was explicit and intended.

The production deps are installed locally on the master branch too, so not so much of change - see https://github.com/cookiecutter/cookiecutter-django/pull/4838 for more context

pievalentin:
> One other idea. I would expect the `[project]` and `[dependency-groups]` section to be the first and second sections in `pyproject.toml` - putting the overall project definition and dependencies at the top where it's easy to find them.
> 
> I can obviously edit it myself after running cookiecutter, so not a big deal if others prefer it at the bottom, but thought I'd mention it.

I agree. Dependencies should be at the top of the file

foarsitter:
We use https://github.com/tox-dev/pyproject-fmt for dictating the order of appearance, so you don't need to worry about it :)

foarsitter:
Rebased with main and added some documentation about the lock file. After the initial build `uv lock` needs to be executed to generate a lock file and, for convenience, the build should be executed again taking advantage of the lock file. 

I think it is finally ready to be merged!

tests/test_hooks.py
browniebroke, line 1: What does this mean? Did you just changed the file permissions?

{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -140,3 +140,146 @@ extend-unsafe-fixes = [
...
+requires-python = "==3.12.*"
+dependencies = [
+  "python-slugify==8.0.4", # https://github.com/un33k/python-slugify
+  "Pillow==11.1.0", # https://github.com/python-pillow/Pillow
+  {%- if cookiecutter.frontend_pipeline == 'Django Compressor' %}
+  {%- if cookiecutter.windows == 'y' and cookiecutter.use_docker == 'n' %}
+  "rcssmin==1.1.2", # --install-option="--without-c-extensions"  # https://github.com/ndparker/rcssmin
+  {%- else %}
+  "rcssmin==1.1.2", # https://github.com/ndparker/rcssmin
+  {%- endif %}
```
browniebroke, line 203: My main concern is how will we keep these up to date over time. Right now we use PyUP, but I would be very surprised if it supports uv. I wonder if there are other dependencies update services that support Jinja tags in `pyproject.toml` or if we need to roll our own script...

(that will also break our Django issue script, but I'm less worried about it)

foarsitter: Python has built-in support for `toml` so we can manipulate `pyproject.toml` from code and use dependatbot for updating. But this PR is already way to large.

tests/test_hooks.py
foarsitter: Changed file permissions, reverted this change.

{{cookiecutter.project_slug}}/compose/local/docs/Dockerfile
```diff
@@ -17,12 +19,15 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   && rm -rf /var/lib/apt/lists/*
 
 # Requirements are installed here to ensure they will be cached.
-COPY ./requirements /requirements
+RUN --mount=type=cache,target=/root/.cache/uv \
+#    --mount=type=bind,source=uv.lock,target=uv.lock \
```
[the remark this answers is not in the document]
foarsitter: The line should be as is, uncommented. Update the PR.

{{cookiecutter.project_slug}}/pyproject.toml
browniebroke: My point is that the templated file contains Jinja tags, which makes it an invalid toml.

browniebroke: You mean leave all deps in the templated file and remove the unsuitable ones in post gen hooks?

foarsitter: Yes, something like that. The same way that `uv-pre-commit` is removed from ` .pre-commit-config.yaml` when Heroku isn't selected.

foarsitter: Updated `post_gen_project.py` with a quick&dirty`manipulate_pyproject`.

browniebroke:
Apologies, I forgot about this and merged [an upgrade](https://github.com/cookiecutter/cookiecutter-django/pull/5770) which caused a conflict 🙈

hooks/post_gen_project.py
```diff
@@ -502,8 +504,131 @@ def main():
...
+            "Mailjet": "mailjet",
+            "Mandrill": "mandrill",
+            "Postmark": "postmark",
+            "Sendgrid": "sendgrid",
+            "Brevo": "brevo",
+            "SparkPost": "sparkpost",
+            "Other SMTP": "",
+        }
+
+        p = map.get("{{cookiecutter.mail_service}}", None)
```
browniebroke, line 593: Are we expecting all options to be in the above map? If yes, maybe we should do `map["{{cookiecutter.mail_service}}"]` to make it error loudly if one is missing. I'm worried that we might forget this if we add another provider one day

{{cookiecutter.project_slug}}/pyproject.toml
browniebroke: I had another idea on how to handle this: we could stick to requirements in the template (more or less the way it is) and have post-gen hook write the dependencies and dev-dependencies in the pyproject.toml? Might be easier to maintain, and would keep working with PyUp?

foarsitter: That would also work but this way we can also generate a `uv.lock` . It would be outdated after generation but it is better then the empty one we have now.

foarsitter: @browniebroke is this still relevant https://github.com/cookiecutter/cookiecutter-django/issues/3804  for our decision?

browniebroke: > ... this way we can also generate a `uv.lock` . It would be outdated after generation but it is better then the empty one we have now.

I'm not sure about that... I was imagining to not bother with including a lock file in the templated code and let one be generated on the first run. I'm concerned that indirect dependencies might end up locked on an older version while a later one is usually be perfectly fine. We haven't locked them so far and had only a few cases where there was. problem. 

That being said, I'm happy to try out the approach you're suggesting, and time will tell if my concern is valid.

> is this still relevant #3804 for our decision?

It seems highly relevant, yes, as it might influence which way we go here... Unfortunately, I believe don't have permissions to action such move. Maybe @luzfcb has?

browniebroke: > I had another idea on how to handle this: we could stick to requirements in the template (more or less the way it is) and have post-gen hook write the dependencies and dev-dependencies in the pyproject.toml?

Example of commands to run in post-gen hooks:

```
uv add -r requirements/production.txt
uv add --group dev -r requirements/local.txt
```

(Only thing is we'll probably have to amend the `requirements/local.txt` to no longer extend the `requirements/production.txt`)

foarsitter: Nice approach, fascinating to see how powerful `uv` is. I don't have a strong opinion on this matter. 

Tried to use `uv add -r` for `rcssmin==1.1.2 --install-option="--without-c-extensions"` but it seems that  `--install-option` is removed: https://stackoverflow.com/a/76276525

hooks/post_gen_project.py
```diff
@@ -4,6 +4,8 @@
 import string
 from pathlib import Path
 
+from tomlkit import dumps, parse
```
browniebroke, line 7: Oh I just realised that this is a 3rd party package, NOT the `tomllib` module from the standard library (which only focuses on parsing TOML). 

Adding a 3rd party package means that we need to explain to our users that they need to `pip install` / `pipx inject` / `uv tool ... --with` that package before they run the cookiecutter CLI: AFAIK cookiecutter does NOT install the template dependencies before generating it.

That's a significant change in usability for this template, not sure we want to start doing that. 

(that's partly why we're doing some pretty basic YAML parsing in the `remove_prettier_pre_commit` instead of using `pyyaml`, although I just realised that `pyyaml` is a dependency of cookiecutter, so we could probably use it)

foarsitter: That is a significant drawback so I will revert the changes regarding the manipulation of pyproject.toml and use the `uv add -r` hooks.

luzfcb (commented):
I think using bash shellscript in post gen probably won't work on windows unless it's running in a bash shell in WSL (

It maybe might be better to do it in pure python, although it's much more verbose.

I don't have Windows available at the moment, so I don't know if this works.

```
#!/usr/bin/env python3
import os
import subprocess
import shutil
import sys

def setup_dependencies():
    print("Installing python dependencies using uv...")

    {% if cookiecutter.use_docker == "y" %}
    # Build the Docker service using Docker Compose
    try:
        subprocess.run(
            ["docker", "compose", "-f", "docker-compose.local.yml", "build", "django"],
            check=True
        )
    except subprocess.CalledProcessError as e:
        print(f"Error building Docker service: {e}", file=sys.stderr)
        sys.exit(1)
    
    # Use Docker to run the uv command
    uv_cmd = ["docker", "compose", "-f", "docker-compose.local.yml", "run", "--rm", "django", "uv"]
    {% else %}
    # Use uv command directly
    uv_cmd = ["uv"]
    {% endif %}

    # Install production dependencies
    try:
        subprocess.run(
            uv_cmd + ["add", "-r", "requirements/production.txt"],
            check=True
        )
    except subprocess.CalledProcessError as e:
        print(f"Error installing production dependencies: {e}", file=sys.stderr)
        sys.exit(1)

    # Install local (development) dependencies
    try:
        subprocess.run(
            uv_cmd + ["add", "--dev", "-r", "requirements/local.txt"],
            check=True
        )
    except subprocess.CalledProcessError as e:
        print(f"Error installing local dependencies: {e}", file=sys.stderr)
        sys.exit(1)

    # Remove the requirements directory
    if os.path.exists("requirements"):
        try:
            shutil.rmtree("requirements")
        except Exception as e:
            print(f"Error removing 'requirements' folder: {e}", file=sys.stderr)
            sys.exit(1)

    print("Setup complete!")

if __name__ == "__main__":
    setup_dependencies()

```

{{cookiecutter.project_slug}}/pyproject.toml
luzfcb: >>is this still relevant https://github.com/cookiecutter/cookiecutter-django/issues/3804 for our decision?

> It seems highly relevant, yes, as it might influence which way we go here... Unfortunately, I believe don't have permissions to action such move. Maybe @luzfcb has?

@browniebroke @foarsitter Sorry, I just saw this now. I think I have permission to enable almost any integration you want in cookiecutter organization.

foarsitter:
According to the cookiecutter docs .bat files are supported too, should that be a solution @luzfcb otherwise we should go with your version.

foarsitter:
@luzfcb adopted your version because the .bat file didn't work out of the box. I think we are ready!

hooks/post_gen_project.py
```diff
@@ -497,8 +505,52 @@ def main():
...
+    if "{{ cookiecutter.use_docker }}".lower() == "y":
+        # Build the Docker service using Docker Compose
+        try:
+            subprocess.run(["docker", "compose", "-f", "docker-compose.local.yml", "build", "django"], check=True)
+        except subprocess.CalledProcessError as e:
+            print(f"Error building Docker service: {e}", file=sys.stderr)
+            sys.exit(1)
+
+        # Use Docker to run the uv command
+        uv_cmd = ["docker", "compose", "-f", "docker-compose.local.yml", "run", "--rm", "django", "uv"]
```
browniebroke, line 525: Apologies if this was already discussed, I didn't follow the latest changes too closely. Why do we need to run uv in Docker? Why not run the uv commands on the host?

No need to change anything, I'm curious to understand what made you opt for this version...

foarsitter: You need to run uv inside your environment, otherwise it will install Python and all the other packages outside of the container.

browniebroke: Something I'm wondering (as a potential improvement after we merge this), would it be worth providing a more lightweight image just to run this? Feels like an "expensive" step just to add the deps... Again, no need to change anything here, and also I have not tried it yet

foarsitter: Yes, you are right, we only need to calculate the dependency tree here and write it down. We do not need compiling psycopg for example at this stage. But 'uv add' does both of them and if libpg-dev isn't available pyproject.toml will not be updated. 

I didn't researched it any further, so maybe uv has a flag for not installing things but I don't know.

browniebroke: Got it. Yes no problem, we can optimise later

acyment:
@foarsitter I am trying to test the branch for an existing project. I created a blank project with the branch and then tried to merged the differences with mine. I am not sure I completely understood what has happened to package requirements. I copied the packages I had added to local.txt and base.txt to pyproject.toml, but it still gives a ModuleNotFound error. Where should the list of needed packages be when using uv? Thanks!

foarsitter:
Are you using docker? How did you add the requirements? Wich module cannot be found?

acyment:
> Are you using docker? How did you add the requirements? Wich module cannot be found?

Yes, using docker. I had added them to base.txt (they worked fine with pip) and when I saw the error I then added them to pyproject.toml too to no avail. The error is for django-money, but I have added all of these:
"django-money==3.5.3",
"html2text==2024.2.26",
"mjml-python==1.3.5",
"mercadopago==2.3.0",
"stripe==12.0.0",
"nameparser==1.1.3",
"django-afip==13.2.0",
"django-weasyprint==2.4.0",

Error is as follows:  ModuleNotFoundError: No module named 'djmoney' (on the django container)

foarsitter:
You can add all your dependencies with uv as we do in the post gen hooks: https://github.com/cookiecutter/cookiecutter-django/pull/5434/files#diff-8601710441b0dde909812d3f5a1d83245fec1f1bf94b6e510d1b72cbd2fb86c1R525 

First you need to remove the extending (`-r`) of product.txt from local.txt. I think it would be wise to remove all the dependencies you added manually from pyproject.toml so things don't get mixed up.

 Then you need to execute the following commands:

```bash
# project dependencies
docker compose -f docker-compose.local.yml add -r ./requirements/base.txt 
docker compose -f docker-compose.local.yml add -r ./requirements/production.txt 
# dev dependency group
docker compose -f docker-compose.local.yml add -r --dev ./requirements/production.txt 
```
This will modify you `pyproject.toml` but doesn't alter your image. Therefore you need to build the image again with `docker compose -f docker-compose.local.yml build django`. 

I  think you didn't rebuild the image, and therefore, missing `djmoney`.

acyment:
I get an error when trying to execute those commands (unknown docker command: "compose add")

Anyway, I had manually added the dependencies (including django-money) to pyproject.toml and did a build `--no-cache` and the error is still there. I actually did a test with `--progress=plain` to check whether django-money is every installed and no, it is never mentioned (although it sits there in pyproject.toml).__

I even tried running `docker compose -f docker-compose.local.yml run django uv lock` as you mention in an earlier comment before. `uv.lock` correctly references django-money, but the problem persists (uv doesn't seem to realize it has to install django-money). Not sure where it is taking the list of packages it needs to install from.

I might have missed merging something when bringing code over from the empty project I generated using this branch🤷🏻‍♂️

foarsitter:
There is a mistake in the commands, it has to be `uv add`. 

Do you have a .venv in your directory? If you have files in the directory it will override your .venv in your container. 

Did you update the compose file according to this PR?

acyment:
Cool, I managed to run uv add

Error is still there. I didn't have a .venv in my directory (django-money is inside `.venv/lib/python3.12/site-packages`)

I merged all of the changes in django's Dockerfile. I didn't find any particular changes in the actual compose file.

foarsitter:
Do you have the following volumes listed in `docker-compose.local.yml` for the docker service and 100% sure there is no .venv folder? (`rm -rf .venv`)
```yml
volumes:
      - /app/.venv
      - .:/app:z
```

hooks/post_gen_project.py
foarsitter: Noticed a severe slowdown in the Github actions due to this. Adding `--no-sync` to `uv add` seems to be the solution.

acyment:
> Do you have the following volumes listed in `docker-compose.local.yml` for the docker service and 100% sure there is no .venv folder? (`rm -rf .venv`)
> 
> ```yaml
> volumes:
>       - /app/.venv
>       - .:/app:z
> ```
> 
> @acyment made some changes to the Dockerfile, can you try these?

Error still there. Made sure .venv is not there. Here are my local Dockerfile and compose files, in case you can spot what might be at fault. Thanks so much!

Github won't let me attach a Dockerfile or a compose file, so pasting them here:
/compose/local/django/Dockerfile
```
# define an alias for the specific python version used in this file.
FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim AS python

# Python build stage
FROM python AS python-build-stage

ARG APP_HOME=/app

WORKDIR ${APP_HOME}

# we need to move the virtualenv outside of the $APP_HOME directory because it will be overriden by the docker compose mount
ENV UV_COMPILE_BYTECODE=1 UV_LINK_MODE=copy UV_PYTHON_DOWNLOADS=0

# Install apt packages
RUN apt-get update && apt-get install --no-install-recommends -y \
  # dependencies for building Python packages
  build-essential \
  # psycopg dependencies
  libpq-dev \
  gettext \
  wait-for-it \
  weasyprint

# Requirements are installed here to ensure they will be cached.
RUN --mount=type=cache,target=/root/.cache/uv \
  --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
  --mount=type=bind,source=uv.lock,target=uv.lock:rw \
  uv sync --no-install-project

COPY . ${APP_HOME}

RUN --mount=type=cache,target=/root/.cache/uv \
  --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
  --mount=type=bind,source=uv.lock,target=uv.lock:rw \
  uv sync


# devcontainer dependencies and utils
RUN apt-get update && apt-get install --no-install-recommends -y \
  sudo git bash-completion nano ssh

# Create devcontainer user and add it to sudoers
RUN groupadd --gid 1000 dev-user \
  && useradd --uid 1000 --gid dev-user --shell /bin/bash --create-home dev-user \
  && echo dev-user ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/dev-user \
  && chmod 0440 /etc/sudoers.d/dev-user


ENV PATH="/${APP_HOME}/.venv/bin:$PATH"
ENV PYTHONPATH="${APP_HOME}/.venv/lib/python3.12/site-packages:$PYTHONPATH"

COPY ./compose/production/django/entrypoint /entrypoint
RUN sed -i 's/\r$//g' /entrypoint
RUN chmod +x /entrypoint

COPY ./compose/local/django/start /start
RUN sed -i 's/\r$//g' /start
RUN chmod +x /start


COPY ./compose/local/django/celery/worker/start /start-celeryworker
RUN sed -i 's/\r$//g' /start-celeryworker
RUN chmod +x /start-celeryworker

COPY ./compose/local/django/celery/beat/start /start-celerybeat
RUN sed -i 's/\r$//g' /start-celerybeat
RUN chmod +x /start-celerybeat

COPY ./compose/local/django/celery/flower/start /start-flower
RUN sed -i 's/\r$//g' /start-flower
RUN chmod +x /start-flower


ENTRYPOINT ["/entrypoint"]

```
local.yml (I haven't renamed the compose files yet)
```
volumes:
  cyment_com_local_postgres_data: {}
  cyment_com_local_postgres_data_backups: {}
  cyment_com_local_redis_data: {}

services:
  django: &django
    build:
      context: .
      dockerfile: ./compose/local/django/Dockerfile
      tags:
        - 'acyment/django'
    image: cyment_com_local_django
    container_name: cyment_com_local_django
    volumes:
      - /app/.venv
      - .:/app:z
    depends_on:
      - postgres
      - redis
      - mailpit
    env_file:
      - ./.envs/.local/.django
      - ./.envs/.local/.postgres
    ports:
      - '8000:8000'
      - '5678:5678'
    command: /start

  postgres:
    build:
      context: .
      dockerfile: ./compose/production/postgres/Dockerfile
      tags:
        - 'acyment/postgres'
    image: cyment_com_production_postgres
    container_name: cyment_com_local_postgres
    volumes:
      - cyment_com_local_postgres_data:/var/lib/postgresql/data
      - cyment_com_local_postgres_data_backups:/backups
    env_file:
      - ./.envs/.local/.postgres

  mailpit:
    image: docker.io/axllent/mailpit:latest
    container_name: cyment_com_local_mailpit
    ports:
      - '8025:8025'

  redis:
    image: docker.io/redis:6
    container_name: cyment_com_local_redis

    volumes:
      - cyment_com_local_redis_data:/data

  celeryworker:
    <<: *django
    image: cyment_com_local_celeryworker
    container_name: cyment_com_local_celeryworker
    depends_on:
      - redis
      - postgres
      - mailpit
    ports: []
    command: /start-celeryworker

  celerybeat:
    <<: *django
    image: cyment_com_local_celerybeat
    container_name: cyment_com_local_celerybeat
    depends_on:
      - redis
      - postgres
      - mailpit
    ports: []
    command: /start-celerybeat

  flower:
    <<: *django
    image: cyment_com_local_flower
    container_name: cyment_com_local_flower
    ports:
      - '5555:5555'
    command: /start-flower

  node:
    build:
      context: .
      dockerfile: ./compose/local/node/Dockerfile
      tags:
        - 'acyment/node'
    image: cyment_com_local_node
    container_name: cyment_com_local_node
    environment:
      NODE_ENV: development
      NODE_OPTIONS: '--inspect=0.0.0.0:9229' # Expose the debugging port
    depends_on:
      - django
    volumes:
      - .:/app:z
      # http://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html
      - /app/node_modules
    command: bun run dev
    ports:
      - '3000:3000'
      - '9229:9229'
        # Expose browsersync UI: https://www.browsersync.io/docs/options/#option-ui
      - '3001:3001'

```

acyment:
> Your files seems fine, can you try these two things?
> 
>     1. sometimes celeryworker is still runing and therefore the image ins't updated. Can you remove all the running containers (`docker compose down`) and very with `docker compose ps`? After that run `docker compose up django --build`.
> 
> 
> ```
> docker compose -f local.yml down
> docker compose -f local.yml ps
> docker compose -f local.yml up django --build
> ```
> 
>     2. can you try creating a new project using this branch and check if it works:
> 
> 
> ```
> cookiecutter https://github.com/foarsitter/cookiecutter-django --checkout uv-generated-project
> ```

1. Made it work by deleting all containers, images and volumes and rebuilding from scratch. Thanks for the patience!

Just to clarify, when I need to add a new package dependency, how should I proceed from now on? My previous workflow was editing the corresponding requirements.txt file (base.txt / local.txt / production.txt) and then rebuilding the django image. Should I manually add it to pyproject.toml or rather run `uv add` with `docker compose run django`?

pievalentin:
Don't add manually into the `pyproject`. You should always have `uv` manage the dependencies for you. In that case, you have to run `uv add` where you have your venv

foarsitter:
@pievalentin I do not agree, when you are in a non-docker environment it is fine to add to your `pyproject.toml` by hand. `us sync` will update the lockfile. When using docker `uv.lock` cannot be written during the build process to `uv sync` will not be updated accordingly. 

@acyment 

Since the discovery of `--no-sync` my workflow would be:
```bash
# update pyproject.toml and uv.lock
docker compose run --rm django uv add --no-sync django-money
# install the lockfile in a frozen way
docker compose build django
```

I don't know why building the images behaves differently after removing. I think that it has something to do with a running container in the background.

acyment:
In order to make this work in production I had to make the following two changes:

compose>production>django>Dockerfile
51+ RUN chown django:django ${APP_HOME}

compose>production>django>start
-  exec /usr/local/bin/gunicorn config.wsgi --bind 0.0.0.0:5000 --chdir=/app
10+ exec gunicorn config.wsgi --bind 0.0.0.0:5000 --chdir=/app

acyment:
Things in production got really messy, going back to the pip version for now. Images got incredibly bloated, which in turn not just used disk space like crazy, but also made image creation steps take many times longer (defeating the original purpose). What's more, the uv step itself is taking almost the same time as the original pip one. 
My take with respect to bloating is with respect to a couple of COPY statements, not just one. The uv step I am not sure why it takes almost as much as the pip one. Anyway, going back to pip, I spent too much time yesterday trying to fix this. Thanks for the ongoing effort!

foarsitter:
@acyment thanks for your feedback. The production image was indeed missing some changes. 

Not using `uv.lock` may caused your troubles, but I doubt that since I was missing it in my production build to. 

How do you build your images? When building locally you may need to ignore some files in `.dockerignore`.

javi-horizon:
is it safe to start a project using this fork?

foarsitter:
@javi-horizon I think it is.

mpsantos:
I tried applying this PR to an existing project that uses Docker. The image builds normally, but when I try to run the containers locally, I get the following error from the flower service:

```
myproject_local_flower        | Celery workers not available
myproject_local_flower        | timeout: failed to run command 'celery': No such file or directory
```

It seems like the celery command is missing. Any idea what might be causing this?

mpsantos:
> Can you check these two options:
> 
> * the local .venv folder is overriding the one in the container, see the changes in the docker-compose.yml files
> * the PYTHONPATH variable isn't updated in the Dockerfile

I got it working. After building the image for the first time, I ran these commands:

```
docker compose -f docker-compose.local.yml run --rm django uv add --no-sync -r requirements/base.txt
docker compose -f docker-compose.local.yml run --rm django uv add --no-sync -r requirements/production.txt
docker compose -f docker-compose.local.yml run --rm django uv add --no-sync -r requirements/local.txt
```
This populated my uv.lock file. After rebuilding the image, everything worked.
I’m not sure if this is the intended workflow, I’ll need to check UV’s documentation.

foarsitter:
@mpsantos this is indeed what we are doing in `setup_dependencies()` in `post_gen_project.py`. 

Did you forget the --dev flag for the local.txt dependencies? The should be in the dev dependency group.

timoijas-ap:
Sorry if this is not related but does cookiecutter-django support uv if it is used without Docker? I'm trying to find a low effort way of trying to spin up django projects and would like to use uv but not Docker. Is there an out of the box option for this?

foarsitter:
You can delete them

foarsitter:
@mpsantos you can fix the file permissions on your machine and commit it I guess. The container created uv.lock I think and therefore uv.lock lacks the right permissions.

GitBib:
I found a good article describing the optimal way to use uv in Docker.
Maybe we should switch to this approach for a better project build process?
https://rob.cogit8.org/posts/optimizing-django-docker-builds-with-astrals-uv/

hooks/post_gen_project.py
```diff
@@ -438,6 +445,7 @@ def main():
 
     if "{{ cookiecutter.use_heroku }}".lower() == "n":
         remove_heroku_files()
+        remove_repo_from_pre_commit_config("uv-pre-commit")
```
browniebroke, line 448: Heroku [just added support for uv](https://www.heroku.com/blog/local-speed-smooth-deploys-heroku-adds-support-uv/), so this is no longer needed 🎉

Jahongir-Qurbonov:
@foarsitter, @browniebroke This is a very important change, the speed of the build process changes serially. This pr should be merged

baipaiha:
Any news on that ?  difficult to start a new project today without using `uv`.

Quick test with this `uv-generated-project` branch.  Seems ok.
Just one thing: 
  - Creating the project automatically start the docker build.  I think it's part of the tests, but it make the process of "creating a project from a template"  a bit heavy and time consuming

foarsitter:
@baipaiha the build is triggered in order to compile uv.lock. Creating a project isn't a daily task so performance shouldn't be an issue I suppose?

browniebroke:
> compile uv.lock

I think we could potentially do that outside of Docker (on the host) if that wasn't for psycopg which may requires some compilation steps... Feels like compilation isn't strictly needed to resolve the dependency graph but last time I checked, I _think_ uv was doing it anyway

baipaiha:
> @baipaiha the build is triggered in order to compile uv.lock. Creating a project isn't a daily task so performance shouldn't be an issue I suppose?

Ok i see, it make sense.  Performance not an issue on normal usage, i agree. It's just to feedback about a change in behavior with the pip version.

Great if it cam be merged !

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
```diff
@@ -1,37 +1,37 @@
...
-ENV BUILD_ENV=${BUILD_ENVIRONMENT}
+COPY . ${APP_HOME}
 
-WORKDIR ${APP_HOME}
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=uv.lock,target=uv.lock:rw \
+    uv sync
 
 {% if cookiecutter.use_docker == "y" %}
```
browniebroke, line 36: Not related to this PR, but feels strange to have this condition in the `Dockerfile`... This file is removed if `use_docker == "n"` so should always be true?

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
```diff
@@ -1,37 +1,37 @@
...
-
-# Python 'run' stage
-FROM python AS python-run-stage
-
-ARG BUILD_ENVIRONMENT=local
-ARG APP_HOME=/app
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=uv.lock,target=uv.lock:rw \
+    uv sync --no-install-project
```
browniebroke, line 27: Why is `uv sync --no-install-project` here and `uv sync` again after adding our code?

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
```diff
@@ -1,37 +1,37 @@
...
 # Requirements are installed here to ensure they will be cached.
-COPY ./requirements .
-
-# Create Python Dependency and Sub-Dependency Wheels.
-RUN pip wheel --wheel-dir /usr/src/app/wheels \
-  -r ${BUILD_ENVIRONMENT}.txt
-
-
-# Python 'run' stage
-FROM python AS python-run-stage
```
browniebroke, line 25: Am I missing something or are we dropping the local multi-stage build in this PR?

{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
foarsitter: Indeed very strange xD will remove it.

{{cookiecutter.project_slug}}/docker-compose.docs.yml
```diff
@@ -9,8 +9,6 @@ services:
       - ./.envs/.local/.django
     volumes:
       - ./docs:/docs:z
-      - ./config:/app/config:z
-      - ./{{ cookiecutter.project_slug }}:/app/{{ cookiecutter.project_slug }}:z
```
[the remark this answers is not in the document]
foarsitter: That seems broken

hooks/post_gen_project.py
browniebroke: I just tested that it and works, so I suggest we remove this hook entirely:

```suggestion
```

<details>
<summary>Logs</summary>

<pre>
remote: Resolving deltas: 100% (13/13), done.
remote: Updated 117 paths from 1997ad2
remote: Compressing source files... done.
remote: Building source:
remote:
remote: -----> Building on the Heroku-24 stack
remote: -----> Using buildpack: heroku/python
remote: -----> Python app detected
remote: -----> Using Python 3.12 specified in .python-version
remote: -----> Installing Python 3.12.11
remote: -----> Installing uv 0.8.13
remote: -----> Installing dependencies using 'uv sync --locked --no-default-groups'
remote:        Resolved 155 packages in 1ms
remote:        Downloading django (7.9MiB)
remote:        Downloading cryptography (4.3MiB)
remote:        Downloading uvloop (4.5MiB)
remote:        Downloading pillow (6.3MiB)
remote:        Downloading botocore (13.3MiB)
remote:        Downloading djangorestframework (1.0MiB)
remote:           Building psycopg-c==3.2.9
remote:         Downloading djangorestframework
remote:         Downloading uvloop
remote:         Downloading cryptography
remote:         Downloading pillow
remote:         Downloading botocore
remote:           Building django-allauth==65.11.1
remote:         Downloading django
remote:              Built django-allauth==65.11.1
remote:              Built psycopg-c==3.2.9
remote:        Prepared 82 packages in 56.47s
remote:        Installed 82 packages in 110ms
remote:        Bytecode compiled 3734 files in 1.70s
remote:         + amqp==5.3.1
remote:         + anyio==4.10.0
remote:         + argon2-cffi==25.1.0
remote:         + argon2-cffi-bindings==25.1.0
remote:         + asgiref==3.9.1
remote:         + attrs==25.3.0
remote:         + billiard==4.2.1
remote:         + boto3==1.40.21
remote:         + botocore==1.40.21
remote:         + celery==5.5.3
remote:         + certifi==2025.8.3
remote:         + cffi==1.17.1
remote:         + charset-normalizer==3.4.3
remote:         + click==8.2.1
remote:         + click-didyoumean==0.3.1
remote:         + click-plugins==1.1.1.2
remote:         + click-repl==0.3.0
remote:         + collectfasta==3.3.1
remote:         + crispy-bootstrap5==2025.6
remote:         + cron-descriptor==2.0.5
remote:         + cryptography==45.0.7
remote:         + django==5.1.11
remote:         + django-allauth==65.11.1
remote:         + django-anymail==13.0.1
remote:         + django-appconf==1.1.0
remote:         + django-celery-beat==2.8.1
remote:         + django-compressor==4.5.1
remote:         + django-cors-headers==4.7.0
remote:         + django-crispy-forms==2.4
remote:         + django-environ==0.12.0
remote:         + django-model-utils==5.0.0
remote:         + django-redis==6.0.0
remote:         + django-storages==1.14.6
remote:         + django-timezone-field==7.1
remote:         + djangorestframework==3.16.1
remote:         + drf-spectacular==0.28.0
remote:         + fido2==2.0.0
remote:         + gunicorn==23.0.0
remote:         + h11==0.16.0
remote:         + hiredis==3.2.1
remote:         + httptools==0.6.4
remote:         + idna==3.10
remote:         + inflection==0.5.1
remote:         + jmespath==1.0.1
remote:         + jsonschema==4.25.1
remote:         + jsonschema-specifications==2025.4.1
remote:         + kombu==5.5.4
remote:         + packaging==25.0
remote:         + pillow==11.3.0
remote:         + prompt-toolkit==3.0.52
remote:         + psycopg==3.2.9
remote:         + psycopg-c==3.2.9
remote:         + pycparser==2.22
remote:         + python-crontab==3.3.0
remote:         + python-dateutil==2.9.0.post0
remote:         + python-dotenv==1.1.1
remote:         + python-slugify==8.0.4
remote:         + pyyaml==6.0.2
remote:         + qrcode==8.2
remote:         + rcssmin==1.1.2
remote:         + redis==6.4.0
remote:         + referencing==0.36.2
remote:         + requests==2.32.5
remote:         + rjsmin==1.2.2
remote:         + rpds-py==0.27.1
remote:         + s3transfer==0.13.1
remote:         + sentry-sdk==2.35.1
remote:         + six==1.17.0
remote:         + sniffio==1.3.1
remote:         + sqlparse==0.5.3
remote:         + text-unidecode==1.3
remote:         + typing-extensions==4.15.0
remote:         + tzdata==2025.2
remote:         + uritemplate==4.2.0
remote:         + urllib3==2.5.0
remote:         + uvicorn==0.35.0
remote:         + uvicorn-worker==0.3.0
remote:         + uvloop==0.21.0
remote:         + vine==5.1.0
remote:         + watchfiles==1.1.0
remote:         + wcwidth==0.2.13
remote:         + websockets==15.0.1
</pre>

</details>

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -50,6 +50,13 @@ repos:
       - id: djlint-reformat-django
       - id: djlint-django
 
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    # uv version.
+    rev: 0.8.14
+    hooks:
+      - id: uv-export
+        args: ['--frozen', '--no-dev']
+
```
browniebroke, line 59:
```suggestion
```

browniebroke:
Just realised that there is the question of PythonAnywhere too: https://cookiecutter-django.readthedocs.io/en/latest/3-deployment/deployment-on-pythonanywhere.html

Not sure what do about that:

https://github.com/cookiecutter/cookiecutter-django/blob/3524f6d352851ec3717d1eb37d7d106c70c72cf8/docs/3-deployment/deployment-on-pythonanywhere.rst?plain=1#L33-L40

I haven't got much experience with this platform, do they support uv?

EDIT: Looks like [they do](https://blog.pythonanywhere.com/219/#python-packages)

> Meta: if you feel that pip is a bit too old-school for your package-management needs, you can try poetry, hatch, or uv.

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
foarsitter: @baipaiha can you open a PR against this repo?

foarsitter:
Ok, I think we are ready to merge.

browniebroke (commented):
Nearly there 💪

hooks/post_gen_project.py
```diff
@@ -77,7 +80,7 @@ def remove_utility_files():
 
 
 def remove_heroku_files():
-    file_names = ["Procfile", "requirements.txt"]
+    file_names = ["Procfile"]
     for file_name in file_names:
         if file_name == "requirements.txt" and "{{ cookiecutter.ci_tool }}".lower() == "travis":
```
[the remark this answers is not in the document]
browniebroke: Yes, Heroku supports uv

{{cookiecutter.project_slug}}/uv.lock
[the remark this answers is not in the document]
timoijas-ap: On the edge of my seat for this one to be merged :eye:

foarsitter:
Tests are failing now and I don't know why. Feel free to take over, need to fix something else first xD

browniebroke:
> Tests are failing now and I don't know why. Feel free to take over, need to fix something else first xD

~I think it's something to do with the duplicated line ending in the pre-commit config... Not sure why but it passes when they're back~

EDIT: I've just seen that you reverted that and it's still failing 😭

baipaiha:
Failing tests seems related to docker container name conflict, like if test containers were not removed/clean.
Or not related at all ?

```
Error response from daemon: Conflict. The container name "/my_awesome_project_local_postgres" is already in use by container "3f4f06979b88e54037ac7ef161af59f2191fe50417e6877844024c94bcb25592". You have to remove (or rename) that container to be able to reuse that name.
```

browniebroke:
I've changed this approach a bit by defining a minimal Docker image to run `uv add` commands and run it as a standalone Docker container (without compose involved). This is much faster to build and avoids having to start the other containers while generating the template, just for resolving dependencies.

foarsitter:
> I've changed this approach a bit by defining a minimal Docker image to run `uv add` commands and run it as a standalone Docker container (without compose involved). This is much faster to build and avoids having to start the other containers while generating the template, just for resolving dependencies.

Sounds like a good fundamental solution to me, thanks! 

I tried it this way using the uv image directly but I couldn't get it to work. Looking to the Dockerfile I'm wondering why it works this way but I'm fine with it.

From my perspective we are ready to merge this as is!

browniebroke:
> From my perspective we are ready to merge this as is!

Let's do it! 🎉🎉🎉

PR #5440, opened 2024-10-07
Use uv sync & uv run in docs build

browniebroke:
## Description

The offical RTD guide was written a few months ago and I realised after merging https://github.com/cookiecutter/cookiecutter-django/pull/5438 that we can in fact simplify a bit more.

Also upstreamed these changes to RTD: https://github.com/readthedocs/readthedocs.org/pull/11655

Checklist:

## Rationale

Simpler build steps

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,7 +14,7 @@ build:
     - asdf install uv latest
     - asdf global uv latest
     - uv sync --extra docs
-    - .venv/bin/python -m sphinx -T -b html -d docs/_build/doctrees -D language=en docs $READTHEDOCS_OUTPUT/html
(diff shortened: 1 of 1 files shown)


foarsitter (changes requested):
This was the flow I expected in the first place but didn't know how to install the optional dependencies and the existence of --extras. 

I would expect the --no-dev and --frozen parameters here. Dev requirements are not required and the lock file should not change.

browniebroke:
> I would expect the --no-dev and --frozen parameters here

Good point! Fixed

foarsitter (approved):

foarsitter:
Forgot that we don't have dev dependencies xD 

Will add some soon ;)

PR #5455, opened 2024-10-11
Update Nginx  dependabot versioning-strategy

luzfcb:
## Description

- Update Nginx  dependabot versioning-strategy as a way to fix #5345 and #5344

Related docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

Checklist:

## Rationale

For some reason, Dependabot is not telling us when there are updates to the Nginx container (and maybe some others). This pull-request is intended to try to do some fine-tuning to fix this issue.

- Helps prevent #5344 from happening again

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -100,6 +100,7 @@ updates:
     directory: "{{cookiecutter.project_slug}}/compose/production/nginx/"
     schedule:
       interval: "daily"
+    versioning-strategy: increase
     labels:
       - "update"
 


luzfcb:
It is also a way to check if Github Dev Team internally solves the https://github.com/dependabot/dependabot-core/issues/10525#issuecomment-2362343084

luzfcb:
The [bug in Dependabot ](https://github.com/dependabot/dependabot-core/issues/10525#issuecomment-2362343084)still exists

https://github.com/cookiecutter/cookiecutter-django/runs/31435188434

PR #5456, opened 2024-10-11
Pin Python version to 3.12

browniebroke:
## Description

Looking at https://github.com/cookiecutter/cookiecutter-django/pull/5450 I noticed that Python 3.13 classifier was added, which promopted me to look at which version of Python was used to run the tests, and it turns out they run using Python 3.13: 

- https://github.com/cookiecutter/cookiecutter-django/actions/runs/11285345637/job/31387993157?pr=5450#step:5:8
- https://github.com/cookiecutter/cookiecutter-django/actions/runs/11285345637/job/31387992848?pr=5450#step:5:9

Checklist:

## Rationale

Run tests on the same Python version as we officially support

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
       - name: Install uv
         uses: astral-sh/setup-uv@v3
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,9 +98,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
       - name: Install uv
         uses: astral-sh/setup-uv@v3
         with:
--- a/.github/workflows/dependabot-uv-lock.yml
+++ b/.github/workflows/dependabot-uv-lock.yml
@@ -11,6 +11,7 @@ permissions:
 
 jobs:
   lock:
+    if: github.repository_owner == 'cookiecutter'
     runs-on: ubuntu-latest
     env:
       GH_PAT: ${{ secrets.GH_PAT }}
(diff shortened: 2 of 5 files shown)


browniebroke:
I also noticed that our bare metal tests run on Python 3.10 since we migrated the template to uv: https://github.com/cookiecutter/cookiecutter-django/actions/runs/11194437208/job/31120912239?pr=5356#step:7:609

Previous run was on Python 3.12 (as expected): https://github.com/cookiecutter/cookiecutter-django/actions/runs/11193688349/job/31119292907?pr=5430#step:7:538

foarsitter:
So the project itself (not the template) is compatible with 3.13?

browniebroke:
No, the template tests were running on 3.13, we don't run any tests in the project itself on that job, just linters. It seems that the project tests were downgraded to python 3.10 in the bare metal tests.

.github/workflows/ci.yml
```diff
@@ -98,6 +98,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
foarsitter, line 103: Is this required? Capping python in pyproject.toml should be sufficient was my expectation.

browniebroke: It doesn't seem to be respecting it because the generated project is not using uv yet. It's for the `test_bare.sh` script

browniebroke: I'll removing it to confirm

browniebroke: The pytest run uses Python 3.10 here `platform linux -- Python 3.10.12`:

https://github.com/cookiecutter/cookiecutter-django/actions/runs/11306934810/job/31448084312?pr=5456#step:7:561

This is because these lines use the first `pip` and `pytest` on the PATH, which is not the one from uv:

https://github.com/cookiecutter/cookiecutter-django/blob/822c1c0327d2025ccde10ec1085f0ecdd7df184a/tests/test_bare.sh#L21

https://github.com/cookiecutter/cookiecutter-django/blob/822c1c0327d2025ccde10ec1085f0ecdd7df184a/tests/test_bare.sh#L24

foarsitter (approved):
Ah yes, you are right. The problem is clear to me now.

browniebroke:
I tried another approach, runing all commands from the script via uv, but that doesn't seem to work. I'll revert back to using actions/setup-python as it seems more straightforward

PR #5493, opened 2024-10-25
Replace docs extra by docs dependency group from PEP 735

browniebroke:
## Description

Leverage [PEP 735](https://peps.python.org/pep-0735/) and migrate docs deps to dependency groups with uv.

Checklist:

## Rationale

New standard more appropriate for this use.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,12 +48,6 @@ dependencies = [
   "tox==4.21.2",
   "tox-uv>=1.11.2",
 ]
-optional-dependencies.docs = [
-  "myst-parser>=4",
-  "sphinx>=8.0.2",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -89,6 +83,14 @@ norecursedirs = [
 
 # ==== djLint ====
 
+[dependency-groups]
```
browniebroke, line 86: I would expect this to be sorted higher in the file, I guess support hasn't landed yet in `pyproject-fmt`

browniebroke: Gave this a try https://github.com/tox-dev/toml-fmt/pull/16

foarsitter:
My compliments for you upstream fix @browniebroke!

PR #5563, opened 2024-11-29
Fix logging configuration for Celery tasks

### LINKED ISSUE #5616: Add CELERY_WORKER_HIJACK_ROOT_LOGGER = False to default settings
## Description

When using Celery with Django, the default behavior is that Celery hijacks the root logger, which means any custom Django logging configurations are not picked up by Celery tasks. This creates an unexpected behavior where logging configurations don't work consistently across the application.

## Proposed Change
Add the following setting to the default configuration:
```python
# https://docs.celeryq.dev/en/latest/userguide/configuration.html#worker-hijack-root-logger
CELERY_WORKER_HIJACK_ROOT_LOGGER = False

pulse-mind:
## Description

The goal is to have logs used in the class of the application when the class is used in a Celery Task. 
The logs were not available on production using supervisor and stdout.

And because logs are so important.

See #5559 

Checklist:

## Rationale

- Fix #5559 
- Fix #5616

### DIFF
--- a/{{cookiecutter.project_slug}}/config/celery_app.py
+++ b/{{cookiecutter.project_slug}}/config/celery_app.py
@@ -1,6 +1,7 @@
 import os
 
 from celery import Celery
+from celery.signals import setup_logging
 
 # set the default Django settings module for the 'celery' program.
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")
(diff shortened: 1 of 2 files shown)


browniebroke (approved):

browniebroke:
Thanks!

PR #5615, opened 2024-12-31
Group all API tests under a `tests.api` namespace

browniebroke:
## Description

Group all API-related tests under a package `users.test.api`.

Checklist:

## Rationale

This is better because:

- It mirrors the structure of the code being tested
- Add a namespace which simplify file names
- Avoids mentioning `drf` in the name


foarsitter (approved):
Much better, thanks!

PR #5621, opened 2025-01-03
Add justfile for use with docker

### LINKED ISSUE #3318: Convenient script for rapid development
## Description

This ticket is a suggestion (since I personally follow and I have seen in some non-Django based projects. 

While working on some projects, using Django cookie-cutter, I had to refer to the cookie-cutter documentation for a number of commands (especially when working with docker). Hence I suggest a convenient script to cover all major functionalities with intuitive command names (example: python manage.py showmigrations => ./app showmigrations) and then the end-user can further extend this as needed.

## Rationale

This could assist the developer in rapid development. Not only they would have the short commands (easy to remember) but also they can extend it further by clubbing 2 or more commands together. Example, taking database backup and uploading to AWS:

Initially we have 3 commands:
```
docker-compose -f local.yml exec postgres backup
docker cp "$(docker-compose -f local.yml ps -q postgres)":/backups ./backups
docker-compose -f production.yml run --rm awscli upload
```

Code in our convenience script:

```
backup_db() {
  docker-compose -f local.yml exec postgres backup
  docker cp "$(docker-compose -f local.yml ps -q postgres)":/backups ./backups
}

upload_backup_db_s3() {
  docker-compose -f production.yml run --rm awscli upload
}
```

End-user would get simpler commands:
```
./app backup_db
./app upload_backup_bd_s3
```

earthcomfy:
## Description

Set up [just](https://github.com/casey/just) to simplify calling docker commands.

Checklist:

## Rationale

It helps to run frequent commands quicker. This [PR](https://github.com/cookiecutter/cookiecutter-django/pull/1879) attempted to do this via MakeFile. However, using `just` seems like a better approach, especially since it is a cross-platform tool (can be installed via pip), as mentioned in this [comment](https://github.com/cookiecutter/cookiecutter-django/pull/1879#issuecomment-2454038873) as well.

Closes #1879 
Fix #3318

### DIFF
--- a/docs/2-local-development/developing-locally-docker.rst
+++ b/docs/2-local-development/developing-locally-docker.rst
@@ -254,7 +254,7 @@ We have included a ``justfile`` to simplify the use of frequent Docker commands 
     "Just" itself rather than its subprocesses.
     For more information, see `this GitHub issue <https://github.com/casey/just/issues/2473>`_.
 
-First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>_.
+First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>`_.
 
 Here are the available commands:
 
--- a/docs/2-local-development/developing-locally-docker.rst
+++ b/docs/2-local-development/developing-locally-docker.rst
@@ -246,7 +246,17 @@ The stack comes with a dedicated node service to build the static assets, watch 
 Using Just for Docker Commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-We have included a ``justfile`` to simplify the use of frequent Docker commands for local development. Here are the available commands:
+We have included a ``justfile`` to simplify the use of frequent Docker commands for local development. 
+
+.. warning::
+    Currently, "Just" does not reliably handle signals or forward them to its subprocesses. As a result, 
+    pressing CTRL+C (or sending other signals like SIGTERM, SIGINT, or SIGHUP) may only interrupt 
+    "Just" itself rather than its subprocesses. 
+    For more information, see `this GitHub issue <https://github.com/casey/just/issues/2473>`_.
+
+First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>_.
+
+Here are the available commands:
 
 - ``just build``
   Builds the Python image using the local Docker Compose file.
--- a/docs/2-local-development/developing-locally-docker.rst
+++ b/docs/2-local-development/developing-locally-docker.rst
@@ -246,12 +246,12 @@ The stack comes with a dedicated node service to build the static assets, watch 
 Using Just for Docker Commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-We have included a ``justfile`` to simplify the use of frequent Docker commands for local development. 
+We have included a ``justfile`` to simplify the use of frequent Docker commands for local development.
 
 .. warning::
-    Currently, "Just" does not reliably handle signals or forward them to its subprocesses. As a result, 
-    pressing CTRL+C (or sending other signals like SIGTERM, SIGINT, or SIGHUP) may only interrupt 
-    "Just" itself rather than its subprocesses. 
+    Currently, "Just" does not reliably handle signals or forward them to its subprocesses. As a result,
+    pressing CTRL+C (or sending other signals like SIGTERM, SIGINT, or SIGHUP) may only interrupt
+    "Just" itself rather than its subprocesses.
     For more information, see `this GitHub issue <https://github.com/casey/just/issues/2473>`_.
 
 First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>_.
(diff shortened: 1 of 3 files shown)


luzfcb:
I like just, but it still doesn't handle SIGNALS behavior very well: https://github.com/casey/just/issues/2473

This pull-request seems good enough for development, however, it should be clear that expanding its use to production can in some cases lead to strange behaviors in signal management between just and the application that is executed with the help of just

earthcomfy:
Thanks for the quick reply. I wasn't aware of the bug. I intended this for local development hence it is using the local compose file.

{{cookiecutter.project_slug}}/justfile
```diff
@@ -0,0 +1,33 @@
+export COMPOSE_FILE := "docker-compose.local.yml"
+
+# Default command to list all available commands.
+default:
+    @just --list
+
+# build: Build python image.
+build:
+    @echo "Building python image..."
+    @docker-compose build
```
browniebroke, line 10: [We moved](https://github.com/cookiecutter/cookiecutter-django/pull/4463) to `docker compose` a while back:

```suggestion
    @docker compose build
```

The standalone `docker-compose` started to break as the compose Docker pluging gained adoption: https://github.com/cookiecutter/cookiecutter-django/issues/3980

browniebroke (commented):
Yes, would be a nice addition. 

Could you please add a paragraph about it in [the documentation about local Docker development (under tips & tricks)](https://cookiecutter-django.readthedocs.io/en/latest/2-local-development/developing-locally-docker.html#tips-tricks) or -even better- update all `docker compose ...` commands to their justfile equivalent?

{{cookiecutter.project_slug}}/justfile
earthcomfy: I see. I have updated it to use `docker compose`

earthcomfy:
@browniebroke I have added the relevant docs too. Let me know if there is anything missing. Thanks.

browniebroke:
That looks good from my perspective

@luzfcb it's documented as being for local development in our docs. Are you hapy with that or do you think we should make the limitations more prominent in the file itself?

docs/2-local-development/developing-locally-docker.rst
```diff
@@ -242,6 +242,31 @@ The stack comes with a dedicated node service to build the static assets, watch
 .. _Sass: https://sass-lang.com/
 .. _live reloading: https://browsersync.io
 
+
+Using Just for Docker Commands
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We have included a ``justfile`` to simplify the use of frequent Docker commands for local development. Here are the available commands:
```
luzfcb, line 249:
```suggestion
We have included a ``justfile`` to simplify the use of frequent Docker commands for local development. 

.. warning::
    Currently, "Just" does not reliably handle signals or forward them to its subprocesses. As a result, 
    pressing CTRL+C (or sending other signals like SIGTERM, SIGINT, or SIGHUP) may only interrupt 
    "Just" itself rather than its subprocesses. 
    For more information, see `this GitHub issue <https://github.com/casey/just/issues/2473>`_.

First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>_.

Here are the available commands:
```

{{cookiecutter.project_slug}}/justfile
```diff
@@ -0,0 +1,33 @@
+export COMPOSE_FILE := "docker-compose.local.yml"
+
```
luzfcb, line 2:
```suggestion

## Just does not yet manage signals for subprocesses reliably, which can lead to unexpected behavior.
## Exercise caution before expanding its usage in production environments. 
## For more information, see https://github.com/casey/just/issues/2473.


```

luzfcb (approved):
@earthcomfy @browniebroke

Following the Zen of Python that says that `Explicit is better than implicit`, I would find it useful to include a warning about this and thus perhaps help prevent our users from losing hours of debugging due to a specific behavior of the current version of Just. 

I myself spent a lot of time in the past debugging this and not understanding what was wrong.

docs/2-local-development/developing-locally-docker.rst
```diff
@@ -242,6 +242,41 @@ The stack comes with a dedicated node service to build the static assets, watch
...
+
+We have included a ``justfile`` to simplify the use of frequent Docker commands for local development.
+
+.. warning::
+    Currently, "Just" does not reliably handle signals or forward them to its subprocesses. As a result,
+    pressing CTRL+C (or sending other signals like SIGTERM, SIGINT, or SIGHUP) may only interrupt
+    "Just" itself rather than its subprocesses.
+    For more information, see `this GitHub issue <https://github.com/casey/just/issues/2473>`_.
+
+First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>_.
```
browniebroke, line 257:
```suggestion
First, install Just using one of the methods described in the `official documentation <https://just.systems/man/en/packages.html>`_.
```

image

browniebroke (approved):

PR #5613, opened 2024-12-31
Move template linting and formatting to ruff

browniebroke:
## Description

The generated project already uses that, let's be consistent and use it everywhere. Based on (merge it first):

- [x] https://github.com/cookiecutter/cookiecutter-django/pull/5614
- [x] https://github.com/cookiecutter/cookiecutter-django/pull/5648
- [x] https://github.com/cookiecutter/cookiecutter-django/pull/5649

Checklist:

- [x] I've made sure that tests are updated accordingly (especially if adding or updating a template option)
- [x] I've updated the documentation or confirm that my change doesn't require any updates

PR #5648, opened 2025-01-18
Migrate post-generation hook to pathlib

browniebroke:
## Description

The post-generation hook makes heavy uses of path operations and Ruff suggests to migrate them to `pathlib` in #5613 

These suggestions can't be fixed automatically by Ruff, so in that other PR, I asked Ruff to add a lot of `noqa` (with the `--add-noa` option) where there was issues. However, that makes the diff bigger that it should... 

This PR intend to fix all the `PTH` issues from that script, and that script only.

Checklist:


hooks/post_gen_project.py
```diff
@@ -72,15 +60,15 @@ def remove_docker_files():
         "justfile",
     ]
     for file_name in file_names:
-        os.remove(file_name)
+        Path(file_name).unlink()
     if "{{ cookiecutter.editor }}" == "PyCharm":
         file_names = ["docker_compose_up_django.xml", "docker_compose_up_docs.xml"]
         for file_name in file_names:
-            os.remove(os.path.join(".idea", "runConfigurations", file_name))
+            Path(".idea", "runConfiguration", file_name).unlink()
```
browniebroke, line 67:
```suggestion
            Path(".idea", "runConfigurations", file_name).unlink()
```

foarsitter (approved):
An abstraction for unlink & rmtree would further improve this script but may be out of scope 
```python
def delete_files(*args: Path): 
    for file in args:
        if file.isdir():
            shutil.rmtree(file)
        else:
            file.unlink()

```

browniebroke:
> An abstraction for unlink & rmtree would further improve this script but may be out of scope
> 
> ```python
> def delete_files(*args: Path): 
>     for file in args:
>         if file.isdir():
>             shutil.rmtree(file)
>         else:
>             file.unlink()
> ```

Yes good idea, we do this all over the place 👍🏻

PR #5651, opened 2025-01-21
Update linter error code from TCH to TC

ghazi-git:
## Description

Update the error code from TCH to TC since ruff has done that in v0.8, here's an extract from the blog post (and [link for reference](https://astral.sh/blog/ruff-v0.8.0#new-error-codes-for-flake8-type-checking-rules))

> This release changes the error codes for the rules in our [flake8-type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) category from TCH to TC. This matches changes that were made in the original flake8-type-checking plugin from which these rules were originally adapted.

Checklist:

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -106,7 +106,7 @@ select = [
   "SLOT",
   "SIM",
   "TID",
-  "TCH",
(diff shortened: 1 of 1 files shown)


browniebroke (approved):

browniebroke:
Thanks!

PR #5652, opened 2025-01-22
Migrate generated project from `runtime.txt` to `.python-version`

edmorley:
Hi! I'm the Heroku Python owner :-)

Heroku now supports the `.python-version` file as an alternative to the Heroku-specific (and now deprecated) `runtime.txt` file, and we recommend apps switch to using the former since it's more widely supported by other tooling.

The `.python-version` file is supported by pyenv, uv, the GitHub Actions `setup-python` action and many others.

In addition, the `.python-version` file supports the version being specified as just the major version (eg `3.12` vs `3.12.N`) which means will pick up Python patch updates automatically (which is the recommended usage on Heroku).

See:
https://devcenter.heroku.com/changelog-items/3005
https://devcenter.heroku.com/articles/python-runtimes#selecting-a-runtime
https://github.com/heroku/heroku-buildpack-python/issues/1642
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input

---

Checklist:


luzfcb (approved):
LGTM.

@edmorley Since you're likely more familiar with the recent changes at Heroku, does the cookiecutter-django Heroku deployment tutorial require any updates to align with current Heroku best practices?

https://github.com/cookiecutter/cookiecutter-django/blob/master/docs/3-deployment/deployment-on-heroku.rst

edmorley:
@luzfcb Thank you for the review!



> Since you're likely more familiar with the recent changes at Heroku, does the cookiecutter-django Heroku deployment tutorial require any updates to align with current Heroku best practices?

Ah good question! Glancing through it, it seems mostly fine - though I spotted a couple of small tweaks that would be worth making - for which I've opened #5656.

browniebroke (approved):

browniebroke:
Thanks!

PR #5656, opened 2025-01-26
Update Heroku deployment guide

edmorley:
Hi! I'm the Heroku Python owner :-)

This:
* Removes setting of `PYTHONHASHSEED` env var, since it's defaulted to `random` since Python 3.3. See:
  https://docs.python.org/3.13/whatsnew/3.3.html#builtin-functions-and-types
* Removes setting of `WEB_CONCURRENCY` env var to a hardcoded value, since on Heroku the Python buildpack dynamically sets the value at container boot based on the available CPUs, so it's best to not override the default with a value that might not match the size of the dyno. See:
  https://devcenter.heroku.com/articles/python-concurrency#default-settings-and-behavior


browniebroke (approved):

kennethreitz:
:sparkles: :cake: :sparkles:

PR #5698, opened 2025-02-22
Group dependabot docker directories

### LINKED ISSUE #5699: Group dependabot updates
## Description

To reduce the number of pull requests Dependabot creates they can be organized in [groups](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--). I propose one group each for github actions, docker, and npm, and separate production and development groups for pip. I can implement the change after #5698 is finished.

## Rationale

Dependabot opens too many pull requests, especially for people making solo projects. This makes keeping track of updates much simpler.

mfosterw:
## Description

Dependabot supports a "directories" keyword now so the docker dependencies can be grouped together now.

Checklist:

## Rationale

It saves a lot of space. Also, the issue dependabot/dependabot-core#2178 has been fixed, so if you're ok pinning all the dependencies to their minor version they can all be captured by one glob.

Fix #5699

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,11 +36,11 @@ updates:
       - "update"
 
   # Enable version updates for Docker
-  # We need to specify each Dockerfile in a separate entry because Dependabot doesn't
-  # support wildcards or recursively checking subdirectories. Check this issue for updates:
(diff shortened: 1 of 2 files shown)


browniebroke (approved):
That's amazing! I didn't realise this was added - thanks!

PR #5700, opened 2025-02-22
Add Docker build caching on GitHub Actions

### LINKED ISSUE #2588: Add cache for Docker-Based Projects in CI

aidoskanapyanov:
## Description

Checklist:

## Rationale

First of all, appreciate the work you've put into this project! 🔥

Implemented caching for docker builds in github actions CI.

Helps tremendously reduce the waiting time for tests via taking advantage of github actions cache backend in docker builds.
Reduces the build times from **~3 minutes** to **~30 seconds**, because of cached steps like pip install requirements/*.txt, apt install/update etc

Fix #2588

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/ci.yml
@@ -68,11 +68,31 @@ jobs:
         uses: actions/checkout@v4
       {%- if cookiecutter.use_docker == 'y' %}
 
-      - name: Build the Stack
-        run: docker compose -f docker-compose.local.yml build django
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
 
-      - name: Build the docs
-        run: docker compose -f docker-compose.docs.yml build docs
+      - name: Build and cache local backend
+        uses: docker/bake-action@v6
+        with:
+          push: false
+          load: true
+          files: docker-compose.local.yml
+          targets: django
+          set: |
+            django.cache-from=type=gha,scope=django-cached-tests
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/.github/workflows/ci.yml
```diff
@@ -68,11 +68,31 @@ jobs:
...
 
-      - name: Build the Stack
-        run: docker compose -f docker-compose.local.yml build django
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
 
-      - name: Build the docs
-        run: docker compose -f docker-compose.docs.yml build docs
+      - name: Build and cache local backend
+        uses: docker/bake-action@v6
```
browniebroke, line 75: I didn't know about [docker bake](https://docs.docker.com/build/bake/), TIL!

browniebroke (approved):
That's really cool, speeds things up masssively. Thank you!

PR #5702, opened 2025-02-23
Update local.txt

sebastian-code:
## Description

Fixes typing error, where the reference to the base file was switched by error, pointing to the production.txt file.

Checklist:

## Rationale
The requirements file structure.


browniebroke (changes requested):
This was changed very intentionally in https://github.com/cookiecutter/cookiecutter-django/pull/4838 

Maybe I'm missing something... What makes you think that it was an "error"?

sebastian-code:
Checking #4838, the rationale behind the change makes total sense. This PR is more than unnecessary.

PR #5709, opened 2025-02-28
Add groups for dependabot updates

### LINKED ISSUE #5699: Group dependabot updates

mfosterw:
## Description

Dependabot doesn't have to create new pull requests for each update. Instead, it can group the updates for a particular ecosystem into one pull request.

Checklist:

## Rationale

Update groups makes it much easier to stay on top of dependabot updates and also prevents them from clogging up your commit history. 

Fixes #5699

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/dependabot.yml
+++ b/{{cookiecutter.project_slug}}/.github/dependabot.yml
@@ -51,10 +51,6 @@ updates:
     # Every weekday
     schedule:
       interval: 'daily'
-    groups:
-      docker:
-        patterns:
-          - '*'
 
 {%- endif %}
 
@@ -68,8 +64,9 @@ updates:
       interval: 'daily'
     groups:
       python:
-        patterns:
-          - '*'
+        update-types:
+          - 'minor'
+          - 'patch'
 
 {%- if cookiecutter.frontend_pipeline == 'Gulp' %}
 
@@ -82,7 +79,8 @@ updates:
       interval: 'daily'
     groups:
       javascript:
-        patterns:
-          - '*'
+        update-types:
+          - 'minor'
+          - 'patch'
 
 {%- endif %}
--- a/{{cookiecutter.project_slug}}/.github/dependabot.yml
+++ b/{{cookiecutter.project_slug}}/.github/dependabot.yml
@@ -9,6 +9,10 @@ updates:
     # Every weekday
     schedule:
       interval: 'daily'
+    groups:
+      github-actions:
+        patterns:
+          - '*'
 
 {%- if cookiecutter.use_docker == 'y' %}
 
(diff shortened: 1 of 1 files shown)


mfosterw:
I wanted to create separate groups for development and production Python dependencies, but according to [stackoverflow](https://stackoverflow.com/a/76133576) any Python requirement in a `.txt` file is considered a production dependency.

{{cookiecutter.project_slug}}/.github/dependabot.yml
```diff
@@ -43,6 +51,10 @@ updates:
     # Every weekday
     schedule:
       interval: 'daily'
+    groups:
+      docker:
+        patterns:
+          - '*'
```
browniebroke, line 57: I'm not sure about this one, each docker image in this group are quite unrelated so I would personally prefer to keep them separated. They are not being updated very frequently, so grouping them isn't solving much IMO.

I think that could potentially bundle 2 updates that I would like to split: e.g a major node update with a patch traefik update... Major node would require more attention or work.

{{cookiecutter.project_slug}}/.github/dependabot.yml
```diff
@@ -9,6 +9,10 @@ updates:
     # Every weekday
     schedule:
       interval: 'daily'
+    groups:
+      github-actions:
+        patterns:
+          - '*'
```
browniebroke, line 15: Nice!

browniebroke (commented):
Thanks. I agree it makes total sense for Github actions and python docker images, but I'm a bit more on the fence for PyPI and NPM packages...

I'm slightly worried about cases where there is 1 package in the bundled update that has a problem and break the build, but maybe that's not all that common, and folks can easily remove the group?

I don't know whether it's possible, but can we avoid grouping major package updates?

mfosterw:
I made the changes you recommended, let me know if you think there's anything else.

> I'm slightly worried about cases where there is 1 package in the bundled update that has a problem and break the build, but maybe that's not all that common, and folks can easily remove the group?

I have experienced this already with `djangorestframework-stubs` and `mypy` (I think just because drf-stubs hadn't made a release with updated minimum dependency versions, not an actual incompatibility) but since `mypy` was already explicitly included by the `[compatible-mypy]` modifier on drf-stubs and django-stubs, I felt ok removing the specific version for it. I don't think I've seen any other dependabot updates other than mypy fail my build, but I understand it might be more common for larger projects, and I agree it's reasonable to only group the minor updates.

browniebroke (approved):

PR #5717, opened 2025-03-05
Fix Celery CPendingDeprecationWarning

### LINKED ISSUE #5085: CPendingDeprecationWarning: The broker_connection_retry configuration setting will no longer determine whether broker connection retries are made during startup in Celery 6.0 and above
## Description

I found this deprecation warning for the celery;

```
CPendingDeprecationWarning: The broker_connection_retry configuration setting will no longer
determine whether broker connection retries are made during startup in Celery 6.0 and above.
If you wish to retain the existing behavior for retrying connections on startup,
you should set broker_connection_retry_on_startup to True.
```

https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-connection-retry

As the newest celery almost using that version;

https://github.com/cookiecutter/cookiecutter-django/blob/b8b470c796c2c91325e9f7349497d2dc148262a1/%7B%7Bcookiecutter.project_slug%7D%7D/requirements/base.txt#L18-L20

## Solution

And found the solution at https://stackoverflow.com/a/78088996/6396981

Where we can utilize the following settings to silence the warning by putting the below code into `celery.py`.

```python
celery_app.conf.broker_connection_retry_on_startup = True
```

or `settings.py`;

```python
CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
```

Just wondering to add this into cookie-cutter-django

MelvinvanMelzen:
## Description
Added the following setting with comment to base.py

```python
https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-connection-retry-on-startup
CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
```

Checklist:

## Rationale
This is the default value for this setting anyway and stops the deprecation warning per the example below.

```
2025-03-05 08:56:48 WARNING 2025-03-05 08:56:48,021 warnings 22 280781127086112 /usr/local/lib/python3.12/site-packages/celery/worker/consumer/consumer.py:508: CPendingDeprecationWarning: The broker_connection_retry configuration setting will no longer determine
2025-03-05 08:56:48 whether broker connection retries are made during startup in Celery 6.0 and above.
2025-03-05 08:56:48 If you wish to retain the existing behavior for retrying connections on startup,
2025-03-05 08:56:48 you should set broker_connection_retry_on_startup to True.
```

Fix #5085

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/base.py
+++ b/{{cookiecutter.project_slug}}/config/settings/base.py
@@ -329,6 +329,8 @@ CELERY_WORKER_SEND_TASK_EVENTS = True
 CELERY_TASK_SEND_SENT_EVENT = True
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#worker-hijack-root-logger
 CELERY_WORKER_HIJACK_ROOT_LOGGER = False
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-connection-retry-on-startup
+CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
 
 {%- endif %}
 # django-allauth


{{cookiecutter.project_slug}}/config/settings/base.py
```diff
@@ -329,6 +329,8 @@
 CELERY_TASK_SEND_SENT_EVENT = True
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#worker-hijack-root-logger
 CELERY_WORKER_HIJACK_ROOT_LOGGER = False
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-connection-retry-on-startup
+CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
```
browniebroke, line 333: Can we move this closer to the others `CELERY_BROKER_...` settings please and set it to `False`:

https://github.com/cookiecutter/cookiecutter-django/issues/5085#issuecomment-2211735674

browniebroke (changes requested):

browniebroke:
> It is basically True right now cause it's the default and setting it to False would be a potentially breaking change

Hum... I see. So basically the decision we took in https://github.com/cookiecutter/cookiecutter-django/issues/5085 was the right one? 

We're not setting `broker_connection_retry` either so we're getting the default value from Celery (enabled) and when v6.0 comes out, we'll keep getting the same default value of enabled, but from a different setting.

MelvinvanMelzen:
I've checked the Celery source, this is the first place where both settings are used. 

    def _get_connection_retry_type(self, is_connection_loss_on_startup):
        return ('broker_connection_retry_on_startup'
                if (is_connection_loss_on_startup
                    and self.app.conf.broker_connection_retry_on_startup is not None)
                else 'broker_connection_retry')

The deprecated retry setting has a default of True, the new retry setting has a default of None. So implicitly it is currently True in the above implementation. Explicitly defining it to True in the Django settings would be preferential to me but actually this seems like a code smell of Celery which triggers the depreciation warning which is a good thing for projects where the old setting is used but the new one isn't. 

Going back to the discussion from last year. The Celery worker will still exit after exhausting the (default) retries. Setting it to False will make it fail faster which is definitely easier to debug but I would say a bit more fragile.

browniebroke:
Unless I'm missing something, the `broker_connection_retry` is not deprecated, as far as I understand, it (will) control retries in a different context, in normal operations. It seems like they had a single setting which they split into 2 to enable more granular control of retries: on start-up vs after start-up, during normal operations

browniebroke:
Thanks for the suggestion but we won't be implementing this at this time. The default are good for us and can stay as they are. The warning from Celery is arguabely a bit misleading, but it's not worth a fix in this project.

PR #5754, opened 2025-04-02
Fix line endings for dotenv merge

igor-wl:
## Description

Replaced `os.linesep` with `\n` to fix line endings for dotenv merge.

Checklist:

## Rationale

`read_text` converts all line ending into `\n` and `write_text` converts `\n` into system dependent line endings, any contents update in between needs to use only `\n`.
Whats currently happening on Windows is that `os.linesep` contains `\r\n` which gets converted by `write_text` into two new lines and out of the box tests are failing.


browniebroke (approved):
> `read_text` converts all line ending into `\n` and `write_text` converts `\n` into system dependent line endings

That's really cool, I didn't know that.

PR #5831, opened 2025-05-08
Fix media folder permissions prod dockerfile

### LINKED ISSUE #5728: Improve media folder permissions in production Dockerfile
## Description

I noticed a small issue with the current permissions setup for media files in the django production Dockerfile.

In PR #4889, the following change was made:

`RUN chown -R django:django ${APP_HOME}`

While this helps make the entire app directory owned by the Django user, I encountered a case where media files were still not writable in production when using the MEDIA_ROOT directory served by Nginx.

To fix this, I had to explicitly create the media folder before changing ownership:

`RUN mkdir -p ${APP_HOME}/{{ cookiecutter.project_slug }}/media`

With this, I ensure that the media directory exists when the chown command is executed, avoiding the automatic creation of the directory, which would result in the media folder being owned by the root user of the container.

## Rationale

By explicitly creating the media directory, we ensure fully accessible resources. This prevents potential issues with broader ownership settings affecting media files specifically and ensures consistent configuration across environments.

dominiqueplante:
## Description

Update the production dockerfile to improve media folder permissions in production

Checklist:

- [x] I've updated the documentation or confirmed that my change doesn't require any updates

## Rationale

Fix #5728

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
@@ -119,6 +119,8 @@ COPY --from=client-builder --chown=django:django ${APP_HOME} ${APP_HOME}
 COPY --chown=django:django . ${APP_HOME}
 {%- endif %}
 
+RUN mkdir -p ${APP_HOME}/{{ cookiecutter.project_slug }}/media
+
 # make django owner of the WORKDIR directory as well.
 RUN chown -R django:django ${APP_HOME}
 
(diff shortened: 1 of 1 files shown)


browniebroke (commented):
I think this should be conditional to having media served by nginx, so should replicate the condition from there:

https://github.com/cookiecutter/cookiecutter-django/blob/c534821b9a78867afedade676e5b751241537f72/%7B%7Bcookiecutter.project_slug%7D%7D/docker-compose.production.yml#L107-L118

dominiqueplante:
Thank you, @browniebroke, for the quick review and helpful feedback. I suspected a conditional was needed, but wasn’t sure of the correct approach—appreciate your guidance.

browniebroke (approved):

PR #5845, opened 2025-05-13
Remove redundant `use_celery` condition in local compose file

### LINKED ISSUE #5843: Potentially redundant use_celery check in docker-compose.local
It looks like the inner check for use_celery can be removed, since this code would be only be entered if same outer check succeeded - or am I missing something?

https://github.com/cookiecutter/cookiecutter-django/blob/b82478e20ff83a3175d471f4ae3f16366059755f/{{cookiecutter.project_slug}}/docker-compose.local.yml#L51-L56

dominiqueplante:
## Description

Remove redundant conditional (same condition is being checked a few lines above)
Checklist:

## Rationale

Simplifies the code and improves readability

Fix #5843


browniebroke (approved):

browniebroke:
Thanks!

PR #5929, opened 2025-07-13
Fix howto docker command

lucaskbr:
## Description

Fix the howto docker command that is pointing to a non existing file

Checklist:

## Rationale

Be able to run docs without having to check if files exists or not


luzfcb:
@lucaskbr good catch. Thanks

PR #6052, opened 2025-09-11
Enable collectfasta for Azure storage

### LINKED ISSUE #5934: Enable collectfasta for Azure blob storage
## Description

Collectfasta was disabled for Azure storage as unsupported in this PR:
https://github.com/cookiecutter/cookiecutter-django/pull/5476

It is now supported as of [Collectfasta 3.3.0](https://github.com/jasongi/collectfasta/releases/tag/3.3.0) (PR: https://github.com/jasongi/collectfasta/pull/7), so it can be re-enabled by undoing the changes in the above PR by adding Azure to the cloud_provider check i.e
```
{%- if cookiecutter.use_whitenoise == 'n' and cookiecutter.cloud_provider in ('AWS', 'GCP', 'Azure') -%}
```
adding the following setting
```
COLLECTFASTA_STRATEGY = "collectfasta.strategies.azure.AzureBlobStrategy"
```

## Rationale

Faster asset upload when using Collectfasta

codenamenam:
## Description

Checklist:

## Rationale

Fix #5934 

This PR re-enables `collectfasta` for Azure blob storage.

Support for Azure was added in `collectfasta==3.3.0`, so the original reason for disabling it (in #5476) is no longer valid.

### Changes

- Updated `requirements/production.txt` to include `Azure` in the `collectfasta` installation condition.
- Updated `config/settings/production.py` to add the `AzureBlobStrategy` when Azure is selected as the cloud provider.

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -205,6 +205,7 @@ STATIC_URL = f"https://storage.googleapis.com/{GS_BUCKET_NAME}/static/"
 {%- elif cookiecutter.cloud_provider == 'Azure' %}
 MEDIA_URL = f"https://{AZURE_ACCOUNT_NAME}.blob.core.windows.net/media/"
 {%- if cookiecutter.use_whitenoise == 'n' %}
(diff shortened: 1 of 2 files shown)


browniebroke:
Thanks 🎉

PR #6074, opened 2025-09-23
Add pyproject-fmt to the generated project

browniebroke:
## Description

We use that in the template and it's pretty good at putting the project metadata first, as opposed to at the bottom like where we currently put them.

Format the templated `pyproject.toml` file to minimise the changes. 

~Unfortuntaly, we can't completely eliminate that at the moment due to [uv using a 4 spaces indentation](https://github.com/astral-sh/uv/issues/5852), but it looks like that's going to be fixed at some point.~ EDIT: turns out uv tries to follow the pre-existing indentation in `dependencies` and `dependency-groups` so we can tell it what we want by adding at least one dependancy there. I chose `django` and `pytest` for each as they they seem the most unlikely to go away for now.

Checklist:

## Rationale

Consistent formatting

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
       - id: detect-private-key
 
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v4.0.0-alpha.8"
+    rev: v4.0.0-alpha.8
     hooks:
       - id: prettier
         args: ["--tab-width", "2"]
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -44,6 +44,11 @@ repos:
       # Formatter
       - id: ruff-format
 
+  - repo: https://github.com/tox-dev/pyproject-fmt
+    rev: v2.6.0
```
browniebroke, line 48:
```suggestion
    rev: v2.7.0
```

browniebroke:
Let's try to get this in and see how it goes

PR #6097, opened 2025-10-06
fix(project generation fails with docker --load flag)

### LINKED ISSUE #6094: Project generation fails with docker/podman
After filling the questionnaire when creating a new project from the template, cookiecutter stops the following error message:

```
Installing python dependencies using uv...
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

Unable to find image 'cookiecutter-django-uv-runner:latest' locally
docker: Error response from daemon: {"message":"denied: requested access to the resource is denied"}

Run 'docker run --help' for more information
Error installing production dependencies: Command '['docker', 'run', '--rm', '-v', '[REDACTED PATH]:/app', 'cookiecutter-django-uv-runner:latest', 'uv', 'add', '--no-sync', '-r', 'requirements/production.txt']' returned non-zero exit status 125.
ERROR: Stopping generation because post_gen_project hook script didn't exit successfully
Hook script failed (exit status: 1)
```

This looks like it is caused by `docker build` in hooks/post_gen_project.py. The parameters in line 521-531 should include "--load" (or "--output=type=docker", apparently). Adding the parameter leads to a successful run.

cookiecutter-django: latest commit + change
podman: 5.6.2
docker (cli): 28.5.0

Tosinibikunle:
fixes #6094

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -522,6 +522,7 @@ def setup_dependencies():
                 [  # noqa: S607
                     "docker",
                     "build",
+                    "--load",
                     "-t",
                     uv_image_tag,
                     "-f",


browniebroke:
Thanks!

PR #6116, opened 2025-10-13
Enable version updates for Docker Compose files

igor-wl:
## Description

Added missing docker-compose config for dependabot updates.

Checklist:

## Rationale

So that images defined in compose files (i.e. `redis:6` is EOL) do not become obsolete without noticing.

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/dependabot.yml
+++ b/{{cookiecutter.project_slug}}/.github/dependabot.yml
@@ -52,6 +52,14 @@ updates:
     schedule:
       interval: 'daily'
 
+  # Enable version updates for Docker Compose files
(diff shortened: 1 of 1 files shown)


browniebroke:
Could you also update the config for the generated project, please?

https://github.com/cookiecutter/cookiecutter-django/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D%2F.github%2Fdependabot.yml

igor-wl:
> Could you also update the config for the generated project, please?
> 
> https://github.com/cookiecutter/cookiecutter-django/blob/main/%7B%7Bcookiecutter.project_slug%7D%7D%2F.github%2Fdependabot.yml

Done

Issue #6121, opened 2025-10-14
Replace Redis by Valkey


browniebroke:
## Description

As discussed in https://github.com/cookiecutter/cookiecutter-django/discussions/6115, we use Redis for caching/broker however we were running on an unsupported version. As the newer versions have a more complex license, it would be safer for us to migrate to the fully open source fork, Valkey.

## Rationale

Simpler license.

PR #6152, opened 2025-10-30
migration to valkey from redis

p-r-a-v-i-n:
This PR addresses #6121 
I would really appreciate suggestions on my approach.


igor-wl:
My thoughts about what else should be done:

- Replace `redis-py` with [valkey-py](https://github.com/valkey-io/valkey-py) in the base requirements
- Replace `django-redis` with [django-valkey](https://github.com/django-commons/django-valkey) - they forked `django-redis` but 0.x.x version suggests they are not very confident :-/
- Look into the progress for adding valkey as a celery backend - last time I checked they were working on it but had some unresolved issue
- Also I think using `valkey:latest` might not be the best choice and could cause problems, I would rather pin that to a specific version

Libraries are all interchangeable with their redis counterparts for now but my understanding is that valkey 8 has already diverged from redis and obviously it will continue to diverge more in the future so we should probably switch.

p-r-a-v-i-n:
HI @igor-wl , really helpful .
Honestly ,  After this issue only I got to know about valkey and i'm still learning it  and this Cookiecutter Django setup, so I haven't explored the entire codebase yet.
I was under the impression that redis and valkey is just url difference and I made the initial migration focusing on the direct Redis to Valkey switch, but after your feedback , i can see there is much more deeper area to update ,which makes sense.
I would like to continue and complete this properly. Could you guide me on the preferred approach for:

>  I would rather pin that to a specific version

Noted. any recommendation

> celery backend

should we switch immediately ?

> switch django-redis to django-valkey

is it ok for now ?

Once I get clarity, I'll update the PR accordingly.

Thank you again.

igor-wl:
> > I would rather pin that to a specific version
> 
> Noted. any recommendation

We can put `9`.


> > celery backend
> 
> should we switch immediately ?

It seems that valkey backend is still in progress so I guess leave that part as is for now. Not sure if this is going to be a blocker...

PR #6159, opened 2025-11-02
local py packages

professorhaseeb:
## Description

Checklist:

## Rationale

uv source pages dont exist before copying the code


professorhaseeb:
recreate

```toml
# uv add ./pkg
[tool.uv.workspace]
members = [
    "pkg",
]

[tool.uv.sources]
pkg = { workspace = true }
```

browniebroke:
I'll need more context please. What is this solving? Why do we need this? Can you link to some relevant pices of documentation where I can learn more?

professorhaseeb:
hi, i'm not sure if this is only my use case, but when a local package is used for development
`uv add ./local-pkg`

uv updates pyproject.toml's sources and workspaces.
when local docker compose is ran. it crashes, 

`--no-sources` before copy runs it fine

https://docs.astral.sh/uv/concepts/projects/workspaces/#workspace-sources

PR #6175, opened 2025-11-10
Fix virtual environment setup to GitLab CI

professorhaseeb:
## Description

Checklist:

### DIFF
--- a/{{cookiecutter.project_slug}}/.gitlab-ci.yml
+++ b/{{cookiecutter.project_slug}}/.gitlab-ci.yml
@@ -20,6 +20,7 @@ precommit:
     paths:
       - ${PRE_COMMIT_HOME}
   before_script:
+    - uv venv
     - uv pip install -q pre-commit pre-commit-uv
   script:
     - uv run pre-commit run --show-diff-on-failure --color=always --all-files


browniebroke (commented):
Why do we need that on CI? AFAIK, the CI worker is only used to run this one installation and then pre-commit creates a venv for each hook so I don't see much benefit here... 

Thinking about it, perhaps it would be more beneficial to install pre-commit as tool?

msamoylov:
> Why do we need that on CI? AFAIK, the CI worker is only used to run this one installation and then pre-commit creates a venv for each hook so I don't see much benefit here...

@browniebroke I'm upvoting this one. Actually, the pipeline is failing without that step. See:

```
$ uv pip install -q pre-commit pre-commit-uv
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

browniebroke:
> Actually, the pipeline is failing without that step. See:
> 
> ```
> $ uv pip install -q pre-commit pre-commit-uv
> error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
> ```

Ok thanks, that's pretty much what I wanted to know 👍🏻

browniebroke:
> perhaps it would be more beneficial to install pre-commit as tool?

What about this?

```yaml
before_script:
  - uv tool install pre-commit --with pre-commit-uv
script: 
  - pre-commit run --show-diff-on-failure --color=always --all-files
```

or just

```yaml
script: 
  - uv -q tool run --with pre-commit-uv pre-commit run --show-diff-on-failure --color=always --all-files
```

PR #6183, opened 2025-11-16
Set default for SENTRY_DSN in production settings

professorhaseeb:
## Description

Checklist:

## Rationale

when sentry dsn is removed from env, production gives erro
solution: set a default on sentry env

### DIFF
--- a/{{cookiecutter.project_slug}}/config/settings/production.py
+++ b/{{cookiecutter.project_slug}}/config/settings/production.py
@@ -410,7 +410,7 @@ LOGGING = {
 
 # Sentry
 # ------------------------------------------------------------------------------
-SENTRY_DSN = env("SENTRY_DSN")
+SENTRY_DSN = env("SENTRY_DSN", default=None)
 SENTRY_LOG_LEVEL = env.int("DJANGO_SENTRY_LOG_LEVEL", logging.INFO)
 
 sentry_logging = LoggingIntegration(


browniebroke (dismissed):

browniebroke:
I see where you're coming from, however if I answer "use sentry=y", I would rather have the app crash loudly if I forget to set the env variable that silently not send crash reports where I'm expecting them (in Sentry).

professorhaseeb:
In my case i created the project with use sentry=y, but i never added the dsn url, and it never gave any error

browniebroke:
I'm confused... In the PR description, you say the opposite:

> when sentry dsn is removed from env, production gives erro

I think this is the behaviour:

- Without default, an error is raised to flag explicitly the missing value from env, which - makes the problem clear. This is how it works on main and my preference
- With `default=None`, if the env variable for the DSN is misconfigured (e.g typo in the name), the problem is silently ignored, and Sentry won't work. It also enables you to skip Sentry configuration, as you did.

Is my understanding correct? Let me know if I'm missing something

professorhaseeb:
what happened in my scenario is i had the project setup as sentry=y, and i never added the variable for sentry dsn in production.

i then later added the sentry dsn and deployed production again. everything was fine,
later i removed sentry dns and deployed production again, and i got the error. which when i set `default=None` it worked

professorhaseeb:
>Without default, an error is raised to flag explicitly the missing value from env, which - makes the problem clear. This is how it works on main and my preference

without default it never gave me any error.

after setting dsn variable - no error
removing dns variable - error
adding default - no error

browniebroke:
As I said before, the lack of default is intentional to make it obvious when something is misconfigured.

Thanks for the suggestion

PR #6210, opened 2025-11-30
Integrate Django-RQ as an alternative task queue

### LINKED ISSUE #6209: Feature: Add django-rq as an alternative task queue
## Description

I’d like to add **django‑rq** as an optional alternative to **Celery** in cookiecutter‑django.

This would not replace Celery. Instead, during project generation there would be a choice of task queue backend, for example:

- Celery  
- django‑rq  
- None

If django‑rq is selected, the template would:

- Add `django-rq` (and `rq`) to the dependencies.
- Add `django_rq` to `INSTALLED_APPS` and configure `RQ_QUEUES`, reusing the existing Redis settings when possible.
- Provide a standard worker command, for example `python manage.py rqworker`.
- Include a small example task and usage pattern that shows how to enqueue a job using `django_rq.get_queue("default").enqueue(...)`.
- Optionally expose the django‑rq admin / dashboard if appropriate.

The Celery setup would remain as is, and users could still choose it instead.

## Rationale

Many Django projects need background jobs for tasks like sending emails, processing uploads, or calling external APIs. Celery is very powerful, but it is also quite complex and comes with more moving parts.

**django‑rq** offers:

- A simpler mental model: jobs are just Python callables pushed into Redis, workers pull and run them.
- A natural fit if you already use Redis for caching or channels.
- Fewer components to configure since Redis doubles as both queue and storage for job metadata.
- Built in management and job monitoring via django‑rq’s admin and web views, without needing extra services like Flower or additional Django apps.

For a lot of cookiecutter‑django users, django‑rq is “enough queue” while being easier to understand and maintain than Celery. Making it a first‑class, opt‑in option would give teams a practical choice between:

- Celery for advanced or high scale setups.
- django‑rq for simpler, Redis based workloads.

Existing projects that rely on Celery would not be affected, since this would only apply to new projects that choose django‑rq at generation time.

millsks:
## Description

This PR adds [django-rq](https://github.com/rq/django-rq) as an optional background task backend. Projects can now opt into django-rq as a simpler alternative to Celery when generating a new project. Celery remains fully supported and unchanged.

**What's included:**

- New `use_django_rq` cookiecutter option
- django-rq dependency added to requirements when enabled
- RQ configuration in settings, reusing existing Redis infrastructure
- Docker Compose `rqworker` service and Procfile worker definition
- Documentation covering worker setup, admin dashboard, and metrics endpoints

**Key features:**

- Reuses existing Redis patterns from Celery integration
- Includes admin dashboard, Prometheus metrics, and JSON stats API
- Optional synchronous mode for easier debugging in local/test environments
- Production-ready with systemd and Heroku deployment examples

Checklist:

## Rationale

Many Django projects need a simple Redis-backed job queue without the complexity of Celery. django-rq is production-ready and provides exactly that. Since cookiecutter-django already integrates Redis for Celery, adding django-rq as an option has minimal overhead while giving users a lighter-weight choice for background tasks.

**Open questions for maintainers:**

- Should `use_celery` and `use_django_rq` be mutually exclusive, or allow both for advanced setups?
- Any preference for the option naming or default queue configuration?
- Are there specific dependency management or Docker conventions I should follow more closely?

Resolves #6209

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/local/django/Dockerfile
@@ -76,10 +76,6 @@ RUN chmod +x /start-rqworker
 COPY ./compose/local/django/rq/scheduler/start /start-rqscheduler
 RUN sed -i 's/\r$//g' /start-rqscheduler
 RUN chmod +x /start-rqscheduler
-
-COPY ./compose/local/django/rq/dashboard/start /start-rqdashboard
-RUN sed -i 's/\r$//g' /start-rqdashboard
-RUN chmod +x /start-rqdashboard
 {% endif %}
 
 ENTRYPOINT ["/entrypoint"]
--- a/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
+++ b/{{cookiecutter.project_slug}}/compose/production/django/Dockerfile
@@ -115,11 +115,6 @@ RUN chmod +x /start-rqworker
 COPY --chown=django:django ./compose/production/django/rq/scheduler/start /start-rqscheduler
 RUN sed -i 's/\r$//g' /start-rqscheduler
 RUN chmod +x /start-rqscheduler
-
-
-COPY --chown=django:django ./compose/production/django/rq/dashboard/start /start-rqdashboard
-RUN sed -i 's/\r$//g' /start-rqdashboard
-RUN chmod +x /start-rqdashboard
 {%- endif %}
 
 # Copy the application from the builder


cookiecutter.json
```diff
@@ -35,6 +35,7 @@
   "use_drf": "n",
   "frontend_pipeline": ["None", "Django Compressor", "Gulp", "Webpack"],
   "use_celery": "n",
+  "use_django_rq": "n",
```
browniebroke: > Should use_celery and use_django_rq be mutually exclusive, or allow both for advanced setups?

AFAIU, Celery and RQ are solving the same problem, so we should design the questions to avoid having both installed. I would change this question to something like `"background_queue": ["None", "Celery", "Django-RQ"]`. Basically turn 2 Y/N questions into one drop-down question.

{{cookiecutter.project_slug}}/docker-compose.local.yml
```diff
@@ -95,6 +102,41 @@ services:
...
     command: /start-flower
 
+  {%- endif %}
+  {%- if cookiecutter.use_django_rq == 'y' %}
+
+  valkey:
+    image: docker.io/valkey/valkey:8.0
+    container_name: {{ cookiecutter.project_slug }}_local_valkey
+    volumes:
+      - {{ cookiecutter.project_slug }}_local_valkey_data:/data
```
browniebroke: I know we're discussing migrating our Redis setup to Valkey, but we should perhaps stick to Redis for now. I would like to avoid us being in a situation where have different datastores, depending on which background queue framework was used.

PR #6233, opened 2025-12-05
Update dependencies for Django 6 compatibility

07pepa:
## Description

update to Django 6

Checklist:

## Rationale

 Django 6 is in "active" supported


luzfcb:
@07pepa, thanks for the PR.

Before any update, we generally need to ensure that all libraries in the dependency tree are compatible with the new Django version. This usually takes weeks or months. https://github.com/cookiecutter/cookiecutter-django/issues/6226

07pepa:
> @07pepa, thanks for the PR.
> 
> Before any update, we generally need to ensure that all libraries in the dependency tree are compatible with the new Django version. This usually takes weeks or months. #6226

thank you.

That is why it is open as draft... it will be in limbo until celery upgrades at least

msamoylov:
FYI, https://github.com/cookiecutter/cookiecutter-django/pull/6402

PR #6258, opened 2025-12-20
Switch Dependabot to uv ecosystem

msamoylov:
## Description

Previously, Dependabot updated only the `pyproject.toml` file without updating the lockfile. This resulted in an outdated `uv.lock` and failed builds. Users then had to run `uv lock` or `docker compose -f docker-compose.local.yml run --rm django uv lock`.

More context:

- [Using uv with dependency bots](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependabot)

Checklist:

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/dependabot.yml
+++ b/{{cookiecutter.project_slug}}/.github/dependabot.yml
@@ -62,7 +62,7 @@ updates:
 
 {%- endif %}
 
-  # Enable version updates for Python/uv - Production
+  # Enable version updates for Python/uv
   - package-ecosystem: 'uv'
     # Look for a `pyproject.toml` in the `root` directory
     directory: '/'


msamoylov:
@browniebroke wait, don't merge yet!

browniebroke:
> wait, don't merge yet!

Ok, I've made it draft. Turn it as "ready for review" when you're good (no need to mention me, it'll show up as changed)

msamoylov:
I added a temporary guard preventing this:

```
Dependabot can't resolve your Python dependency files
  Dependabot failed to update your dependencies because there was an error resolving your Python dependency files.

  Dependabot encountered the following error:

  × No solution found when resolving dependencies:
  ╰─▶ Because django-celery-beat==2.8.1 depends on django>=2.2,<6.0 and your
  project depends on django==6.0, we can conclude that your project and
  django-celery-beat==2.8.1 are incompatible.
  And because your project depends on django-celery-beat==2.8.1, we can
  conclude that your project's requirements are unsatisfiable.
```

PR #6402, opened 2026-03-02
Update dependabot.yml to remove Django 6.0 ignore rule

msamoylov:
## Description

`django-celery-beat` now supports Django 6, see https://github.com/celery/django-celery-beat/releases/tag/v2.9.0.

Checklist:


browniebroke:
Thanks! 🎉

PR #6404, opened 2026-03-02
fix: add DATABASE_URL to .postgres env template to ensure it's available in all container contexts  fix #6273

### LINKED ISSUE #6273: DATABASE_URL is exported in django container entrypoint but not available in container
## What happened?
I started my production build and execed into the django container to run a command that complains about a missing DATABASE_URL.   But the DATABASE_URL is exported in the container entrypoint.sh and should be present.

The DATABASE_URL environment variable should be available inside the docker container.  At least that is what I would expect from the statement on[ line 11 of entrypoint.sh ](https://github.com/cookiecutter/cookiecutter-django/blob/c78fe583367fcf86e879cd1df9dae4655ff20160/%7B%7Bcookiecutter.project_slug%7D%7D/compose/production/django/entrypoint#L11C1-L11C120):
```
export DATABASE_URL="postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}"
```
Perhaps it is either being deleted for security?  or is it not occurring because some/all of the POSTGRES_ environment variables are empty at that point?
I know that I can solve this issue by creating:
```
DATABASES = {
    "default": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "mind_survey_app",
        "USER": "postgres_debug_user",
        "PASSWORD": "....",
        "HOST": "postgres",
        "PORT": "5432",
    }
}
```
in the appropriate settings.py, but I think this might be a security issue for production use.
I can also place the DATABASE_URL line above into the POSTGRES .env file, and I think that works and should be secure. 
*Please let me know if I am wrong!!*

## Additional details
`sudo docker compose -f docker-compose.production.yml up`

- Host system configuration:
  - Version of cookiecutter CLI (get it with `cookiecutter --version`):   latest one, as of today.
  - OS name and version:

    ```bash
    # Insert here the OS name and version
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.3 LTS
Release:	24.04
Codename:	noble
    ```

  - Python version, run `python3 -V`:   - python is inside a venv inside the django container.
  - Docker version (if using Docker), run `docker --version`:
  - docker compose version (if using Docker), run `docker compose --version`:
  - 
  - Docker version 29.1.3, build f52814d
  - ...

- Options selected and/or [replay file](https://cookiecutter.readthedocs.io/en/latest/advanced/replay.html):
  On Linux and macOS: `cat ${HOME}/.cookiecutter_replay/cookiecutter-django.json`
  (Please, take care to remove sensitive information)

```json
 its on my dev vm, which I am not on currently, but standard stuff, whitenoise, vscode, github, containers, etc
```

<summary>
Logs:
<details>
<pre>
$ cookiecutter https://github.com/cookiecutter/cookiecutter-django
project_name [Project Name]: ...
</pre>
</details>
</summary>

shriramThakare3:
### Description

I have added an explicit `DATABASE_URL` environment variable to the `.env` template. This ensures the connection URI is available in all container contexts (including `docker exec` sessions) rather than being dynamically exported only during the entrypoint execution.

I also added a single-line comment to the template to ensure developers keep the URI in sync with the individual `POSTGRES_USER` and `POSTGRES_PASSWORD` variables.

### Checklist:

### Rationale

The current implementation constructs `DATABASE_URL` within `entrypoint.sh`. While this works for the primary container process, it does not persist for subsequent `exec` sessions or external debugging tools. 

By moving this into the environment template:
* **Persistence:** The variable remains available throughout the entire container lifecycle.
* **Clarity:** Other developers can see exactly how the URI is constructed without digging into shell scripts.
* **Consistency:** Using the same placeholders ensures that the user fills in credentials consistently across all fields.

### DIFF
--- a/{{cookiecutter.project_slug}}/.envs/.production/.postgres
+++ b/{{cookiecutter.project_slug}}/.envs/.production/.postgres
@@ -5,3 +5,6 @@ POSTGRES_PORT=5432
 POSTGRES_DB={{ cookiecutter.project_slug }}
 POSTGRES_USER=!!!SET POSTGRES_USER!!!
 POSTGRES_PASSWORD=!!!SET POSTGRES_PASSWORD!!!
+
+#full connection URL 
+DATABASE_URL=postgres://!!!SET POSTGRES_USER!!!:!!!SET POSTGRES_PASSWORD!!!@postgres:5432/{{ cookiecutter.project_slug }}


sdarwin:
Alternate solution:  https://github.com/cookiecutter/cookiecutter-django/pull/6747

PR #6417, opened 2026-03-06
Upgrade factory-boy to 3.3.3 and fix mypy compatibility

a5m0:
## Description

Fix mypy compatibility with factory-boy 3.3.3 type stubs.

factory-boy 3.3.3 added a `py.typed` marker (PEP 561), causing mypy to start using its bundled type stubs. This broke `UserFactory` with errors like `"UserFactory" has no attribute "set_password"`.

- Bump factory-boy 3.3.2 -> 3.3.3
- Annotate `self: User` on `@post_generation` method so mypy resolves User attributes correctly
- Replace `_after_postgeneration` override with `skip_postgeneration_save = True` in Meta, saving explicitly in the password hook when creating
- Use `UserFactory.create()` instead of `UserFactory()` for proper return type inference (`-> T` instead of `Self`)

Checklist:

## Rationale

factory-boy 3.3.3 exports type annotations via PEP 561, which breaks mypy on the generated project's `UserFactory`. Verified locally:

- `uv run mypy .` passes on generated project (only pre-existing `ANYMAIL` annotation issue in `production.py`)
- `reveal_type(UserFactory.create())` returns `User`; `reveal_type(UserFactory())` returns `UserFactory` — confirms `.create()` is needed
- Upstream `test_ruff_check_passes` — 85/86 passed (1 failure is missing Docker in my environment, unrelated)

Ref: https://github.com/FactoryBoy/factory_boy/issues/1113

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements/local.txt
+++ b/{{cookiecutter.project_slug}}/requirements/local.txt
@@ -33,7 +33,7 @@ pre-commit==4.5.1  # https://github.com/pre-commit/pre-commit
 
 # Django
 # ------------------------------------------------------------------------------
-factory-boy==3.3.2  # https://github.com/FactoryBoy/factory_boy
+factory-boy==3.3.3  # https://github.com/FactoryBoy/factory_boy
 
 django-debug-toolbar==6.2.0  # https://github.com/jazzband/django-debug-toolbar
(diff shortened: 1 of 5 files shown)


browniebroke (approved):

browniebroke:
Thanks!

PR #6420, opened 2026-03-07
Simplify Traefik config to use official image with Docker labels

msamoylov:
## Description

Simplify the production Traefik setup by removing the custom Docker image and using the official `docker.io/traefik` image directly with CLI args and Docker labels for routing.

## Rationale

The current Traefik setup uses a custom Docker image that copies in a `traefik.yml` config file. This means any routing change requires rebuilding the image and all routing config lives in a separate file disconnected from the services it routes to.

By switching to the official image with CLI args + Docker labels:

- No custom image to build/maintain: one fewer `Dockerfile`, one fewer build step
- Routing lives next to the service: each service's labels declare its own routes, making the config easier to understand and modify
- Domain and email are env vars: `DOMAIN` and `ACME_EMAIL` come from `.envs/.production/.traefik` instead of being baked into the image at build time, so changing domains doesn't require a rebuild
- Flower no longer needs a dedicated port: served at `/flower/` on port 443 instead of exposing port 5555 separately, reducing the attack surface

Checklist:


msamoylov:
I changed my mind, getting 404 instead of proper 503 sucks.

PR #6464, opened 2026-03-26
Ignore some more generated files

offby1:
"tested" via

    printf "FROM scratch\nCOPY . /" | DOCKER_BUILDKIT=1 docker build -q -f- -o- . | tar t

That command shows you what docker would copy into the build context.

## Description

Checklist:

## Rationale

I've seen docker builds take longer than I'd expected, and this change will likely speed them up.

### DIFF
--- a/{{cookiecutter.project_slug}}/.dockerignore
+++ b/{{cookiecutter.project_slug}}/.dockerignore
@@ -10,3 +10,5 @@
 venv
 .git
 .envs/
+.venv/
+**/__pycache__/


offby1:
Note: I did *not* run either `tests/test_bare.sh` nor `tests/test_docker.sh`, since both of those immediately invoked "sudo", and I'm too lazy to analyze them to see if they're safe.

browniebroke:
Thanks!

PR #6587, opened 2026-05-24
Fix missing HTML title to allauth manage template

mschoettle:
## Description

The manage template from allauth is missing the title block, so the browsers tabs are taking the project name, instead of having meaningful names. This is visible in the Email Addresses and 2FA pages.

Checklist:

## Rationale

For example, when changing the password (`/accounts/password/change/`) no title is shown because the base template uses the `title` block and allauth uses `head_title`.

### DIFF
--- a/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/allauth/layouts/manage.html
+++ b/{{cookiecutter.project_slug}}/{{cookiecutter.project_slug}}/templates/allauth/layouts/manage.html
@@ -1,5 +1,10 @@
 {% raw %}{% extends "base.html" %}
 
+{% block title %}
+  {% block head_title %}
+  {% endblock head_title %}
+{% endblock title %}
+
 {% block main %}
   {% block content %}
   {% endblock content %}


browniebroke:
Can you share a screenshot of before and after, please? I _think_ that was done [on purpose](https://github.com/cookiecutter/cookiecutter-django/pull/4843#discussion_r1501645006) in https://github.com/cookiecutter/cookiecutter-django/pull/4843... @foarsitter can confirm on that

foarsitter:
@browniebroke `entrance.html` does contain the same block, so I think I mist it. 
https://github.com/cookiecutter/cookiecutter-django/blob/8e45d7912d54d4979f9b115099f4a014fe13ce1e/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/templates/allauth/layouts/entrance.html#L6-L10

browniebroke:
Ok I got it, it's to set the _HTML document title_, so it shows up in the browser tabs 👍🏻

browniebroke (approved):

browniebroke:
Thanks!

PR #6603, opened 2026-06-01
feat: add optional django-import-export with generic FK-context import support

gourav-sraoss:
## Summary

Adds an **opt-in** `django-import-export` integration to generated projects, including a **reusable contrib package** that solves a common pain point: selecting one FK value once in the import wizard and having it automatically applied to every imported row — without requiring that column in every line of the CSV file.

Default behaviour is unchanged (`"use_django_import_export": "n"`).

---

## Motivation

Admin-side CSV import/export is one of the most frequently needed features for operational Django projects — data migrations, bulk updates, support workflows. Yet wiring it up correctly (especially the shared-FK pattern) requires boilerplate that every team re-writes. This PR gives generated projects the integration out of the box, and eliminates that boilerplate entirely via a two-class contrib package.

---

### 1 — Template option and app wiring

| File | Change |
|---|---|
| `cookiecutter.json` | New option `use_django_import_export` (default `n`) |
| `pyproject.toml` | Conditionally adds `django-import-export` dependency |
| `config/settings/base.py` | Conditionally adds `import_export` to `THIRD_PARTY_APPS` |

### 2 — Generic FK-context contrib package

When the option is enabled, the generated project gains a ready-to-use contrib package:

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -35,6 +35,7 @@
   "use_async": "n",
   "frontend_pipeline": ["None", "Django Compressor", "Gulp", "Webpack"],
   "use_celery": "n",
+  "use_django_import_export": "n",
   "use_mailpit": "n",
   "use_sentry": "n",
   "use_whitenoise": "n",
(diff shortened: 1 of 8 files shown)


browniebroke:
I used django-import-export in a few project, but I'm not convinced it should be added to the template. 

While making it a project option could be a good compromise, a new option always adds complexity in terms of testing, and adds a question when people generate a new project.

At this stage, I don't think it's worth it, but I will leave this open for while to let people give feedback.

PR #6652, opened 2026-06-19
Skip OpenAPI test file generation when `rest_api` is `None`

hisham-hussien:
## Description

Guard `test_openapi.py` against `rest_api = None` by wrapping the entire file in a top-level Jinja2 conditional, so the file renders empty (valid Python) when no REST API is selected instead of producing broken function bodies.

Checklist:
- [x] I've made sure that tests are updated accordingly, the fix is applied directly to the test script `{{cookiecutter.project_slug}}/users/tests/api/test_openapi.py`, which now renders an empty (valid) Python file when `rest_api = None` instead of broken function bodies

## Rationale

When `rest_api` is set to `None`, the three test functions in users/tests/api/test_openapi.py` each break in a different way:

1. `test_api_docs_not_accessible_by_anonymous_users`, its entire body is wrapped in `{%- if/elif/endif -%}` blocks with whitespace-stripping modifiers. With `rest_api = None` neither branch matches, the body renders empty, and Python raises a `SyntaxError` ("expected an indented block") at collection time, preventing the whole file from loading.

2. `test_api_docs_accessible_by_admin` and `test_api_schema_generated_successfully`, both assign `url` exclusively inside the same conditional blocks, so `url` is never defined in the rendered output. A `NameError` at runtime would follow if the `SyntaxError` above were fixed in isolation.

The file has no purpose when no REST API is selected, there are no OpenAPI docs to test. Wrapping the entire file in:

    {%- if cookiecutter.rest_api != 'None' %}
    ...
    {%- endif %}

renders an empty (but syntactically valid) Python module for the `None` case, correctly reflects the template's intent, and protects any future functions added to this file without requiring per-function `else` guards.


browniebroke (commented):
Rather than rendering this file empty, we should make sure it's deleted in post-generation hook:

https://github.com/cookiecutter/cookiecutter-django/blob/5be68677e89aed5c9a5869a3b63ae9ff278adbaf/hooks/post_gen_project.py#L424

hisham-hussien:
@browniebroke 

We were aware of that, `remove_rest_api_files()` already calls shutil.rmtree on tests/api when `rest_api == 'None'`. However, Cookiecutter supports generating templates without running hooks (`accept_hooks=False`), and this file is the only case in the entire template that is not covered in that mode: the rendering itself produces a `SyntaxError` that the hook never gets a chance to clean up.

To address both concerns, we updated the fix: instead of the outer guard that rendered an empty file, each function now has an else: `pytest.skip("No REST API framework selected")` branch. This means the rendered output is always valid Python (`no SyntaxError`, and `no NameError` on `url`) regardless of whether hooks run.

Together, the two layers cover all cases:
  - `Hooks enabled`  → shutil.rmtree removes the directory; no artifact left behind.
  - `Hooks disabled` → file renders with skip-only functions; valid Python, self-explanatory.

browniebroke:
Oh I didn't know that... I'm pretty sure you'd hit more issues with this template if you skip running hooks, as we rely pretty heavily on that. 

Having said that, the first fix you sent makes more sense than using pytest skip we have now.

hisham-hussien:
@browniebroke 
You're right that disabling hooks would likely expose other issues in this template, given how heavily it relies on them. However, we developed an automated testing tool for Cookiecutter templates that instantiates templates with `accept_hooks=False` (while accounting for pre-gen hooks constraints) to isolate rendering defects from hook side effects, and this file is the only case in the entire template where a `SyntaxError` surfaces in that mode.

We've reverted to the original fix (the outer if guard), which we believe is the right approach here. We've also added a comment in the else branch to make the intent explicit:

    # No REST API framework was selected (rest_api = 'None').
    # These tests do not apply to this configuration.
    # When hooks are enabled, the post-generation hook removes this directory entirely.

This way, the file is never empty silently, the comment explains exactly why no tests are present and points to the hook as the cleanup mechanism for normal usage.

browniebroke (approved):

PR #6714, opened 2026-07-10
Stop setting public read ACL on static files stored in object storage

### LINKED ISSUE #6713: AWS S3 support AccessControlListNotSupported
When creating an AWS S3 bucket, the following is selected by default in AWS.  

`
ACLs disabled (recommended)
All objects in this bucket are owned by this account. Access to this bucket and its objects is specified using only policies.
`

Notice "ACLs disabled".

Then when launching a project with cookiecutter-django, the web server crashes:

`
django-1        | botocore.exceptions.ClientError: An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs
`

A solution seems to be setting: "default_acl": None

```
STORAGES = {
    "default": {
        "BACKEND": "storages.backends.s3.S3Storage",
        "OPTIONS": {
            "location": "media",
            "file_overwrite": False,
        },
    },
    "staticfiles": {
        "BACKEND": "storages.backends.s3.S3Storage",
        "OPTIONS": {
            "location": "static",
            "default_acl": None,
        },
    },
}
```

Could this just simply be set in the cookiecutter templates?   `"default_acl": None`.

Or perhaps better:

AWS_DEFAULT_ACL = None

And remove default_acl from OPTIONS.

TemRevil:
Fixes #6713

New S3 buckets are created with ACLs disabled (Object Ownership: bucket owner enforced), which AWS has made the default and recommends keeping. On such buckets, any `PutObject` that carries an ACL fails with `AccessControlListNotSupported` — so `collectstatic` crashes for projects generated with `cloud_provider=AWS` and `use_whitenoise=n`, because the staticfiles storage sets `default_acl: public-read`. The media storage already writes without an ACL and works fine on the same buckets.

This drops the ACL from the staticfiles options (django-storages then sends no ACL at all) and leaves public read access to a bucket policy, matching AWS's guidance for ACL-disabled buckets. A comment in the template points users at the bucket-policy approach.

Verified by generating projects from the modified template for both `use_whitenoise=y` and `=n` with `cloud_provider=AWS`: no `default_acl` remains anywhere and both rendered `production.py` files parse.

### DIFF
--- a/docs/3-deployment/cloud-storage.rst
+++ b/docs/3-deployment/cloud-storage.rst
@@ -1,0 +1,114 @@
+.. _cloud-storage:
+
+Configuring cloud storage
+=========================
+
+.. index:: storage, S3, Cloud Storage, Azure Storage
+
+If you generated the project with a cloud provider, static files and user uploads are served from that provider's object storage rather than from your application server. This is independent of the platform you deploy to, so follow this page alongside the deployment guide of your choice.
+
+How the template uses your bucket
+---------------------------------
+
+A single bucket (or container, on Azure) holds both kinds of files, under two prefixes:
+
+- ``static/``: the output of ``collectstatic``, meant to be **publicly readable**.
+- ``media/``: user uploads, served through ``MEDIA_URL``.
+
+The storages do **not** set a per-object ACL on upload. Uploaded objects simply inherit whatever access rules the bucket has, so making static files reachable is a one-off bucket configuration step, described below for each provider. This matches the current default of all three providers, which is to disable per-object ACLs in favour of bucket-wide policies.
+
+.. warning:: The instructions below deliberately expose only the ``static/`` prefix. Granting public read on the whole bucket also exposes ``media/``, and since the template sets unsigned URLs for media, every user upload would then be readable by anyone able to guess its URL. See `Keeping media private`_ if uploads in your project are not meant to be public.
+
+Amazon S3
+---------
+
+New buckets have ACLs disabled (*Object Ownership: bucket owner enforced*) and all four **Block Public Access** settings enabled. You need to relax the two policy-related ones, then attach a bucket policy scoped to the ``static/`` prefix.
+
+.. code-block:: bash
+
+    BUCKET=your-bucket-name
+
+    # Outside of us-east-1, add:
+    #   --create-bucket-configuration LocationConstraint=$AWS_REGION
+    aws s3api create-bucket --bucket $BUCKET --region us-east-1
+
+    # Allow public *policies*, while still blocking public ACLs
+    aws s3api put-public-access-block --bucket $BUCKET \
+        --public-access-block-configuration \
+        "BlockPublicAcls=true,IgnorePublicAcls=true,BlockPublicPolicy=false,RestrictPublicBuckets=false"
+
+    cat > /tmp/static-policy.json <<EOF
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Sid": "PublicReadForStaticFiles",
+          "Effect": "Allow",
+          "Principal": "*",
+          "Action": "s3:GetObject",
+          "Resource": "arn:aws:s3:::$BUCKET/static/*"
+        }
+      ]
+    }
+    EOF
+
+    aws s3api put-bucket-policy --bucket $BUCKET --policy file:///tmp/static-policy.json
+
+The same can be done from the console, under the bucket's **Permissions** tab: first edit **Block public access**, then **Bucket policy**. S3 will label the bucket as *Publicly accessible* afterwards, which is expected.
+
+The policy applies to objects already in the bucket, so there is no need to re-run ``collectstatic``.
+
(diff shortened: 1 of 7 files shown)


browniebroke:
Should we update the docs to explain how to configure the bucket? The comment is a bit buried and will be easily missed, no?

TemRevil:
Good point, pushed a troubleshooting entry (AWS S3: AccessControlListNotSupported on collectstatic) with an example bucket policy and where to attach it. Left the inline comment as is since it's still useful context for anyone reading the settings file directly.

browniebroke:
Does it mean that the media storage privacy changed? Looks like it's coming from the same bucket but from a different location, and we were not setting the object ACLs under it...

TemRevil:
Media storage privacy is unchanged. The default storage block (location: "media") never set default_acl for the AWS branch, before or after this diff, only the staticfiles block (location: "static") had default_acl: "public-read", which is the one I removed. Same bucket, different key prefix, and the media prefix's access was already whatever the bucket's default object ownership/policy gives it, that part of the config is untouched.

If a project wants public media too, that would need its own bucket policy statement scoped to media/*, but that is a separate decision from this crash fix, so I left it out. Happy to add a short note about that if you think it belongs in the docs.

browniebroke:
Sorry for the delay, I just tried to deploy a project using S3 to store static files from the main branch and I cannot reproduce the problem you describe. How are you creating the bucket? Can you provide a repeatable script, preferably using the S3 API from the AWS CLI.

I'm using:

- `aws s3api create-bucket --object-ownership BucketOwnerPreferred --bucket $bucket_name` - [docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/create-bucket.html)
- `aws s3api delete-public-access-block --bucket $bucket_name` - [docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/delete-public-access-block.html)
- `aws s3api put-bucket-acl --acl public-read --bucket $bucket_name` - [docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/put-bucket-acl.html)

sdarwin:
"Can you provide a repeatable script".    Maybe this question should be on the issue, rather than the PR.  To answer,
 
The repeatable problem is:   create an S3 bucket in the AWS Console.  (not the cli).   It's not uncommon for it to be done that way.   During the process, AWS tells you ACLs are deprecated.

Then...specifically, don't run `aws s3api put-bucket-acl --acl public-read --bucket $bucket_name`.  Why...  Because "that command sets a bucket ACL, which is exactly the mechanism AWS now steers people away from."   By setting an ACL this is not following their best practices.

Then, django will throw the error mentioned before.

TemRevil:
sdarwin's diagnosis matches the original issue: creating a bucket through the Console now defaults to ACLs disabled (bucket owner enforced), and that's the state that trips the error. Your repro doesn't hit it because `--object-ownership BucketOwnerPreferred` plus `put-bucket-acl` explicitly opts the bucket back into ACL support, which is the older, non-default configuration.

An AWS CLI equivalent of the default Console path:

```
aws s3api create-bucket --bucket $bucket_name --region us-east-1
```

(add `--create-bucket-configuration LocationConstraint=$region` for anything other than us-east-1). No `--object-ownership` flag, so it lands on `BucketOwnerEnforced`, same as a bucket created through the Console with default settings.

Generate a project from main with `cloud_provider=AWS` and `use_whitenoise=n`, point `DJANGO_AWS_STORAGE_BUCKET_NAME` at that bucket, and run `collectstatic`. It should crash on the first static file, since the current template still sets `default_acl: public-read` for the staticfiles storage and this bucket won't accept an ACL at all.

Let me know if that reproduces on your end.

browniebroke:
Thanks for chipping in @sdarwin, you seem to know more about this than me :) 

> Because "that command sets a bucket ACL, which is exactly the mechanism AWS now steers people away from." By setting an ACL this is not following their best practices.

Do you know the rationale for that? What's the problem with setting Bucket ACL? Why is it not best practices and why is it still possible? Are there situations where that would still be needed / desirable?

I feel like they do that to make it harder to have publicly accessible bucket, as historically it was too easy and there was some incident about private files that ended up being stored on a publicly accessible bucket.

Sorry for being picky, but I'd like to really understand the implications of this change. Thanks

@TemRevil your answer doesn't add a whole lot to the discussion and sounds LLM generated. If that's correct, may I ask you to please don't do that, I can ask Claude myself, thank you.

sdarwin:
> Do you know the rationale for that? What's the problem with setting Bucket ACL?

you may ask Claude yourself.  :-)  "ACLs are the 2006 access-control system".  They are outdated. AWS has implemented IAM policies and similar which are now preferred.  

"""
December 13, 2022 announced → April 2023 rolled out — the defaults flipped. Both Block Public Access and ACL-disabling became automatic for all new buckets regardless of creation path (console, CLI, SDK, CloudFormation); these had already been the console defaults since 2018 and 2021 respectively, and existing buckets were left untouched. Deployment began April 5, 2023.
"""

browniebroke:
> you may ask Claude yourself

Touché 😆

browniebroke:
Ok I just tried and it indeed fixes the `collectstatic` issue, however my files in the bucket aren't accessible and return a 401 when the browser tries to access them. Do you also have that? 

It looks like we need to enable public access at the bucket level, via the bucket policy. ~Would be nice to improve our docs on that front, but that's borderline out of scope of this PR.~ I just realised that this documented in the troubleshooting section 🤦🏻

browniebroke (approved):

PR #6732, opened 2026-07-16
Replace `use_mailpit` option by `mail_catcher` with 3 choices: None/Mailpit/Mailtrap

tsokolovs:
## Description

Adds a `use_mailtrap_local` option as a peer alternative to the existing `use_mailpit` prompt. Same docker-compose drop-in pattern; different tool choice for developers who prefer Mailtrap Local. Setting `use_mailtrap_local=y` wires [Mailtrap Local](https://github.com/mailtrap/mailtrap-local) into `docker-compose.local.yml` and the local Django email settings the same way `use_mailpit=y` does today (SMTP on `3535`, web UI on `3550`).

Checklist:

## Rationale

This adds a peer local-email-testing option; it does not change the existing Mailpit default. [Mailtrap Local](https://github.com/mailtrap/mailtrap-local) is an MIT-licensed, single self-contained Go binary that catches local SMTP and exposes a web UI + REST API — the same drop-in role Mailpit fills today. It gives developers who prefer it a first-class local option while leaving Mailpit users completely unaffected. The two options are mutually exclusive (enforced in the pre-gen hook), and the docs, tests, `.gitignore`, and docker-compose wiring mirror the existing Mailpit setup.

**Changes**
- `cookiecutter.json` — new `use_mailtrap_local` prompt (default `n`), peer to `use_mailpit`
- `hooks/pre_gen_project.py` — aborts generation if both catchers are enabled
- `docker-compose.local.yml` — `mailtrap-local` service (with a `tmpfs` mount so the non-root image can start) + `depends_on` wiring for django/celeryworker/celerybeat
- `config/settings/local.py` — `EMAIL_HOST`/`EMAIL_PORT` for docker and non-docker (port `3535`)
- `.gitignore` — ignore the `mailtrap-local` binary (non-docker)
- `tests/test_cookiecutter_generation.py` — supported combinations + mutual-exclusion case
- docs + READMEs — mirror the Mailpit sections


browniebroke:
> the same drop-in role Mailpit fills today

Why make it a separate option, then?

tsokolovs:
> > the same drop-in role Mailpit fills today
> 
> Why make it a separate option, then?

From my point of view, to provide the least amount of friction (keeping it backwards-compatible) and follow existing conventions of having to pick one tool. I don't mind refactoring this into a single option, but that would be a minor breaking change for existing configs that set `use_mailpit: y`.

browniebroke:
We've done these kind of changes before so _if_ we add mailtrap, I would combine it into a single option to choose between mailpit and mailtrap. If it doesn't make sense to pick them together, so let's make it impossible in the questionnaire to pick them...

I'm saying if, because I'm not sure it's worth introducing another choice the user has to make... Folks using this template may not have heard of neither mailpit nor mailtrap, so at the point of prompting they would have to potentially go learn what each does and decide which one to pick if they want this kind of solution. Right now, they only have to decide if they want local SMTP or not. 

IMO this is an area where we should be opiniated and pick the best solution of the 2, based on feature easy to use and licensing. If they are equal in all these areas, the status quo wins. If the current solution isn't maintained, then that's an incentive to switch.

Offering more options or choices always adds complexity: longer CI run, more cognitive overhead on the users, etc...

Thats my 2 cents, I'm not convinced about this at the moment.

tsokolovs:
I've made the change so that there is one choice for mail_catcher (None / Mailpit / Mailtrap Local), defaulting to None, so you can't pick both by construction.

> IMO this is an area where we should be opiniated and pick the best solution of the 2, based on feature easy to use and licensing. If they are equal in all these areas, the status quo wins. If the current solution isn't maintained, then that's an incentive to switch.

Is this additional choice worthwhile? I'd say it's more of the "what does the team already have" than "which one is better". There are multiple options of the "this tool is better than the other" type, e.g., mail_service, cloud provider, CI and in those cases, it's explicitly stated what the default is. Local catcher seems to be in the same boat – teams that are already running Mailtrap somewhere may find it easier to use the same catcher in their dev environment. But as the default is None, people who don't need local SMTP just leave it there and forget about it, as the choice matters only to people who have already decided to use the catcher. 

Right now, both of them might be similar on features and licensing — but that's kind of my point: when two tools are comparable, 'which is objectively best' stops being the useful question, and 'which does the team already run' takes over. That's the same reasoning behind offering a choice for mail_service or CI rather than hard-coding one.

browniebroke:
> I've made the change so that there is one choice for mail_catcher (None / Mailpit / Mailtrap Local), defaulting to None, so you can't pick both by construction.

Thanks, that looks better.

> teams that are already running Mailtrap somewhere may find it easier to use the same catcher in their dev environment.

I see, that's a fair reasoning, when taking into consideration advanced users who know what they want: experienced devs, or established teams. I was more thinking about less experienced folks (in general or just newer to the Django ecosystem) coming to cookiecutter-django without knowing that ahead of time.

> That's the same reasoning behind offering a choice for mail_service or CI rather than hard-coding one.

Yes, I think the difference with these other examples is that I see them as much more important to justify the additional friction of front-loading this choice via an option... For mail catcher, I don't think the tool matters much but perhaps it's because I don't use them myself 🤷🏻 

> But as the default is None, people who don't need local SMTP just leave it there and forget about it, as the choice matters only to people who have already decided to use the catcher.

Another good point, this is pretty much how I would use it. Also, as we already have an option for that, we "just" trading one for another, so it's not as bad as if we were adding an entirely new one.

docs/2-local-development/developing-locally-docker.rst
```diff
@@ -216,14 +216,25 @@ Notice that the ``container_name`` is generated dynamically using your project s
...
 .. _Mailpit: https://github.com/axllent/mailpit/
 
+Mailtrap Local
+~~~~~~~~~~~~~~
+
+When developing locally you can go with `Mailtrap Local`_ for email testing provided ``mail_catcher`` was set to ``Mailtrap Local`` on setup. To proceed,
+
+#. make sure ``<project_slug>_local_mailtrap`` container is up and running;
+
+#. open up ``http://127.0.0.1:3550``.
```
browniebroke, line 237: Should we group these 2 under a "Mail catcher" section? Same for the local dev without Docker

tsokolovs: Yes, that makes sense and aligns with overall changes in this PR. Grouped under `Mail catcher`.

browniebroke (approved):

browniebroke (approved):

browniebroke:
Thanks!

Issue #3224, opened 2021-06-14
Docs update: Can't exec into container to run management commands


camrail:
Hi All,

Firstly thanks for this repo, it's amazing and as you well know saves countless hours (first time using it).

Usually when I setup my django with docker projects I do it in such a way that I exec into the container
`docker exec -it django bash` and then run management commands `python manage.py migrate`

This wasn't working (I was getting a DATABASE_URL missing error), and after some running around thinking I had somehow configured something wrong I came across this post https://github.com/pydanny/cookiecutter-django/issues/2848 which made me realise that I needed to run them this way without exec'ing in:
`docker-compose -f local.yml run --rm django python manage.py migrate`

I gather it's because of the way the environment variables are injected.  I'm not sure if this is obvious to most people but it wasn't to me and I was just wondering if it would be helpful to call out in the docs that you cannot exec into the container and run manage.py that way?  I know it calls out "this is how you execute management commands" but I thought those two methods were interchangeable.  Again could just be me not knowing enough about docker.

Thanks

Andrew-Chen-Wang:
hmm `docker exec` requires a container id no? And if not, then you'd need to add `--name` before `django` (I think) assuming that's the container name that you see in `docker container ls -a`. In other words, I *think* you may just have a syntax error

temp5374:
Hi Andrew `django` is the container name and you don't need `--name` in my experience.

If there was a syntax error in the exec and it was failing completely then I wouldn't have been seeing the DATABASE_URL Django stacktrace.

Based on your response though, are you expecting that environment variables are injected in a way that this should work?  and  that it is different to #2848?

Thanks!

Andrew-Chen-Wang:
Ah you know what I think you're right.

> Based on your response though, are you expecting that environment variables are injected in a way that this should work?

Yea, I assumed that because you weren't injecting environment variables via compose (since you were using `exec` directly), the .env file would have been parsed by Django-environ. I guess it wasn't.

Seems like #2848 does answer this issue though.

Regarding actually adding this to the docs, I think it's already in there for docker: https://cookiecutter-django.readthedocs.io/en/latest/developing-locally-docker.html#execute-management-commands

Though it specifically mentions for management commands, I thought it was well known for compose configuration to use `docker-compose run` (though I learned docker from this repo, so that might be my bias). Idk thoughts?

browniebroke:
[...]

Issue #2821, opened 2020-09-13
django.core.exceptions.ImproperlyConfigured: Set the DATABASE_URL environment variable


EVENFATE:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/environ/environ.py", line 273, in get_value
    value = self.ENVIRON[var]
  File "/usr/local/lib/python3.8/os.py", line 675, in __getitem__
    raise KeyError(key) from None
KeyError: 'DATABASE_URL'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/manage.py", line 31, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 328, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/commands/runserver.py", line 60, in execute
    super().execute(*args, **options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/commands/runserver.py", line 67, in handle
    if not settings.DEBUG and not settings.ALLOWED_HOSTS:
  File "/usr/local/lib/python3.8/site-packages/django/conf/__init__.py", line 76, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python3.8/site-packages/django/conf/__init__.py", line 63, in _setup
    self._wrapped = Settings(settings_module)
  File "/usr/local/lib/python3.8/site-packages/django/conf/__init__.py", line 142, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
[...]

Issue #1888, opened 2018-12-18
The DATABASE_URL environment variable couldn't found in shell


Neulana:
## What happened?
/app # python manage.py shell_plus

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/environ/environ.py", line 273, in get_value
    value = self.ENVIRON[var]
  File "/usr/local/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'DATABASE_URL'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 371, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 365, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 204, in fetch_command
    settings.INSTALLED_APPS
  File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 56, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 43, in _setup
    self._wrapped = Settings(settings_module)
  File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 106, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/local/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/app/config/settings/local.py", line 1, in <module>
    from .base import *  # noqa
  File "/app/config/settings/base.py", line 41, in <module>
    'default': env.db('DATABASE_URL'),
[...]

PR #6747, opened 2026-07-21
Migrate DATABASE_URL settings directly into config/settings/base.py

sdarwin:
Fixes an issue from #6273, #2821, #1888, #3224, #1085, and others.

Currently if you run "docker exec" to an existing container, DATABASE_URL isn't set, because that variable is configured by the entrypoint and "docker exec" will bypass the entrypoint.   This is a fragile configuration which requires developers to understand the mechanism and work around it by manually assigning the variable, or calling the entrypoint.  It keeps tripping people up, as demonstrated by the number of github issues.

A solution is to ensure "DATABASES" is populated directly in config/settings/base.py.   Then it's fine if entrypoint is missed.  We have been using this method in production for years without incident.

### DIFF
--- a/{{cookiecutter.project_slug}}/compose/production/django/entrypoint
+++ b/{{cookiecutter.project_slug}}/compose/production/django/entrypoint
@@ -4,12 +4,6 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-if [ -z "${POSTGRES_USER}" ]; then
-    base_postgres_image_default_user='postgres'
-    export POSTGRES_USER="${base_postgres_image_default_user}"
-fi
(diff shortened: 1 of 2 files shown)


browniebroke:
Hi, thanks for opening this. This is common pain point with the template, it would be nice to fix it. Additionally to the existing PR you found, there was also: https://github.com/cookiecutter/cookiecutter-django/pull/4862. 

Do we need to update the docs?

sdarwin:
Interesting, 4862 would solve it also, right? Why hasn't that one been merged?

browniebroke:
> Why hasn't that one been merged?

There was some unanswered questions in the PR and the conversation died down... I wasn't fully convinced about the change but my stance changed and I'm much more ready to accept it

sdarwin:
The other PR is more complete and thorough, so it's probably better to merge that one, and close this.

The main difference is between a "try-except" block in `config/settings/base.py` versus "if-else".   I don't feel strongly either way, perhaps both methods are acceptable. It's not very different.

PR #6773, opened 2026-07-31
Fix uv lock file Python requirement

igor-wl:
## Description

This was probably missed when project was switched from Python v3.13 to v3.14.
This fixes the after installation warning:
```
warning: Resolving despite existing lockfile due to fork markers being disjoint with requires-python: python_full_version == '3.13.*' vs python_full_version == '3.14.*'
```

Checklist:


browniebroke (approved):

browniebroke:
Good catch... Thanks!

